PR #6, opened 2015-03-03
Move effects from signals into handler functions

paf31:
@jdegoes @cryogenian This PR moves effects out of signal functions and into the top level _handler function_. I have also added `runUIEff` which can be used to build more general top-level loops.

I spent some time thinking about replacing signal functions with regular signals, but I came across some issues which I can discuss tomorrow.

John, I also had some issues with `Aff` which I can discuss separately, so this uses `Eff` for now.

What do you think?

### DIFF
--- a/MODULES.md
+++ b/MODULES.md
@@ -2,10 +2,18 @@
 
 ## Module Halogen
 
+#### `Heff`
+
+``` purescript
+type Heff eff = Eff (dom :: DOM, ref :: Ref | eff)
+```
+
+Wraps the effects required by the `runUI` and `runUIEff` functions.
+
 #### `runUI`
 
 ``` purescript
-runUI :: forall i eff. Signal1 (dom :: DOM, ref :: Ref | eff) i (HTML i) -> Eff (dom :: DOM, ref :: Ref | eff) Node
+runUI :: forall i eff. Signal1 i (HTML i) -> Heff eff Node
 ```
 
 `runUI` takes a UI represented as a signal function, and renders it to the DOM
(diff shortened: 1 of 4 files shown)


garyb:
Looks nice so far.

jdegoes:
Looks good, can you squash commits? Also interested in the issues with `Aff`.

paf31:
Yes, can do.

I still don't know if we have reached a consensus on the direction we want to go with this. I think the current incarnation provides enough to implement real apps (or will, once I finish #3), but as we've discussed, there are trade-offs.

My remaining question is:

> Do we want to go with signal functions or signals?

If we pick signal functions, we get a very clear, pure interpretation of the UI, but lose some expressiveness. If we pick signals, we have to decide _what flavor of signals_, each of which comes with its own tradeoffs. We also lose the nice separation of effects.

jdegoes:
@cryogenian / @garyb Thoughts on the above???

paf31:
@jdegoes The issue with `Aff` was that I couldn't get the asynchronous effects to line up in the types. If you try to implement `runUIAff` using `runUIEff`, you'll probably see what I mean. I'll take another stab at it later.

jdegoes:
@paf31 I feared this representation may be too much for the type inferencer. If we can't get it working, I'll revert to the old formulation in which there are only "effects".

paf31:
I've had similar problems in the past. The issue seems to be that you need to a) avoid duplicate labels, and b) keep things composable, which means various `eff`s need to unify. That seems to imply that once you either run into infinite types, or you lose monadic bind without `unsafeInterleaveEff`.

jdegoes:
OK, I'll revert `aff` today.

paf31:
Well, maybe we can still figure it out.

jdegoes:
OK, keep me posted.

garyb:
@paf31 Sorry if I'm being a bit slow, but do you have an example of what signal functions vs signals might look like? I have an idea of what you mean but just want to be sure.

paf31:
@garyb Sure. Signal functions are a pure model of "stateful functions":

```purescript
data SF a b = SF (a -> Tuple b (SF a b))
```

You put in an `a`, you get out a `b` and a new signal function. Signal functions don't deal with the source of data (e.g. `setTimeout`, the mouse position, whatever) but only with what happens to the data once it arrives (update the application state and output a HTML document in our case). The problem then becomes, how do you recover what you lost, namely effects and connecting the signal function to the world. In the current model, this is handled by a separate function, so we end up with something like

```purescript
type App i o = 
  { view: SF i (HTML o)
  , update: o -> ContT (Eff _) i
  }
```

So we've separated the app into a pure part and an impure part. At the top-level `runUI` takes this structure and uses `virtual-dom` to render it and keep it up to date.

In the signal case, we'd use something like `purescript-signal`, signals would be effectful and connected to the world, and at the top-level you have a `Signal HTML` which you can subscribe to, and pass the resulting `HTML` to `virtual-dom` for rendering. To handle events, I suggested a top-level representation like `Signal i -> Signal (HTML i)`, and again, `runUI` would be responsible for tying the knot.

Issue #12, opened 2015-03-06
Hooks


cryogenian:
Could you please make event handlers hooks not attributes, and make those hooks composable some way? 
The main consideration of doing it: 

* Everytime ```VTree``` rerender it adds nodes with handlers that are not hooks.
* Adding such handler will remove any other handler one has added to this node.

paf31:
What about either using something like `dom-delegator`, or `data-*` attributes and bubbling?

I really don't know, this isn't my area of expertise.

Maybe @garyb would be interested in this thread too.

cryogenian:
I don't insist on using hooks but switching from ```on*``` attributes. So, any other libs that allow to add multiple handlers and don't force to rerender elements with handlers would be nice.

paf31:
Yes, I agree it's not really ideal right now :)

jdegoes:
@cryogenian What are the features you desire? Just that the `on*` attributes are not used, and instead, `addEventListener` etc., are used for adding these handlers?

The handlers are basically `Event -> i`, these can compose nicely since the library can generate multiple `i` per single event (so in theory multiple handlers can be supported).

The tricky bit is that these handlers are functions. So if you don't want to remove them, and re-add them, for every re-render, then you have to have some way of knowing when they change. That's function equality. :scream: Maybe there's some clever way to do it or some constraints that can be imposed to make the problem tractable.

paf31:
@jdegoes But they don't need to be functions. With the `data-*` approach, we can set attributes which basically represent the `i`s, have the events bubble and catch them at the root, then pass the event with the input to the handler.

jdegoes:
Well, let's figure out how to hide all this complexity as well as provide a nice type-safe API. If at all humanly possible. :smile:

cryogenian:
The problem is not in removing/adding listeners, it's acceptable. But in ```patch```es. When one get ```diff``` from one ```vdom``` with ```onclick``` and this ```onclick``` unchanged he has to know that functions are equal, ```vdom``` just says that it's impossible and put elements with such handlers in ```patch```. I don't know is there performance issues with ```addEventListener``` but updating **DOM** is slow. So, I suppose, that *hooks* will work better. 
Bubbling/delegating will work also.

One example of composing: 
* Need to trigger file upload dialog
* Define ```HTML i -> HTML i``` where result has hidden ```<input type="file" />```
* Make listeners on parent click to trigger ```input```'s click
* Compose this listener with parent ```click``` listeners 
[...]

PR #16, opened 2015-03-08
Use hooks for events

paf31:
This seems to give a nice trade-off between the simplicity of `on*` events, and the efficiency of `dom-delegator`.

I basically took this straight from @cryogenian's `slamdata` code. It results in a few calls to `{add,remove}EventListener` on each render, but those are pretty much necessary for the reasons explained in #12.


paf31:
Resolves #12.

PR #35, opened 2015-03-12
Added onInput to FormEvents

cryogenian:


garyb:
:+1:

paf31:
Looks good. 

Could you please just change `onSomething` to something a little more descriptive? I know it's internal, but maybe something like `handleForeignEvent` could work? Then it could even get moved into its own module along with other functions using `Foreign`.

Also, we've lost docs on the functions which are exposed. Could you please add them back?

cryogenian:
* Events are not foreign :) I suppose ```addForeignPropHandler``` is not good name too, but it's declare that ```prop``` is foreign. 
* Added back docs 

Can we add some lib to work with ```Node```? ```purescript-simple-dom``` has ```value``` function i.e. 
Unfortunately it's effectful. Or maybe it should be effectful, since it's mutable. What do you think?

jdegoes:
Personally, I'd prefer not to add a dependency to simple-dom since we've managed to avoid it so far.

cryogenian:
Then we must to provide almost all operation on ```Node``` like getting parent, children, properties etc.

jdegoes:
> Then we must to provide almost all operation on Node like getting parent, children, properties etc.

Why is this needed?

cryogenian:
* _properties_ - getting ```value```, ```files```
* _children_ and _parent_ - to proxy events

jdegoes:
Instead of exposing the raw Node, maybe another option is to provide another facility for accomplishing the same ends?

cryogenian:
Adding fields to ```Event``` could work. Also one can use something like widget to proxy events. But it's much more complicated then using ```Node``` because of composition problems.

paf31:
`Foreign` just means that a value has an unknown or unreliable type. Events are foreign in the sense that I don't trust them to have any particular type, at least not across browsers. In order to use the `change` event in a type-safe way across different controls, I need to handle the cases where `value` is a `String`, a `Number`, etc., so for all intents and purposes, they're foreign.

Issue #39, opened 2015-03-18
Finally tagless HTML representation


paf31:

paf31:
@jdegoes This is the one remaining item on my list for 0.4, and I should have some time tomorrow evening to spend on it.

jdegoes:
:+1:

paf31:
So I made another attempt at this, here:

https://github.com/slamdata/purescript-halogen/tree/final2

I spent two hours on it, but I just kept coming up against typechecker issues:

- I can't use `[Attr i]` because of impredicative type issues.
- I lose the ability to use `Semigroup` on `Attr i` because you can't write things like `class (Semigroup (f i)) <= Attr f` so I have to use `Plus` which is a bit weird.
- The type signatures and error messages are baffling. If I were a newcomer, I would probably find it pretty offputting.

I feel like I'm pretty close, but it's not working yet, and at this point, I'm starting to feel like it's a dead end, sadly. 

Maybe I can address some of these issues by wrapping things with newtypes, instead of using rank-2 types, so I'll spend a little time on that and see how far I get.

paf31:
Ok, newtypes fixed pretty much everything here :smile: 

It's still not entirely done, but things build and the demos work. I'll make a PR so that you can take a look.

PR #43, opened 2015-03-20
Final encoding of HTML

paf31:
@jdegoes This is pretty much done, except that:

- Traversals have been removed. I need to experiment with adding an initial encoding on which we can perform general-purpose traversals.
- The `STProps` approach had to go. Maybe I can figure out a way to get it back in, but I don't think it's the end of the world.
- I might need to tidy up the whole `Bifunctor`/`Plus` thing. I haven't made my mind up yet, but using `bifunctors` will require some upstream changes anyway.

Let me know what you think. I'll let you decide if the complexity of the implementation is worth the flexibility. Now that this is working again, I'm leaning towards thinking that it is.


paf31:
Resolves #39

paf31:
@jdegoes @garyb @cryogenian Any thoughts on this? Do you think it's worth spending time to finish this off?

jdegoes:
Will review today.

garyb:
I don't think I can review this really. I've read up on "finally tagless" again, tried to get my head around the code, but honestly I just don't understand what this change actually means.

jdegoes:
I do think it's worth the time. This change did not seem to complicate the surface API that much, in fact, none of the examples changed at all. 

In addition to having the ability to render to strings, directly to the DOM, or to in-memory data structures (maybe useful for unit tests on node), if this approach were factored out into a library, it would give third-party libraries like `purescript-jtable` a way to generically create HTML-like data, with less performance overhead than a `Free` monad approach.

What are your thoughts? Didn't seem like we lost too much here, and gained a lot, without complicating the external API, so what are the drawbacks?

cryogenian:
* Examples has not changed
* Code has become more abstract 
* Code amount has not changed
* I don't think that code is much more complicated than before 

So, I think it's worth the time

paf31:
I guess my main concern is that you still have to build up a data structure before rendering. It may or may not be as computationally intensive as the old `HTML`. I think this approach only has a benefit if we can exploit some kind of fusion to render directly to `VTree`s without the intermediate `HTML`. 

As it stands right now, we haven't gained much because the `HTML` newtype has to be constructed along with its child "nodes" etc., before the wrapped function gets applied all at once.

On the other hand, without the newtype, it's currently not possible to define things like `Functor` for `HTML`, which is much worse.

So, all in all, we've lost traversals (without the initial representation, which I'll work on next), props are slower, and the internals are more complicated. The only plus I can see is that we might have improved performance.

paf31:
Ok, I think this is ready to review. Barring the one main concern I explained above, I think this is good to go.

cryogenian:
Does it make neccessary to use `fromAttr`/`toAttr` in case of adding css classes as a reaction on user input?

paf31:
No, but you would need to use it to modify an existing `HTML` structure to add or remove a class.

cryogenian:
That's not good :disappointed: 
Switching classes is very common operation

paf31:
Yes, but you can switch classes using the view function based on a Boolean in the model. You only need this if you want _modify_ something.

cryogenian:
You are right! Thanks

paf31:
It is a disadvantage though.

paf31:
@jdegoes Given the above, do you still want to merge this?

jdegoes:
And am I correct in thinking that with this work, we could potentially make libraries like `purescript-jtable` work across both Halogen and Thermite (for example)?

paf31:
I'm not sure how that would work without completely changing how `thermite` works internally. Changing the representation of HTML is one thing, but this would need a change to how events are represented as well. I think it's just as possible as it was with the initial encoding.

jdegoes:
Yes, it would require a rework, but the main advantage I see in this approach is that multiple UI libraries (and some things which are not UI libraries, e.g. sever-side testing frameworks, etc.) could work with the same views. For example, thermite and halogen using the same Bootstrap themes, the same jtable component, etc. Of course, you can do that with the initial encoding (through costly data transformations), but you cannot "directly" generate the React HTML without something like this final approach.

If you think that's not going to be a benefit, then the benefit of this approach is limited to, I suppose, whatever kind of renderers we can define and use in halogen.

paf31:
I think it _is_ a benefit, and I think it would be possible to change `thermite` to use this approach. It's a little tricky because of how the React `this` gets passed around.

My main concern is that we don't get as big a performance gain as we might with this approach, because we have to construct all the `HTML` newtypes up front. We don't really generate anything "directly" because you have to build a big tower of functions, which then get applied all at once. The intermediate data structure is just as big as before, in fact, just represented differently. Maybe adding some functions to defer execution would help there, but then it's just syntactic noise with no obvious purpose when it comes to the user.

paf31:
What I'm saying is that, ideally, you would have one type class (`HTMLRepr`) constraint, in the view, so you specify a dictionary once, and you build the `VTree` right there, or whatever. But it doesn't work that way. You constantly have to unpack and repack the dictionary into a `newtype` for every node you render. So you build this big structure of dictionary-taking functions, and then `renderHTML` collapses that into a `VTree`.

If we could use rank-2 types instead of `newtype`s, it might be better, but you wouldn't have access to even basic things like a `Functor` instance for `HTML`. Also, I ran into some major typechecker hurdles this way, and I worry that the error messages might put new users off.

jdegoes:
Don't newtypes disappear in compilation? Or I guess not in this case because they capture a dictionary?

paf31:
Newtypes have the same representation as their underlying type, but in this case, the underlying type is a constrained type, which corresponds to a function.

We can lose the `Functor` and `Bifunctor` instances and the speed could improve, at the expense of some odd errors. To get `<$>` and `bimap` back, you would have to move to the initial representation, or you could use a monomorphic equivalent (not use the type class).

paf31:
I've wrapped up some independent things, but most of what I have to work on now relies on finalizing an API (slamdown integration, integrating slamdata frontend), so hopefully if you're online tomorrow @jdegoes, we can make a decision on this.

paf31:
@jdegoes I managed to simplify this a little bit by using the old approach but skipping the rank 2 types where possible. How does this look? I'm much more inclined to merge this back to master now, since the performance should in theory be better.

paf31:
One thing remains which I'm not quite happy about: I can't eliminate the `Attr` newtype without something like associated types or fundeps, neither of which we have, obviously. The irritating thing is that the final approach allows us to do nice things like associate an attribute name with its value type, which is less simple with the initial approach.

paf31:
Ok, this is about as much as I can do to simplify the `HTML` data structure. I can separate some of the performance changes into another PR.

@jdegoes Do you want to review one more time, before I merge?

After this my main priorities in order (unless you have something else) are:

- Try out some alternative `View` types for better composition
- Try to speed up the generation of `VTree`s by avoiding passing various functions around

which are the remaining blockers for a release. Then:

- Integrate `purescript-markdown`
- Extra Bootstrap components

jdegoes:
This looks good. Let's get this in and we can push remaining changes to a separate PR. Also worth considering if we should break out the final machinery into a separate lib so things like `jtable` can be written to it independently of halogen (if you want to do that just pick a name and I'll create a repo).

As for final issues before release, I think:

1. Trying out alternate formulations of `View` / `runUI` etc., in particular I'm quite interested in existentially hiding the user events, and your idea of existentially hiding the state. We can talk about these approaches on the other ticket.
2. Making sure we can embed something like echarts or ace using the widget machinery. In fact, an example around this would be great, you could add the lib as a devDependency so as not to bloat the whole distribution.

Performance and other goals could wait till our release unless you think they'd affect the API. Both of the above are issues that have the potential to do that.

paf31:
Sounds good. I should have some time to work on this tonight.

Some notes:

- Pulling out the final HTML stuff into a library sounds good, but I'd want to pick the attributes representation first. Namely, it's quite bulky right now, and I'd be interested to try the combination of initial `Attr` (with `Exists`) and final `HTML`. 
- There's also the issue of what to do about placeholders which are pretty much a Halogen-only concept right now. Maybe we could pull the `placeholder` function out, and add it in a custom subclass.
- State is already existentially hidden in `SF`, via the "nu-type-as-existential" isomorphism, I'm just thinking about how to make that iso more obvious. Using the existential encoding is bad for the API, and leaving the existential off and just using functions is bad for composability, so I think `SF` gives us something useful. I do think there is some mileage in making something like `Component m p ~ exists i. SF1 i (HTML p (m i))` and then providing some ways to compose those. I'll look into that later.
- I can spend some time trying to wrap ace later, since I'm most familiar with that one obviously.

jdegoes:
> Pulling out the final HTML stuff into a library sounds good, but I'd want to pick the attributes representation first.

Good idea.

> There's also the issue of what to do about placeholders which are pretty much a Halogen-only concept right now.

Extensible "GADTs" FTW! :smile: Although thermite does not yet have a nice story for 3rd party widgets, so who knows where it could be useful.

> I do think there is some mileage in making something like `Component m p ~ exists i. SF1 i (HTML p (m i))`

Yes, super easy to compose those, and only marginally more difficult to compose the more powerful `Component m p req res`.

> I can spend some time trying to wrap ace later, since I'm most familiar with that one obviously.

Super. Having this in the examples will ensure it's production ready when Maxim & Gary get around to those parts of the app, and also will help with hand-holding for 3rd party users of the library.

PR #46, opened 2015-03-21
Combine latest

### LINKED ISSUE #44: Composing two Views is somewhat problematic due to event handlers
Given two trivial views with different state types:
```purescript
data Count = Inc | Dec

app1 :: PureView Count
app1 = render <$> stateful 0 update where
  render count =
    H.p_
      [ H.text (show count)
      , H.button (A.onclick (const $ pure Inc)) [ H.text "+" ]
      , H.button (A.onclick (const $ pure Dec)) [ H.text "-" ]]
  update count Inc = count + 1
  update count Dec = count - 1

app2 :: PureView String
app2 = render <$> stateful "" (flip const) where
  render val =
    H.p_
      [ H.input (A.type_ "text" <> A.value val <> A.onInput pure) []
      , H.text val ]
```

I want to compose these together so they are shown one after another (so basically I want to `<>` their HTML). These seems like something you could easily do with Profunctor and Apply (split the combined states, and then append the HTML) but the problem is that the SF/SF1 instances don't touch the event handlers, which would have to know about the new combined state. You can't rmap an additional map on to the HTML to modify the resulting handlers because you don't have access to the composite state to do an update. Also I think you'd have to continually pay for each map, which would likely be bad for app written from lots of smaller parts. What's the best way to go about something like this?

paf31:
Resolves #44.

@natefaubion @jdegoes How does this look?


paf31:
Oops, wrong branch :(

Issue #48, opened 2015-03-24
Child to parent event signalling


natefaubion:
Running with a typeahead example, I'm wanting to wrap it up into its own component that I can reuse. It has a decent amount of state and event handlers (clicks, keypresses), but I'm wanting to be able to signal to the parent when the user makes a selection (so basically a custom `onselect` event handler, which is not a DOM event). Communication is currently one way from parent to child, and there's no way for a child to notify a parent with the idioms mentioned in #44 (at least as far as I can tell).

The problem with the `<$>` idiom is that it only, and always touches all DOM handlers. So if I do `TAState <$> myTypeahead`, all results from event handlers will always be wrapped (and thus can't be reasonably intercepted by the parent).

It seems like you could do another level of wrapping in the handler and do something like `listen TAEvent TAState <$> myTypeahead` where `listen` say wraps with `TAEvent` when given a `Left` to signal the parent, or with `TAState` when given a `Right` if its not signalling the parent. And then the typeahead component would have to make sure to map a `Right` over its inner components. The result given to `TAEvent` would then naturally wrap a sum of possible events raised by the typeahead.

natefaubion:
Basically something like this, if this makes sense

```purescript
data Respond a b = Respond a | Continue b
listen :: forall a b r p f. (Functor f) => (a -> r) -> (b -> r) -> HTML p (f (Respond a b)) -> HTML p (f r)
```

Which essentially looks like mapping a bimap

natefaubion:
Here's what I landed on

```purescript
module App where

import Data.Tuple
import Data.Maybe
import Control.Monad.Eff

import Halogen
import Halogen.Signal
import qualified Halogen.HTML as H
import qualified Halogen.HTML.Attributes as A
import qualified Halogen.HTML.Events as A
import qualified Halogen.HTML.Events.Forms as A
import qualified Halogen.HTML.Events.Handler as E

data Respond a b = Respond (Maybe a) b

listen :: forall a b r. (Maybe a -> b -> r) -> Respond a b -> r
listen f (Respond a b) = f a b

---

type CountState = Number
data CountEvent = Counted
data CountInput = Count

countTil :: Number -> CountState -> H.HTML _ (Respond CountEvent CountInput)
countTil til n =
  H.div_
    [ H.button (A.onclick \_ -> pure $ handleClick (n + 1)) [ H.text "count" ]
    , H.text (show n) ]
  where
  handleClick n | n == til  = Respond (Just Counted) Count
[...]

Issue #45, opened 2015-03-21
Composing Views that rely on effects is problematic


natefaubion:
Currently the effect/request handler is completely separate from the View machinery. While I can potentially work with the SF/SF1 instances to manipulate state and HTML, there's no story for composing views that need to make requests. Say I want to create a reusable typeahead component that makes AJAX requests. These requests exists purely within event handlers and there's no clear way to pass these requests to the request handler and have it map back to the correct nested state.

paf31:
We used to have effects in the signal functions which made this easier, but removed them because it's easier to reason about the view if there are no effects involved. I guess this is the price for a pure signal function. I'll have a think about how we might resolve it.

paf31:
@jdegoes I can see certain benefits of keeping the view pure, but this is an interesting point. Do you think it is worth thinking about moving `Aff` back under the signal function?

paf31:
@jdegoes Another option here is to stick a monad inside `SF`, then we can keep pure signal functions where they're wanted, using `Identity`, or use `Aff` when we need requests. We could use type classes like `MonadEff` for composition (or `MonadCont` + `MonadEff`, etc.)?

jdegoes:
Will post some thoughts here later today, have a meeting right now.

jdegoes:
Here are some preliminary thoughts on this:

Currently, `i` represents user events which are handled internally, user events intended to bubble up, external events (such as ticking of a clock), and possibly responses to requests. I'd suggest this is perhaps too overloaded and composition might improve if we are able to separate out distinct use cases.

This is just one idea, not very well thought out, but what are there portions of `i` that could be hidden via existentials to simplify external composition? i.e. internal events intended to only affect the state?

Regardless, currently the type of `View` is:

```purescript
type View i p r = SF1 i (HTML p (Either i r)) 
```

This is very specific, the obvious generalization is:

```purescript
type View i p r = SF1 i (HTML p r) 
```

I think this introduces other complications since then you need a way to determine whether or not to feed back an event into the signal function or to pass it up to a handler. But still, worth thinking about as the more general signature could change composition (for better or worse).

Another approach would be to go more specific and incorporate requests `q` and responses `s` into the signature:

```purescript
type View i p s q = SF1 (Either i s) (HTML p (Either i q))
```

The signal function is thus forced to process either a response to a request or a user input, and can emit either a user input, or a request.

If one hides `i` via existentials and forces notifications along the request channel, one achieves the simpler:

```purescript
[...]

PR #50, opened 2015-03-27
First pass at component module

### LINKED ISSUE #44: Composing two Views is somewhat problematic due to event handlers

paf31:
Resolves #44, #45, #48.

@jdegoes Could you please review?

/cc @natefaubion @garyb @cryogenian

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -21,13 +21,14 @@
     "purescript-refs": "~0.1.2",
     "purescript-dom": "~0.1.1",
     "purescript-foldable-traversable": "~0.3.0",
-    "purescript-profunctor": "~0.1.0",
+    "purescript-profunctor": "~0.2.1",
     "purescript-aff": "~0.6.0",
     "purescript-arrays": "~0.3.3",
     "purescript-foreign": "~0.4.0",
     "purescript-nullable": "~0.1.1",
     "purescript-maps": "~0.3.2",
     "purescript-bifunctors": "~0.3.1",
-    "purescript-void": "~0.2.0"
+    "purescript-void": "~0.2.0",
+    "purescript-exists": "~0.1.1"
   }
 }
--- a/docs/Halogen-Component.md
+++ b/docs/Halogen-Component.md
@@ -1,0 +1,81 @@
+# Module Documentation
+
+## Module Halogen.Component
+
+
+This module defines a type of composable _components_.
+
+#### `Component`
+
+``` purescript
+newtype Component p m node req res
+```
+
+A component.
+
+The type parameters are, in order:
+
+- `p`, the type of _placeholders_
+- `m`, the monad used to track effects required by external requests
+- `node`, the type of node used to render content
+- `req`, the type of external requests
+- `res`, the type of external responses
+
+Request and response types are public, but the component may also use an _internal_ type
+of messages, as illustrated by the type of the `component` function.
+
+The main interface to Halogen is the `runUI` function, which takes a component as an argument,
+with certain constraints between the type arguments. This module leaves the type arguments
+unrestricted, allowing components to be composed in various ways.
+
+If you do not use a particular feature (e.g. placeholders, requests), you might like to leave 
+the corresponding type parameter unconstrained in the declaration of your component. 
+
+#### `component`
+
+``` purescript
+component :: forall p m node req res i. (Bifunctor node, Functor m) => SF1 req (node p (m res)) -> Component p m node req res
+```
+
+Create a component by providing a signal function.
+
+The signal function should consume external requests and produce DOM nodes. The DOM
+nodes in turn will create (monadic) external requests.
+
+See the `Halogen.Signal` documentation.
+
+#### `component'`
+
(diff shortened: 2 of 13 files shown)


cryogenian:
What does `combine` do? Does it make one component from two siblings?

cryogenian:
What does `combine` do? Does it make one component from two siblings?

paf31:
Yes exactly.

jdegoes:
I'll review this more thoroughly over the weekend, but a ton of things seem to get simpler and more composable. The examples seemed to get simpler or not change, except for the `pure (pure .)` stuff which could probably use a helper or two (especially for `component'`, it'd be nice to have helpers).

Again, no final thoughts on this until I do a more thorough review, but what are your initial thoughts??

paf31:
> but what are your initial thoughts??

I like it :)

I need to try using `compose` for something - so far, I've only typechecked it.

jdegoes:
> I need to try using compose for something - so far, I've only typechecked it.

Then of course it's going to work, right? :smile: 

The other thing I was thinking, re: compose: maybe it would be possible to use placeholder to allow embedding a component *inside* a component? That would be insanely awesome.

paf31:
It should be possible using `graft` I think if the types line up.

natefaubion:
I'll try to play with it some this weekend as well.

>The other thing I was thinking, re: compose: maybe it would be possible to use placeholder to allow embedding a component inside a component? That would be insanely awesome.

This is desirable, and I think you'll immediately want something analogous to Angular's transclusion. Ie, you'll want to run a component where you give it `HTML` that retains a binding to its original component. Whereas in Angular you want it to stay bound to the original `$scope` (and not the child's `$scope`), here you'd want handlers and state to route back to the original component.

natefaubion:
One thing missing is a way to provide a component an initial input, to say kick off a request to fetch data on load.

paf31:
@natefaubion Doesn't `onload` give you that?

natefaubion:
The onload event doesn't work for any arbitrary element. It's for body and things that load external resources like img or script.

paf31:
Ok, but it seems like there must be some way that could be solved using events. It seems like a bit of a shame to introduce effects at the top level. Using `Aff` would make the initial render potentially delayed for example.

natefaubion:
I mean, you can do it if you bundle everything into a widget and use init. But I don't think there's going to be a quick, widely supported DOM event you can hook into.

paf31:
I think this is close enough to merge, yes? We can probably figure something out for initialization later on, given that the internals of the module aren't exported anyway.

natefaubion:
Providing support for [hooks](https://github.com/Matt-Esch/virtual-dom/blob/7cd99a160f8d7c9953e71e0b26a740dae40e55fc/docs/hooks.md) as well as widgets would be useful and allow you to implement initialization, though I don't know how much of the virtual-dom API you want to leak through.

paf31:
Agreed, and maybe we can wrap them in a restricted way like we have for widgets. Could you maybe give a sketch of how you see hooks working for initialization?

I think the tricky bit is going to be keeping a sane API/ `Component` is already impure because of the `m` argument, so maybe we just have to wrap the whole signal function with an extra `m`. My plan was to just bubble everything to `runUI` where the initializers would get run.

jdegoes:
> Ok, but it seems like there must be some way that could be solved using events. It seems like a bit of a shame to introduce effects at the top level.

:+1: for this, even if we have to fake our own `oninit`-style event (i.e. intercept and generate it). Wrapping the whole thing in effects is not a good solution as it could lead to blocking the UI indefinitely. The current formulation is carefully constructed so that effects may block UI updates but may not block UI rendering, which is a very important property to preserve.

jdegoes:
Also, I know this is already merged, but :+1:. I will have a few comments or questions on helper functions, component nesting, etc., but overall this change seemed to do a lot of good so I'm in favor of it.

Issue #55, opened 2015-03-29
Make widgets as powerful as components


paf31:
and update the `ace` example

jdegoes:
In fact, I'd argue that at some level, a widget must be able to be wrapped up into a component, so that it can accept and generate messages exactly like a component. Strictly one-way communication will not work for many use cases and will necessitate messy workarounds.

paf31:
I think it should be possible to wrap widgets as components using some kind of helper function.

The tricky part right now is that I'm running into variance issues trying to make `Widget` into a `Functor`. The type of the renderer becomes:

```purescript
(i -> Eff eff Unit) -> (p -> Widget eff i) -> HTML p i -> VTree
```

which is invariant in `i`. `Widget` essentially has become 

```purescript
type Widget eff i = (i -> Eff eff Unit) -> VTree
```

which is covariant.

Maybe I've just been staring at this for too long, but I feel like these pieces should be fitting together.

jdegoes:
> which is covariant, but this is tricky because of the finally tagless approach.

Well, finally tagless is subservient to good composable abstractions (IMO), so if it proves an obstacle, we can always go back to initial.

jdegoes:
This ticket should really be retitled, "Widgets should be as powerful as Components".

paf31:
Fixed in #58. The Ace editor example is a good demo of this.

PR #58, opened 2015-04-03
Widgets as components

paf31:
@jdegoes This is pretty much ready to review. I just have to update the Ace example.

- Widgets now have access to the driver function and can therefore send events back into the signal function.
- Components can now be created from Widgets.
- Widgets have access to the driver function so are able to provide outputs.
- Components created from widgets can respond directly to inputs, without having to update an intermediate state. This is good for things like Ace where you don't want to call `setDocument` after every unrelated update.

I unfortunately had to get rid of the final encoding to support `<$>` on widgets, due to the variance issue I mentioned in #55.


docs/Halogen-Component.md
```diff
@@ -47,15 +46,23 @@ See the `Halogen.Signal` documentation.
...
-component' :: forall p m node req res i. SF1 (Either i req) (node p (m (Either i res))) -> Component p m node req res
+component' :: forall p m req res i. SF1 (Either i req) (HTML p (m (Either i res))) -> Component p m req res
 ```
 
 A variant of `component` which creates a component with some internal, hidden input type.
 
+#### `widget`
+
+``` purescript
+widget :: forall eff i s m. (Functor m) => String -> String -> ((i -> Eff eff Unit) -> Eff eff { node :: Node, state :: s }) -> (i -> s -> Node -> Eff eff (Maybe Node)) -> (s -> Node -> Eff eff Unit) -> Component (Widget eff i) m i i
```
jdegoes: Could a record help clean this up? Also hoping we can find a way to remove these `Eff` callbacks through Aff-level MVar / Stream at some point.

docs/Halogen-HTML.md
```diff
@@ -1742,6 +1777,84 @@ wbr_ :: forall p i node. (HTMLRepr node) => [node p i] -> node p i
 This module enumerates some common HTML attributes, and provides additional
 helper functions for working with CSS classes.
 
+#### `ExistsR`
+
+``` purescript
+data ExistsR (f :: # * -> *)
```
jdegoes: Would this be useful inside the `purescript-exists` package, or too specific to Halogen's use case?

jdegoes:
:+1: Looks good to me, ready to merge when you're ready.

docs/Halogen-HTML.md
paf31: Possibly, it's a real shame we don't have `PolyKinds`. `Exists` would make such a nice application of it. I thought it best not to move it there, since then we would need to support `Exists` for some (suitably small but) arbitrary set of kinds. I'd rather wait for `PolyKinds` honestly.

docs/Halogen-Component.md
paf31: Definitely, I can change this over.

paf31:
Ok, I'm really pleased with this now actually :) It's a little bit of a shame about the final/initial change, but the Ace example now has a widget acting as a component, and interacting with another component via `combine`.

jdegoes:
The `Ace` example looks amazing, although in general, it is necessary to place child components at particular locations inside the parent component (and of course have two-way communication between them). So hopefully we can shift over the ace example from `combine` to `install` or whatever that ends up becoming once we have more control over child placement (v0.5.0).

paf31:
@jdegoes An alternative to `install`, which I think I prefer, is to just write a function between components and use `combine` internally.

paf31:
^^^ Also, this has the advantage that you get to use a richer internal structure, like a `Maybe Component` or a list of `Component`s.

PR #61, opened 2015-04-04
Add EventHandlerT, fix #57

### LINKED ISSUE #57: Make EventHandler a monad transformer
By using `WriterT`. This will avoid the need for `pure <<< pure` etc.

paf31:
@jdegoes @puffnfresh @garyb @cryogenian Could you please review this before I merge?

The goal was to get rid of `pure <<< pure`, which this does, but `EventHandlerT` is _not_ a `Monad` transformer, but only an `Applicative` transformer, for a technical reason.

Naively, it seems like `MaybeT` with `WriterT` is the right thing to interpret the effects, since we want the ability to both bail out, and accumulate a list of things to do with the event object. However, we need to run things like `preventDefault` _before_ any effects (we don't want to wait for `Aff` to return before we `preventDefault`, which doesn't make sense). 

So the thing we want is 

```purescript
data EventHandlerT m a = EventHandlerT [EventUpdate] (Maybe (m a))
```

which isn't a `Monad` in any useful way, but it is an `Applicative`, which I think is enough to be useful.

I've provided helper functions `liftEventHandler` and `withEventHandler` to be able to use `EventHandlerT` where it is useful.

How does this look?

Aside from docs, which I plan to write now, this wraps up 0.4 I think...

### DIFF
--- a/docs/Halogen-Events.md
+++ b/docs/Halogen-Events.md
@@ -50,7 +50,7 @@ they may be safely embedded in HTML documents.
 #### `input`
 
 ``` purescript
-input :: forall i m a. (Applicative m) => (a -> i) -> a -> EventHandler (m i)
+input :: forall i m a. (Applicative m) => (a -> i) -> a -> EventHandlerT m i
 ```
 
 A helper function which can be used to create simple event handlers.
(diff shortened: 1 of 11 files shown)


jdegoes:
> Naively, it seems like MaybeT with WriterT is the right thing to interpret the effects, since we want the ability to both bail out, and accumulate a list of things to do with the event object. However, we need to run things like preventDefault before any effects (we don't want to wait for Aff to return before we preventDefault, which doesn't make sense).

The above type signature you ended up with makes sense, although I'd suggest that `m (Maybe a)` would allow the emission of `a` to depend on an effect, and would thus be strictly more powerful.

A monad transformer can't really make sense with an arbitrary effect given the need to stop propagation / cancel / etc. *immediately*.

paf31:
Yes, I think that's right. I can try to change this over to `m (Maybe a)`.

docs/Halogen-Events.md
```diff
@@ -64,6 +64,14 @@ This function provides an alternative to making two nested calls to `pure`:
 onclick (input \_ -> Input)
 ```
 
+#### `withEventHandlerT`
+
+``` purescript
+withEventHandlerT :: forall i m e. (e -> EventHandlerT m i) -> e -> EventHandler (m i)
```
jdegoes: Is this a core operation? I'm trying to imagine the use case.

docs/Halogen-Events.md
```diff
@@ -330,10 +338,21 @@ import Control.Functor (($>))
...
 
+#### `EventHandler`
+
+``` purescript
+type EventHandler = EventHandlerT Identity
+```
+
+`EventHandler` is a synonym for `EventHandlerT` applied to the `Identity` monad.
+
+That is, `EventHandler` only adds the `preventDefault`, `stopPropagation`,
```
jdegoes: I don't understand this description. Those functions produce `EventHandlerT`, not `EventHandler`.

docs/Halogen-Events.md
paf31: Well, `Attr` is defined in terms of `EventHandler` (otherwise would need another `m` type argument), and the smart constructors for event handlers (`onclick` etc.) therefore require `EventHandler (m i)`. If you want to use `EventHandlerT`, this is just here as a helper function. Another option is that we change things like `onclick` to use `EventHandlerT`.

docs/Halogen-Events.md
paf31: Yes, this is a bit odd. I'll rewrite.

docs/Halogen-Events.md
jdegoes: That seems like it would be the cleaner approach, right? Then users would never have to wrap like this, whether they want the `T` version or not.

jdegoes:
Good to merge when you're ready.

paf31:
@jdegoes I'll merge this once the CI build finishes, and then that wraps up 0.4 I think. Shall we make a release so that we can update `slamdown` and `slamdata`?

paf31:
Switching to `m (Maybe a)` is slightly trickier than I originally thought, but I hope to have something soon, then we should be able to release 0.4.

paf31:
The one downside with using `EventHandlerT` is that the `m` argument now leaks into the `HTML` type everywhere :-1:

paf31:
I've been trying different things for an hour, trying to come up with a satisfactory to stick `Maybe` under `m` and keep a sane API. So far, nothing pleasant has made itself obvious. The problem is that `HTML` doesn't care what `m` you use (it doesn't mention `m`), but if we go with `EventHandlerT`, we have to have the event handlers mention `m`, which seems wrong. 

This is why I liked things pre-`EventHandlerT` - the code was nicely separated into two parts, but now things are starting to get blurred. I'd like to spend some more time on this before we release 0.4, but it's going to have to be later tonight.

paf31:
One other thing. This might be simpler if `Aff` supported a concept of success with no return value. This started me thinking again about the case where we want a button click to pass multiple values to the signal function asynchronously. It seems like a legitimate use case, but one which `Aff` can't currently handle if my understanding of `Aff` is correct. This is by design, but I still think the `Aff` constraint is too strong. It captures the vast majority of use cases, but I think we need an escape hatch to `Eff` as well. I'd like to spend some time thinking about that for 0.5 or later.

jdegoes:
If you don't think these are good changes, then let's not make them. I do wonder if the use cases would suggest a better solution -- i.e. in which cases will a client choose to cancel the event and produce nothing, and do those suggest a better ADT for EventHandler?

I think instead of `Aff`, we want something like:

```purescript
type AStream e a = ListT (Aff e) a
```

i.e. every `req` produces an asynchronous stream of `res`. The stream may be empty (your "success with no return value") or may contain many elements.

Likely we'd want specific (e.g. parallel) composition semantics for such a thing, so a newtype and instances inside `purescript-aff` could make sense, especially with helpers to create sources, sinks, etc.

paf31:
I think all I need is a `Monad` with some handler like this:

```purescript
class MonadDriver eff m where
  drive :: (a -> Eff eff Unit) -> m a -> Eff eff Unit
```

which captures `Eff eff`, `Aff eff` and `ListT (Aff eff)` among others.

paf31:
> If you don't think these are good changes, then let's not make them.

I definitely think there is _something_ not quite right here, but I don't know what yet, and I don't know the best fix, but I hope to have something better by this evening.

jdegoes:
> I think all I need is a Monad with some handler like this:

This is not sufficient without clear semantics: how many times will you call the callback with a value? Etc. Also the design of this class is extremely imperative: there is nothing declarative about it. 

If the requirement is "a component should be able to emit zero or many outputs for every input", then a simple, declarative model that's easy to reason about is to have the component emit zero or more outputs for every input. :smiley: This approach feels a *lot* cleaner than an ad hoc, imperative type class whose semantics seem to prevent a sane implementation for an asynchronous IO-like monad.

Perhaps the `ListT` part could even be baked in, e.g.:

```purescript
newtype ComponentF p m req res i = ComponentF (SF1 (Either i req) (HTML p (ListT m (Either i res))))
```

This seems straightforward to reason about and directly models the fact that a single input may translate into zero or more outputs, which may carry with them effects `m`.

I'm more arguing to be the devil's advocate than anything else, btw, although I do think there are some legitimate points above. :smiling_imp:

paf31:
Good point. `ListT Aff` does seem to capture exactly what `runUI` needs. On the other hand, it would be nice if the user could choose `m` - if they don't need multiple return values, maybe the type should reflect that.

jdegoes:
Well, they can still choose it if you model it as `ListT m`. Then the outputs would always be zero to many, but the effects would be arbitrary (Eff, free monad, Aff, etc.)

paf31:
Oh sure, I'm saying `MonadDriver` allows them to choose `MaybeT` instead of `ListT`, which says more. Or `IdentityT`.

On a related note, we should dust off `ListT` and get it into core with some docs, and then drop `Data.List.Lazy` from `purescript-lazy`.

paf31:
The imperative intuition is somewhat solidified by the fact that `MonadDriver` is isomorphic to

```purescript
class MonadDriver eff m where
  cont :: m a -> ContT Unit Eff a 
```

jdegoes:
But as we discussed on IRC, the callback can't represent the continuation of your entire program, at least, not without "multiverse semantics".

paf31:
Yes, good point.

jdegoes:
Btw, if we did switch to something like `ListT`, the `Maybe` goes away, right (although it's at a different level)? So maybe this topic is relevant to your current work.

paf31:
Yes, it's definitely relevant. It's what got me thinking about this :)

jdegoes:
Another thing: versus a callback approach, a `ListT` provides the ability to execute an effect even if nothing is emitted. I don't think this is possible with `MonadDriver` (you need an `a` to execute an effect). 

Given that EventHandler needs something like cancel, to prevent emission of anything, whatever we end up with will be either "zero or one", or "zero or greater" (or some generalization which encompasses both of them) -- not "one or greater" or "exactly one". It's just a question of whether that "and zero" lives in the EventHandler and / or at a higher level (e.g. `Component`).

```purescript
-- changes to ListT
data EventHandlerT m a = EventHandlerT [EventUpdate] (ListT m a)

-- m goes into event handler:
newtype ComponentF p m req res i = ComponentF (SF1 (Either i req) (HTML m p (Either i res)))
```

paf31:
I went through using `ListT`, and everything compiles, but I noticed two things:

- `Aff` already supports multiple return values via `makeAff`, but I'm guessing there is an unwritten rule that callers should not use it that way?
- Is the `Monad` instance for `ListT` what we need? `ListT` adds non-determinism, but I want to be able to write

    ```purescript
    action = do
      yield Busy
      async someComputation
      yield NotBusy
    ```

    Is that possible?

jdegoes:
> Aff already supports multiple return values via makeAff, but I'm guessing there is an unwritten rule that callers should not use it that way?

Well, if you use it that way, I can't tell you what it means if `a -> Eff eff Unit` represents the rest of your program. I should probably add a note, and document the FFI structure of `Aff` so that you don't have to use `makeAff` for FFI code (with its many, many layers of wrapping).

> Is that possible?

No, it's unfortunately not possible because the `Monad` allows you to change the type. One should be able to write a `(>>=)` like construct that does what you want, but keeps the type fixed, then using rebindable `(>>=`) syntax, one could still use do notation for this. That'd make a good addition to `ListT`.

PR #64, opened 2015-04-07
Use ListT in Event monad instead of MaybeT in Handler functor

### LINKED ISSUE #57: Make EventHandler a monad transformer

paf31:
Reverts slamdata/purescript-halogen#61

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,11 +85,11 @@ With our signal function model, it is now simple to support things like AJAX whi
 type UI4 m input = SF input (HTML (m input)) 
 ```
 
-This allows us to create documents which generate inputs in some effectful way. In practice, Halogen requires `m` to be the `Aff` monad, defined in the `purescript-aff` library. `Aff` allows us to define asynchronous actions, and provides a simple AJAX API in the `affjax` library.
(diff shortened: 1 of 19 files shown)


paf31:
I'll extend this PR with the `ListT` approaches.

paf31:
@jdegoes I think this is as good as it's going to get. It's not quite what I was after, but it supports the use cases we need.

I tried using something like `FreeT` to define a `Producer` monad, which worked nicely, but the types didn't fit well with `Component`, so I ditched it.

As a compromise, I added the `andThen` combinator, which allows us to use `ListT` a bit more like `FreeT`.

Let me know what you think.

jdegoes:
This looks good to me. The specialization to `ListT (Aff eff)` definitely reduced the complexity of many type signatures, and allows us to get a (potentially effectful) `cancel` for "free" (`pure empty`).

I know you didn't want to prematurely specialize to `Aff` but leave the door open to other approaches. This seems to model the domain very well (zero or more outputs for every input), and has clear and obvious semantics (asynchronous possibly empty stream of outputs), but is certainly less general-purpose than a `MonadDriver` approach. What are your thoughts?

paf31:
I think it works nicely. In the end, it doesn't matter too much, since we can map a monad morphism over a component anyway, so if the user wants to use a different monad, they can, but I think this gives a nice default.

jdegoes:
:shipit:

PR #68, opened 2015-04-12
Fix #63, support data attributes.

### LINKED ISSUE #63: data attributes
there is no way to set them. 
```purescript
attr (attributeName "data-foo") "bar" 
```
doesn't work

paf31:
@jdegoes @cryogenian @garyb Could you please review?

Details here:

https://github.com/Matt-Esch/virtual-dom/blob/7cd99a160f8d7c9953e71e0b26a740dae40e55fc/docs/vnode.md#custom-attributes-data-

### DIFF
--- a/src/Halogen/Internal/VirtualDOM.purs
+++ b/src/Halogen/Internal/VirtualDOM.purs
@@ -43,7 +43,13 @@ data Widget (eff :: # !) i
 foreign import emptyProps 
   "var emptyProps = {}" :: Props
 
--- | Update a set of mutable properties by specifying a key/value pair
+-- | Create a property from a key/value pair.
+-- | 
+-- | Properties named `data-*` will be added as attributes, and any other properties will
+-- | be set directly.
+-- |
+-- | Users should use caution when creating custom attributes, and understand how they will
+-- | be added to the DOM here.
 foreign import prop
   "function prop(key, value) {\
   \  var props = {};\
@@ -65,7 +71,7 @@ foreign import handlerProp
   \  Hook.prototype.unhook = function(node) {\
   \    node.removeEventListener(key, this.callback);\
   \  };\
-  \  props['data-halogen-hook-' + key] = new Hook(f);\
+  \  props['halogen-hook-' + key] = new Hook(f);\
   \  return props;\
   \}" :: forall eff event. Fn2 String (event -> Eff eff Unit) Props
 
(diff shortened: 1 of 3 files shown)


garyb:
Is it possible for `attributes` to be set in `props`? If so, that may overwrite `data-` properties set this way - we'd need need to merge a `props.attributes` and `props1.attributes` instead.

src/Halogen/Internal/VirtualDOM.purs
```diff
@@ -128,7 +128,19 @@ foreign import vnode
   \  return function(props) {\
   \    return function(children) {\
   \      var VirtualNode = require('virtual-dom/vnode/vnode');\
-  \      return new VirtualNode(name, props, children);\
+  \      var props1 = {\
+  \        attributes: {}\
+  \      };\
+  \      for (var key in props) {\
```
garyb: We might want to iterate over `Object.keys(props)` here instead, as then you don't need the `hasOwnProperty` call for every key.

paf31:
The only way that's possible right now is if the user creates a custom attribute, of type `{ attributes: _ }`, which I consider to be a misuse anyway. 

I actually tried a different approach in which the `props` constructor branched depending on the key, but then the `combineProps` function got a bit confusing. 

I guess it depends what we want the `Props` type to represent. If it's just a map for internal use, I think it's probably fine, as long as people stick to the API. If we want it to represent valid `virtual-dom` property objects, ready to be passed to `vnode`, then this obviously won't work.

src/Halogen/Internal/VirtualDOM.purs
paf31: That would need a polyfill for ES3 right? Which is fine obviously, but would need documenting.

garyb: Yep. I thought we were kind of heading that way for all PS code anyway. I know there's ES5 stuff that crept into `Array` already at least.

paf31: Ok, sounds good.

paf31: So would that be `Object.keys(props).forEach(...)` then?

garyb: For optimum speed it'd still be a for loop, just looping over the array of extracted keys instead. `forEach` is probably plenty fast enough though.

jdegoes:
Sorry, didn't have a chance to review this (was scheduled for Monday), get sorted out all right???

paf31:
Yep, @garyb and I went over it on IRC. I think this is probably the best approach for now, but we can revisit it if necessary.

PR #69, opened 2015-04-13
Fix #53, support component initialization

### LINKED ISSUE #53: Support Component initialization

paf31:
@garyb @natefaubion @jdegoes Could you please review?

This supports initialization by adding an init hook, and a new type of `Attr`.

There is also an example in `examples/ajax`.

### DIFF
--- a/docs/Halogen-Events.md
+++ b/docs/Halogen-Events.md
@@ -506,6 +506,13 @@ instance monadEvent :: Monad (Event eff)
 ```
 
 
+#### `monadEffEvent`
+
+``` purescript
+instance monadEffEvent :: MonadEff eff (Event eff)
+```
+
+
 #### `monadAffEvent`
 
 ``` purescript
--- a/docs/Halogen-HTML.md
+++ b/docs/Halogen-HTML.md
@@ -1825,7 +1825,8 @@ the `Attr` type.
 data Attr i
   = Attr (Exists AttrF)
   | Handler (ExistsR (HandlerF i))
-  | Initializer (Boolean -> i)
+  | Initializer i
+  | Finalizer i
 ```
 
 A single attribute is either
@@ -1859,13 +1860,18 @@ Create an event handler
 #### `initializer`
 
 ``` purescript
-initializer :: forall i. (Boolean -> i) -> Attr i
+initializer :: forall i. i -> Attr i
 ```
 
-Create an initializer function.
+Attach an initializer.
 
-The initializer will receive a Boolean value which indicates whether the component is being
-initialized for the first time, or just rerendered.
+#### `finalizer`
+
+``` purescript
+finalizer :: forall i. i -> Attr i
+```
+
+Attach a finalizer.
 
 #### `ClassName`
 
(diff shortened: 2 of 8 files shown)


garyb:
:+1: 

My only question would be, do you think having the "is-init" `Boolean` as the argument is worth having, instead of just filtering out the updates? It's sort of leaking `virtual-dom` internals to `Attr`, as for other renderers it'd we'd probably have to go out of our way to make it ever fire in the non-init case.

paf31:
@garyb Yes I think you're right. I don't see any use case for an "update hook". Anyone? Otherwise, I can change this and merge it in.

cryogenian:
Firing event after component rerender can be very useful for communication with outside libs. I.e. chart rerender or something.

cryogenian:
Or maybe it would be better to add additional hook for rerendering and filter initializing hook to only first run

garyb:
I think that would be preferrable.

When would the update hook fire also? I'm not sure the default behaviour of virtual dom retriggering the hook is necessarily the same thing as what we'd want here.

cryogenian:
I think update hook should be fired after rerender element just like in vdom. 
Moreover there should be something like `unhook` or `componentDismounted`.

garyb:
I thought `hook` only fires if the value associated with it changes?

cryogenian:
Don't know precisely about it, I thought that they fire after rerender.

cryogenian:
found this https://github.com/Matt-Esch/virtual-dom/blob/master/docs/hooks.md

paf31:
Yes, that's the doc I've been working from.

jdegoes:
It seems to me we need:

 * On initialization (a one-time event)
 * On (re) render (occurs potentially multiple times)
 * On destruction (a one-time event)

Anything else?

paf31:
Just curious - I (sort of) see the need for a initialization hook, but why do we need to have hooks for updates and destruction? It seems to go against the abstraction we've built a little. When would you need to know that a pure component (not a widget) was rerendered?

jdegoes:
@paf31 Well, what if you wanted to fire off an event to persist some data on destruction? 

I'm not sure about re-rendering, but I assume it would be useful in some circumstances if your pure component embeds an impure one. @cryogenian / @garyb ?

paf31:
:+1: for destruction, but I think I'd prefer that users use widgets if they need update notifications. The only use case I can think of is to sync a widget's state with the state of the component. Pure components should use virtual-dom to operate at a higher level of abstraction than individual updates IMO.

jdegoes:
:+1: Sounds good to me, pending comments from everyone else.

garyb:
I missed the component/widget distinction originally - I'd agree with @paf31 that you'd probably only need update/destruction in widgets.

cryogenian:
The use case of that rerender hook is reuse not widget js components (i.e. bootstrap's tooltip, tab, popover etc).
We definitely can provide all this functionallity by halogen (or bootstrap-halogen) but there is much more other libs, that can be used without adding any code.

paf31:
Wouldn't all of those be the init hook though?

cryogenian:
Yes, all of those can be the init hook :smile: 
I have no other comments then

PR #75, opened 2015-04-14
Adjust case of events

### LINKED ISSUE #74: Case consistency for events
In `Halogen.HTML.Events` the event names are all lowercase, `onclick`, `onmouseover` and so on, in `Halogen.HTML.Events.Forms` they are camel case, `onInput`, `onValueChanged`, etc. I guess we should make these all consistent. Personally I much prefer the camelcase versions.

Any thoughts before I open a PR for this?

garyb:
Also dropped `onInput` from the non-form events, after discussing with @paf31 in IRC (it conflicted with a more useful version in `Events.Forms`)

Resolves #74


paf31:
:+1:

jdegoes:
Looks good to me, but I suggest we *not* make a new release until all breaking changes for 0.5 are in. We could even push this to a 0.5 branch if we think there will be more (non-breaking) changes to 0.4 as part of the 0.5 work. Thoughts?

garyb:
A 0.5 branch makes sense to me - the PR I opened for bumping `purescript-aff` should be a breaking change also, so that could be merged in there too.

PR #76, opened 2015-04-15
Bump aff dependency

garyb:


jdegoes:
Yeah let's push this one to 0.5, too, and when 0.5 is ready we can merge back to main.

PR #81, opened 2015-04-17
Fix #65, version inputs to widget components

### LINKED ISSUE #65: Widgets receive inputs multiple times
In the Ace example, if you copy text, then click the Copy button, then copy text again, the text area clears, which is wrong. I think this is something to do with `init` vs `update`.

paf31:
@jdegoes @garyb @puffnfresh Could you please review?

After several attempts, this is the only thing I could come up with which works. The idea is to version each input to the component by pairing the inputs with a counter. Inside the widget, if the version has not been bumped, we don't invoke the `update` action during the diff.

I'm not exactly proud of this fix, but I think it's about as good as it's going to get without an effectful signal function (unless there are other ideas)...

Thanks.


jdegoes:
There's no need for an effectful signal function. A widget could, for all intent and purposes, be an ordinary HTML element which supports adding event listeners that generate events ultimately producing `i`'s; that is, there is no information theoretic means by which a user could tell the difference between an artificial element that generates artificial events, and a real DOM element that generates real events.

```html
<div></div>
```

versus:

```html
<ace-editor></ace-editor>
```

The differences are only superficial and due to the way virtual-dom models widgets (which we inherit).

I'm convinced that a *different* way of handling element / widget creation and diffing for widgets can solve these problems and expose a uniform interface, one which is suitably compatible with the existing definition of `Component`. 

I'll try to sketch that up if I get a chance, in the meantime, is this production ready? We're going to start using real widgets next week so I'm hoping the answer is yes. :smiley:

paf31:
After my short testing, it seems good to go, but the only example I have right now is my mini Ace example.

The issue isn't with events on the output side. Those _do_ behave like pure components. The problem is when we want events to behave like inputs, but all we have is a function `input -> Eff _ _`. We can partially apply this and embed the `Eff` action into the `update` part of the widget, but a pure signal function captures the same `input` even when there is nothing to update. This change basically makes the update function only fire the first time it is called after the last input.

paf31:
Also, I'm not convinced there is a better way to solve this problem with a different vdom implementation. The issue is because of the way we want to handle inputs. At least if we reimplement something like virtual-dom, we can give descriptive types, like how `patch` depends on the effects in our widget's `update` function.

PR #82, opened 2015-04-20
Fix #52, add install and hide combinators

### LINKED ISSUE #52: Lift "graft" to the level of components

paf31:
@jdegoes Could you please review?

With this approach, we avoid the pi-types issue by assuming the placeholder type is something small and finite.


jdegoes:
I'm good merging this in, at least on an experimental basis until we figure all this stuff out.

jdegoes:
^^^ @paf31

PR #83, opened 2015-04-20
Squashed commit of the following:

cryogenian:
There are properties that must be setted as `attribute` (I found `readonly`, maybe there are some other)

### DIFF
--- a/src/Halogen/Internal/VirtualDOM.purs
+++ b/src/Halogen/Internal/VirtualDOM.purs
@@ -154,6 +154,7 @@ foreign import vtext
   \  return new VText(s);\
   \}" :: String -> VTree
 
+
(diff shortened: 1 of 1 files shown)


jdegoes:
Review / merge by @paf31 or @garyb

paf31:
It seems like we probably want to have the data structure represent attributes vs. properties, but the irritating thing is that such a distinction doesn't necessarily belong in the `HTML` type. It would be nice if `virtual-dom` just handled this sort of thing automatically. I'm not certain which things are properties and which are attributes, but I'd be willing to bet there are more changes needed, and I wouldn't want to include them all as special cases. Maybe we need an adapter layer which presents everything as a property?

Reference: https://github.com/Matt-Esch/virtual-dom/blob/master/docs/vnode.md

Issue #87, opened 2015-04-21
Emit input signals from initializer 


cryogenian:
https://gist.github.com/cryogenian/5821dbed7f910194796a

This is slightly modified counter example. I've used **0.5** branch. 

* It compiles, but for every initializer that hasn't last sentence empty it produces unhandled error.
* Initializers don't send any messages to `update` function. It starts with `0` and after it reaches `6` there is no jump to `7`.

paf31:
What is the unhandled error?

paf31:
Also, you're using `yield` incorrectly. `yield` is basically just `pure` because of the way `Event` uses `ListT`. I should probably remove it, since the name is confusing.

cryogenian:
> Uncaught error in asynchronous code: Failed pattern match

So, is there no way to emit two messages from event handler and event monad? only one or zero?

paf31:
No you can: ```yield a `andThen` yield b```

cryogenian:
ok, good to know

paf31:
The pattern match failure is a problem though. Are you able to figure out where it happened?

cryogenian:
Nope, there is lot of `Aff` code, one `hook` and `createElement` in stack. But no any place with `throw new Exception()` :disappointed: 
I suggest that it is in the `runEvent` but I'm not sure

paf31:
The issue is here:

https://github.com/slamdata/purescript-halogen/blob/master/src/Halogen.purs#L110

paf31:
Basically, I think you're trying to cause a rerender during the initial `createElement`, and the node hasn't been created yet. I didn't think this was possible, but there you go. Never use a partial function, I guess.

cryogenian:
Huh 0_o

cryogenian:
Maybe it's possible to decouple message queue and rerender queue.

jdegoes:
We should find some way of solving this that is "safe by construction".

PR #88, opened 2015-04-23
Patch for #87

paf31:
@jdegoes This is a band-aid for now. Could you please review?

A correct-by-construction approach is in the works.

_Edit_ : I'm not sure my approach will work. We basically want to ensure that the action associated with the initializer doesn't return immediately, but is asynchronous, so that we don't try to rerender during the first render. I thought of using an indexed monad with the new custom do notation feature, but that won't work, since even if we label `async` and `sync`, there's no way to guarantee an `Aff` action wasn't constructed using, say, `pure`.

### DIFF
--- a/src/Halogen.purs
+++ b/src/Halogen.purs
@@ -107,10 +107,13 @@ runUI' ref sf = do
   
   driver :: Driver (Either i req) eff
   driver e = do
-    Just { signal: signal, node: node } <- readRef ref
-    let next = runSF signal e
-    node' <- patch (head next) node
-    writeRef ref $ Just { signal: tail next, node: node' }
+    ms <- readRef ref
+    case ms of
+      Just { signal: signal, node: node } -> do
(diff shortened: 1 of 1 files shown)


jdegoes:
We could add a phantom type parameter to `Aff` to indicate whether or not the action is asynchronous. The problem with that is it inhibits composition, anytime you want to compose a sync with an async, you have to manually `async` the `Aff` to switch over the type param.

You can use `later` to make a sync action asynchronous, which may be a better way of solving this.

jdegoes:
@paf31 ^^^

paf31:
Sure, that sounds like a good way to address this.

jdegoes:
I assume you want a merge on this, right?

paf31:
Yes please if that's ok. I'll add a call to `later` in a separate PR if we want to go that route, but that seems like the simplest safe-by-construction approach.

Issue #85, opened 2015-04-20
Finalizer and Initializer elements


cryogenian:
Is there a way to get a target element in initializer?

paf31:
No, but I think it's simple to add that. What do you need it for out of curiosity?

cryogenian:
Those bootstrap `tooltip` and other third-party things (i.e. zeroclipboard)

paf31:
I'm generally not a fan of accessing DOM nodes in event handlers since we don't have a good way to ensure that access is read-only right now, but I guess we don't have a good alternative :(

PR #89, opened 2015-04-23
Provide HTML element to initializer and finalizer

paf31:
@jdegoes A not-very-safe fix for #85. We probably want to limit the `HTMLElement` to read-only access in the event handler. This will let us get the job done for now, but I think I would like to implement this using a fixed set of allowed primitives once there is more time. If we merge this maybe we can create a new issue for that. What do you think?

### DIFF
--- a/docs/Halogen-HTML.md
+++ b/docs/Halogen-HTML.md
@@ -1833,8 +1833,8 @@ the `Attr` type.
 data Attr i
   = Attr (Exists AttrF)
   | Handler (ExistsR (HandlerF i))
-  | Initializer i
-  | Finalizer i
+  | Initializer (HTMLElement -> i)
+  | Finalizer (HTMLElement -> i)
 ```
 
 A single attribute is either
@@ -1868,7 +1868,7 @@ Create an event handler
 #### `initializer`
 
 ``` purescript
-initializer :: forall i. i -> Attr i
+initializer :: forall i. (HTMLElement -> i) -> Attr i
 ```
 
 Attach an initializer.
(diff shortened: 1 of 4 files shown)


garyb:
How would you limit it to being read-only? Also isn't the purpose of this to use the element while initialising 3rd party components, which inherently will need to modify the element somehow?

:+1: for the fix for now though.

cryogenian:
Thanks! I agree with @garyb, the main purpose of initializer is to modify element somehow (at least it hidden state) not only to modify `State` because modifying `State` on init is absolutely the same as providing already modified state at start.

paf31:
@garyb I was thinking of only allowing access to things like `getProp`, `getAttr`, etc., but it seems that isn't going to work for this use case.

Question though - why not use `widget` instead if you need to set up the DOM node on intialization - that's what it's there for.

jdegoes:
> Question though - why not use widget instead if you need to set up the DOM node on intialization - that's what it's there for.

Yeah if we can somehow simplify this by using widget for these use cases, could be an architectural win.

paf31:
I'll leave this open for now then, in case we need it, but if we can use `widget` I think this should be closed.

jdegoes:
If widget can already do initialization / finalization, can we remove those from the Attr ADT???

paf31:
I suspect it is still useful to know when a particular node is rendered, even if you don't get to use the `Node` itself. Forcing users to use `widget` might be a bit too much when the subcomponent is quite large.

jdegoes:
Well, could combinators be supplied that build the same functionality on top of `widget` (and make it super easy like the Attr constructors), just so there's one fewer way of doing this?

paf31:
The case I'm thinking of is where you have a large _pure_ HTML node, and you want to know when it gets rendered for the first time. I can't think of any good way to do the same thing with a widget. You would have to replace the entire subtree with a widget.

jdegoes:
OK.

paf31:
@cryogenian Were you able to find a way to use `widget`? Should I close this?

cryogenian:
Yes, I think you should close it

Issue #94, opened 2015-04-25
Examples broken after bootstrap/foundation changes


paf31:
The examples can be built by installing `purescript-halogen-bootstrap` using Bower, but that leads to a cyclic dependency issue. We can either just add detailed instructions, or move each project into its own `bower.json` file. Maybe if we did the latter, we could also move the examples over to `psc-make`.

PR #95, opened 2015-04-27
Pulpify all examples

paf31:
This change makes it a bit easier to manage dependencies in the example projects, and also speeds up their builds, by moving each one into its own project, with its own `bower.json`. This also lets us move the `devDependency` on `purescript-ace` into the `ace/` directory.

I'd like to eventually add a `purescript-echarts` example, so I made this change to avoid any more dev dependencies.

@jdegoes Does this look ok?

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -31,8 +31,5 @@
     "purescript-exists": "~0.1.1",
     "purescript-lists": "~0.6.0",
     "purescript-integers": "~0.1.0"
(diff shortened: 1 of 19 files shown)


paf31:
Resolves #94.

garyb:
:+1:

jdegoes:
:+1:

PR #96, opened 2015-04-27
Move versioning logic out of 'widget'

paf31:
This change moves the versioning code, needed to fix the "multiple inputs" issue, out of `VirtualDOM.Internal` and into the `Component` module, the only place where it is used.

This should make it easier for widgets to implement their own custom diffing code. Now each widget contains:

- Its "context" `ctx`, created during `init` and passed to `update`.
- Its "value" `val`. In the `Component` module the value is the version, an `Int`, and we only update if the version increased. In general though, this could be used for more fine-grained diffing.

The high-level API hasn't changed apart from to rename `state` to `context`, since the state never really changed.

@jdegoes @garyb Could you please review?

### DIFF
--- a/docs/Halogen-Component.md
+++ b/docs/Halogen-Component.md
@@ -54,7 +54,7 @@ A variant of `component` which creates a component with some internal, hidden in
 #### `widget`
 
 ``` purescript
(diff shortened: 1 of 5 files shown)


garyb:
:+1:

PR #104, opened 2015-05-02
EQ, Semigroup and Show instances or URL

passy:
I don't know of this is something you'd like to have for simple wrappers.

By the way, is something like `NewtypeDeriving` somewhere on the roadmap for PS? :)

### DIFF
--- a/src/Halogen/HTML/Target.purs
+++ b/src/Halogen/HTML/Target.purs
@@ -23,6 +23,16 @@ import qualified Halogen.HTML.Events.Handler as E
 -- | A type-safe wrapper for a URL
 data URL = URL String
 
+instance eqURL :: Eq URL where
+  (==) (URL a) (URL b) = a == b
+  (/=) a       b       = not (a == b)
+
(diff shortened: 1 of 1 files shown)


garyb:
:+1:

paf31:
:+1:

jdegoes:
Thanks @passy!

jdegoes:
Ultimately it would be nice to have `purescript-url` which has URL parser, uses Pathy's `Path` type for the path component, represents query string parameters using `StrMap String` or something similar, etc. This `URL` type can contain a lot of things which have no relation to URLs. :smile: Also occurs to me it should probably be a `newtype` rather than a `data` declaration -- if someone wants to change it the next time they're working in that file.

garyb:
I've made a start on that: https://github.com/garyb/purescript-uri but it doesn't break things up that finely yet.

Happy to move that repo to `slamdata` also, if you like?

jdegoes:
[Yes, please!](https://github.com/slamdata/purescript-uri)

Issue #84, opened 2015-04-20
more component discussion


jdegoes:
I spent part of the weekend convincing myself the types for component are wrong. Components are a black hole, it's not possible to interrogate them (for example, asking an Ace component what the current contents of the buffer are). While components emit declarative information in response to every input, that's not useful for communication; they cannot yield event information except in response to user input.

I think the signature I'm really looking for is something like this:

```purescript
newtype Component p m i req res = Component (SF1 (Either i req) (Tuple (HTML p (m i)) (m res)))

type ComponentW p m req res = ComponentW (exists i. Component p m i req res)
```

This is possibly not exactly right, one can imagine wanting the `Tuple` to be an `Either`, but that's not precise enough either and allows a component to refuse to emit `HTML`. OTOH encoding more strictly reduces the power (e.g. prevents emitting `res` from user input or prevents updating HTML in response to requests).

puffnfresh:
For people still new, like me, the current Component looks like:

```purescript
newtype ComponentF p m req res i = ComponentF (SF1 (Either i req) (HTML p (m (Either i res))))

newtype Component p m req res = Component (Exists (ComponentF p m req res))
```

paf31:
> Components are a black hole, it's not possible to interrogate them

I thought that was by design - you don't get to interrogate, only observe what messages are exposed.

If you use `Tuple` as in your first example, you won't be able to send non-external messages on user input, and if you use `Either`, you don't necessarily ever get a document to render.

What's the problem we're trying to solve?

paf31:
I do think there is a problem with `Component` though, which is that they aren't as composable as raw signal functions. I think the existential gets in the way. It makes it hard to have things like lists of components, because you can't construct the appropriate existentially hidden type in any sensible way to combine them.

I think a good step would be to make the internal message type into a type variable, instead of quantifying it. Then we could write combinators like

```purescript
list :: forall p m i res req. List (Component p m i res req) -> Component p m (Tuple Int i) (Tuple Int res) (Tuple Int req)
``` 

Either that, or just get rid of the `Component` abstraction entirely, since at that point we're basically back to a newtype on top of raw signal functions anyway.

jdegoes:
> What's the problem we're trying to solve?

[...]

Issue #99, opened 2015-04-28
Remove Component API?


paf31:
This has been mentioned a few times, so I'll put this issue here so that we can discuss it properly.

Most applications I've seen (other than the Ace example) use one large component, and other techniques (`Inject1`...) to deal with large sum types. Maybe `Component` is not worth the drop in composability?

garyb:
To be honest, I couldn't figure out how `install` was supposed to work, that's part of the reason I ended up with the `Inject1` stuff. I still wonder if maybe it would be cleaner than the inject approach...

Since I haven't taken advantage of components right now, I'm fine with getting rid of them, but I feel I should at least make another attempt at figuring out how the composition of them *should* work.

paf31:
The idea with `install` is to replace an existing placeholder with another `Component`. On the `HTML` level, the corresponding thing (`graft`) is monadic:

```purescript
graft :: forall a b res. HTML a res -> HTML (a -> HTML b res) -> HTML b res
```

grafting trees onto leaves (placeholders) of the tree. But with `Component`, we can't do that, because the existentially-chosen type in `HTML b res` on the right of the inner function arrow _depends on_ the value of type `a`. That is, we need dependent types to properly describe the internal message type of the result.

So, instead, we treat the placeholder type like a finite sum, using `Maybe a` to represent "one more value than `a`". When you graft, you use up one position, so `Maybe a` becomes `a`.

This works fine for static, finite things, but won't work if you want to install a list of components. In that case, you will probably have to use `runComponent` manually, or use `combine` with a fold, essentially creating a data structure from nested `Either`s.

paf31:
@cryogenian What is your experience with components so far?

cryogenian:
I haven't use them. But I will if I start project using halogen. At least will try :smile:

puffnfresh:
I managed to use `install` today. I took me all morning to get things to line up:

https://github.com/slamdata/slamdata/pull/67

PR #105, opened 2015-05-02
Component is now just a type synonym

paf31:
@garyb @jdegoes @puffnfresh @cryogenian Could you please review?

This doesn't _remove_ `Component`, as I originally suggested, but turns it into a type synonym (minus the existential `i` part), and keeps some of the functions around, since they are still useful.

Things worth noticing:

- The examples didn't change much, mostly just removing calls to `component`
- The type of `install` has changed. Making the installed component dependent on the placeholder value means we have to have `HTML` on the RHS, not a whole `Component`, but I think this is okay, since the function will close over the state of the outside component in practice. 

I think it's worth thinking about putting together a library of isomorphisms (`Either`, `Tuple` and `Maybe` should do it) to deal with the few more `Either`s we'll see as a result of ditching the existential.


paf31:
Resolves #84, #99.

cryogenian:
I like it. Code is simpler, probably more performant and this simplification doesn't remove any important functionality. :+1:

paf31:
I started a bit of work on an isomorphisms library, and I'll work on it some more this weekend:

https://github.com/paf31/purescript-isomorphisms/

puffnfresh:
The new `install` definitely seems a lot more useful. Hopefully it will fix my blocker.

src/Halogen/Component.purs
```diff
@@ -183,25 +124,7 @@ combine :: forall p q r m req1 req2 res1 res2.
              Component p m req1 res1 -> 
              Component q m req2 res2 -> 
              Component r m (Either req1 req2) (Either res1 res2)
-combine f = runComponent \sf1 -> runComponent \sf2 -> component' (mergeWith' f1 f2 sf1 sf2)
+combine f = mergeWith f1
```
puffnfresh: This is great. The previous stuff was really bad :fireworks:

puffnfresh:
This cleans things up dramatically. I'm not sure if my blocker will be fixed, but :shipit:

jdegoes:
Big improvement. :+1:

Issue #100, opened 2015-04-29
Ace integration


paf31:
@jdegoes Do you have a list of the features you will need for the initial version of `slamdata`?

Should we create a `halogen-ace` repo?

paf31:
Duplicate of #102.

Issue #98, opened 2015-04-28
Generate events


cryogenian:
How do you think is it possible to add pseudoattributes for `HTML` like `clicked`, `focused` etc? And if it setted to `true` generate an event. Can be helpful to change focus or select texts.

paf31:
It should be fairly simple to implement these in terms of `init`, no?

cryogenian:
Erm... Could you give a clue how to do it?

paf31:
Something like

```purescript
H.div [ A.init if shouldClick then yield ThingToDoOnClick else empty ] []
```

?

cryogenian:
Do you mean `initializer`? I can't find `A.init`

paf31:
Yes, sorry.

cryogenian:
I mean not to do something that we do when events are fired, but to fire those events. In example `e.focus()` or `e.select()`. I'm afraid it can be done only by `Widget`'s `init` not `A.initializer`.

paf31:
Sorry, I was thinking of the jQuery-style, where `el.click()` means "generate a click event", and assumed you just wanted to generate the input to the state machine without a click from the user. Maybe we do need #89 after all...

paf31:
Can we do this with hooks?

paf31:
@jdegoes @cryogenian I've put an (untested) sketch of a possible solution on the `hooks` branch. I think we can use this to define custom attributes like `focus`. I don't have time to work on this any more tonight, but I can take a look tomorrow if this approach looks ok.

paf31:
Here's a thought. What if you used a data attribute, and we did an effectful pass over the DOM after each render, to apply any custom attributes, using jQuery or `simple-dom`? I could add a hook to `runUI`, and it would probably be cleaner than hook attributes.

garyb:
That might be an option, but manipulating the DOM outside of virtual dom can really be asking for trouble when you're not doing it with hooks, widgets, etc. Or is that what you meant by using the data attribute? Use that as a marker for "make change here"?

paf31:
> Or is that what you meant by using the data attribute? 

Pretty much. The case I was thinking of was focussing a text box. We could set a `data-should-focus` attribute, and then use jQuery to focus or blur that element if necessary after rendering.

We could do the same thing with a hook in an attribute, defining

```purescript
focussed :: Boolean -> Attr
focussed = attr "halogen-hook-focussed" <<< makeFocusHook

[...]

PR #106, opened 2015-05-03
Remove Widget, add post-render hooks

paf31:
/cc @jdegoes @cryogenian @garyb @puffnfresh

This PR is a bit more drastic. @garyb and I were discussing widgets, and this demonstrates an idea we were considering - since widgets are effectful, instead of trying to shoehorn them into the pure world of `Component`, it makes sense to provide a low-level, imperative escape hatch to deal with them. It's not pretty, but widgets aren't either, and it fits with the "push effects to the application boundary" approach. 

Namely, I've removed `Widget`, and added support for a "post-render hook" to `runUIWith`. Widgets can be synchronized in the post-render method after every patch is rendered.

The Ace example shows an example of using `data-` attributes to synchronize the Ace widget state with the state of the component. 

@jdegoes I think this is more in line with the "big record of widget state" approach that we discussed in #84.


paf31:
Also, if we go this route, I don't think we need to have a separate `halogen-echarts` repo, since we will just be using the ECharts API directly.

paf31:
Resolves #98, since we can create `data-` attributes to represent the focus state.
Resolves #100.

docs/Halogen-HTML-Renderer.md
```diff
@@ -5,7 +5,7 @@
 #### `renderHTML`
 
 ``` purescript
-renderHTML :: forall p i eff. (i -> Eff eff Unit) -> (p -> Widget eff i) -> H.HTML p i -> VTree
+renderHTML :: forall i eff. (i -> Eff eff Unit) -> H.HTML Void i -> VTree
```
puffnfresh: :+1:

docs/Halogen.md
```diff
@@ -60,7 +60,7 @@ main = do
 #### `runUI`
 
 ``` purescript
-runUI :: forall req eff. Component (Widget (HalogenEffects eff) req) (Event (HalogenEffects eff)) req req -> Eff (HalogenEffects eff) (Tuple HTMLElement (Driver req eff))
+runUI :: forall req eff. Component Void (Event (HalogenEffects eff)) req req -> Eff (HalogenEffects eff) (Tuple HTMLElement (Driver req eff))
```
puffnfresh: Nice! :+1:

puffnfresh:
Brilliant. I think these updates will unblock me with integration of Ace into SlamData. Thank you!

README.md
```diff
@@ -100,17 +100,15 @@ See the AJAX example for a demonstration of the `Event` monad.
...
 
-In practice, Halogen chooses `p` to be the type `Widget eff input`, which describes a `virtual-dom` widget, but as with the type argument `m`, leaving things polymorphic has the advantage that we can choose to interpret our model in a different way for testing.
+The `runUIWith` function accepts an extra parameter, a function which will be called whenever the DOM is updated. By using this function in conjunction with custom data attributes, it is possible to synchronize the internal state of a third-party component with your component's internal state.
 
-The `p` type parameter also gives us a way to "graft" HTML documents at a point which is marked by a placeholder. We can think of the type `HTML p a` as describing HTML documents with "holes" marked with values of type `p`.
+See the `examples/ace` directory for a worked example.
 
-See the Ace editor example for a demonstration of placeholders.
+Note: this is considered an advanced use case - incorrect use of this feature could lead to unexpected behavior, so it is recommended that you understand the rendering pipeline before using post-render hooks.
```
jdegoes: This comment makes me sad. :frowning: Is the interface as safe as can be, or are there additional ways we can improve its safety?

jdegoes:
> @jdegoes I think this is more in line with the "big record of widget state" approach that we discussed in #84.

I can't help but feel this approach will implode at some point in the very near term (Ace with a large enough buffer, a Canvas-based paint component, a chart with a lot of data points, etc.). But overall I think these changes are a positive direction from where we are right now, so in the spirit of making incremental progress. :+1: .

jdegoes:
@paf31 The other thing: there's no need for `p` anymore, is it time to remove that to simplify `HTML` and `Component`?

paf31:
My original PR also removed `p`, but I decided to change things in stages because it broke the downstream libraries, and I also wasn't sure if we wanted to keep `install` etc.

If we want to remove `p`, I can work on that today.

garyb:
:+1: for removing `p`.

PR #107, opened 2015-05-03
Remove placeholders

paf31:
/cc @jdegoes @puffnfresh @garyb @cryogenian

Ok, this is ready for review.


paf31:
Note that even though `install` has gone, we can do the same thing with functions anyway:

```purescript
componentWithAHole :: forall m i. H.HTML i -> Component m i i
``` 

etc.

jdegoes:
:+1:

PR #114, opened 2015-05-06
Refactoring

paf31:
/cc @puffnfresh @jdegoes 

I broke apart `runUIWith` into `componentProcess` and `mainLoop` in the hope that it would expose some structure we could exploit to build widgets. At least now we can use `SF` and `Process` to build the top level functions which update things like Ace and ECharts.

Thoughts?

### DIFF
--- a/docs/Halogen.md
+++ b/docs/Halogen.md
@@ -82,5 +82,27 @@ each update.
 This is considered an advanced feature, and should only be used with an understanding of
 the rendering pipeline.
 
+#### `Process`
+
+``` purescript
+type Process req eff = SF req (Eff (HalogenEffects eff) HTMLElement)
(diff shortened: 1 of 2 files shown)


puffnfresh:
Seems like a good place to start :+1:

jdegoes:
Can you explain the simplest way to listen to all events?? It's still not clear to me. :frowning:

paf31:
Sure - the function passed to [`runUIWith`](https://github.com/slamdata/purescript-halogen/blob/master/docs/Halogen.md#runuiwith) now takes the input which caused any rerendering to happen.

jdegoes:
Ah, I missed that, great we'll give it a go!

PR #117, opened 2015-05-08
Remove pure and unsafeInterleaveEff

puffnfresh:
This were not being used to do anything useful.


garyb:
:+1:

paf31:
I assume you'd need the extra `Eff` when using `Process` for integrating widgets. `unsafeInterleaveEff` should definitely go.

PR #120, opened 2015-05-08
Hack to fix exponential event listeners

puffnfresh:


paf31:
:+1:

jdegoes:
@paf31 @puffnfresh As discussed, I don't think this is the right place to put this (best to put it into driver), but I do think it's the correct thing to do (given the current architecture of halogen). Merged for now since it's an improvement.

Issue #121, opened 2015-05-11
Input is processed twice because of postRender


puffnfresh:
After thinking about the problem, @jdegoes and I realised that driver is updating of the `Ref { process :: Process req eff, node :: HTMLElement }` in a way which is broken:

https://github.com/slamdata/purescript-halogen/blob/219bab3df0/src/Halogen.purs#L152-L157

This is because both `mainLoop` and `postRender` use the driver. The outer call running over the first element, inner running over the second, we then unwind and process the second input **again**.

This is why the `setTimeout` fixed things - `postRender` will then read the ref *after* it gets updated from `mainLoop` so the input it gets is processed only once.

So the `setTimeout` will work, but it's a workaround to the way the ref is being updated. We should get around to fixing the underlying problem sometime.

puffnfresh:
So @jdegoes came up with a little bit safer solution: always call `setTimeout` when the driver is executing, so that *all* input processing is happening un-nested.

PR #123, opened 2015-05-11
Make all calls to driver by async

puffnfresh:
This is a little bit safer than #120, and bit of a safer hack for the
fundamental state problem described in #121.

The driver can not be called recursively, since updating the mainLoop
ref will conflict, leading to inputs being processed many times.


paf31:
:+1: We should probably get rid of the call to `later` in the `Event` monad then, since this is handled there too.

PR #125, opened 2015-05-19
Drop IsForeign from onInput

garyb:
`IsForeign` here is a bit misleading, as the value is always a string for this event. I've been tripped up several times trying to use `IsForeign` to cast the value to a number for me, but it won't ever do that as `IsForeign` reading only succeeds for prim values with the right runtime tag.

The same might be true for `onValueChanged`, but I'm not 100% certain.

### DIFF
--- a/docs/Halogen-Forms.md
+++ b/docs/Halogen-Forms.md
@@ -26,7 +26,7 @@ Attach an event handler which will fire when a checkbox is checked or unchecked
 #### `onInput`
 
 ``` purescript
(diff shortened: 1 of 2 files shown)


jdegoes:
:+1: I hate foreign. :smiling_imp: 

> The same might be true for onValueChanged, but I'm not 100% certain.

How can a value ever be a non-string

garyb:
I was thinking in the case of `<select>` - I'll do a test to see you get values or indices, I suspect it is values though.

PR #129, opened 2015-05-29
Added data_ attribute

eskimor:
Disclaimer: I tested the code in my module, but I did not compile purescript-halogen with the changes. The change is pretty simple though.


jdegoes:
Is there such an attribute?

garyb:
There is a `data` attribute yes, we'd have to call it `data_` as `data` is a reserved word in PureScript. I'm not sure how this plays with the other `data-` stuff we have going on though.

puffnfresh:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/object

PR #132, opened 2015-05-30
Fixed "for" HTML attribute

hansler:
The "for" smart constructor now uses the proper attributeName.


jdegoes:
:+1: Thanks!

garyb:
I only just noticed this after bashing my head against a wall for half an hour on a `for` related problem. Thanks!

Issue #133, opened 2015-05-31
Attributes v. Properties, looking toward SVG support


ryanartecona:
First, some context: attributes are set in HTML, and properties are set on DOM objects. For some attributes and their corresponding properties, you can get/set with either interface, and the other will always have the same value, e.g. `id`, and `class`/`className`. For others, there's a meaningful difference, e.g. `value` on form inputs: there the `value` attribute is the value of the element when it was created, and the `value` property is the one that updates when the user interacts with it.

This is important for SVG support, since SVG nodes need to be created with `document.createElementNS('http://www.w3.org/2000/svg', tagName)`, and their attributes need to be set with `node.setAttributeNS(null, k, v)`.

The virtual-dom library does have this attribute/property distinction, and exposes it to the user. With this in place, it can support SVG with a [simple wrapper](https://github.com/Matt-Esch/virtual-dom/blob/master/virtual-hyperscript/svg.js) which sets the element's namespace, so it gets created appropriately with `document.createElementNS` instead of `document.createElement`, and which does [some autodetection](https://github.com/Matt-Esch/virtual-dom/blob/master/virtual-hyperscript/svg-attribute-namespace.js) on known attributes, so it can convert properties to attributes and make the appropriate `getAttributeNS`/`setAttributeNS` calls.

**1. exposing a attribute/property distinction**
Currently, halogen auto-detects `data-...` attributes, which get treated properly as attributes, and every other attribute gets treated as a property. That is, halogen doesn't distinguish between attributes and properties anywhere except in [the implementation of `Halogen.Internal.VirtualDOM.vnode`](https://github.com/slamdata/purescript-halogen/blob/fba54c53fe5d470094275b91fae80f0682becd15/src/Halogen/Internal/VirtualDOM.purs#L184-L210). As such, there's **(a)** no way for a user to create SVG elements, and **(b)** no way to add support for SVG without changing Halogen internals.

**2. Halogen refers to properties as [`Attr`](https://github.com/slamdata/purescript-halogen/blob/fba54c53fe5d470094275b91fae80f0682becd15/src/Halogen/HTML/Attributes.purs#L117)**
[...]

Issue #111, opened 2015-05-05
Change representation of Attributes


paf31:
`Attr` has a complex representation, and doesn't offer any real type safety. We would get just as much type safety if we used an `Any` type and hid the implementation behind smart constructors.

Alternatively, we could represent `Attr` as a large sum type, but that would make it more difficult to support new attributes.

jdegoes:
I'm not sure what I think about this, but I definitely agree that `Attr` is messy as it stands. What problem would this solve, or does it more fall in the order of cleaning things up?

paf31:
It's just about cleaning up the code.

garyb:
Closing as things have changed a little since this issue was raised (although we're still essentially using the same formulation for `Prop` which is the new `Attr`).

PR #134, opened 2015-06-03
Svg support (WIP)

ryanartecona:
:warning: _Not ready for merge_ :warning: 

This is a working POC, following the proposed plan in #133. 

Uncertainties:

  - In `Halogen.Internal.VirtualDOM`, things dealing with `prop` and `fooProp` are now `vprop` and `fooVProp`. This avoids conflict between `Halogen.HTML.Attributes.prop` and (previously) `Halogen.Internal.VirtualDOM.prop`.
    - Is this necessary? It seemed like the thing to do to me.
    - Should it go further, and prefix _everything_ in `Halogen.Internal.VirtualDOM` with 'v'? Things like `vtree`, `vnode`, and `vtext` are prefixed, but `diff`, `patch`, and `createElement` still aren't.
  - The new `Prop` constructor just copies `Exists AttrF` from `Attr`, which I assume is fine since it's the same shape, and may change soon anyway with #111.
  - The stuff so far in `Halogen.SVG` just use strings directly in attributes. I imagine this is ok for a first pass, maybe revisit when that module is more complete.

What's left to do:

  - [ ] change all (most?) of the Attrs in `Halogen.HTML.Attributes` from using `attr` to using `prop` now (since the old `attr` behaves like the new `prop`).
  - [ ] flesh out `Halogen.SVG` elements and attributes.
  - [ ] docs

### DIFF
--- a/docs/Halogen-HTML.md
+++ b/docs/Halogen-HTML.md
@@ -1804,11 +1804,22 @@ data HandlerF i fields
 The data which represents a typed event handler, hidden inside an existential package in
 the `Attr` type.
 
+#### `AttrNS`
+
+``` purescript
+data AttrNS
+  = SvgNS 
+  | XmlNS 
+  | XlinkNS 
+```
+
+
 #### `Attr`
 
 ``` purescript
 data Attr i
-  = Attr (Exists AttrF)
+  = Attr (Maybe AttrNS) (Exists AttrF)
+  | Prop (Exists AttrF)
   | Handler (ExistsR (HandlerF i))
   | Initializer i
   | Finalizer i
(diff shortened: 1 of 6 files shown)


eskimor:
svg support? awesome! :-)

jdegoes:
Thanks for working on this, @ryanartecona, let us know when you're ready for a code review.

ryanartecona:
@jdegoes If there's nothing objectionable in #133, or if it's easier/quicker to just look at code changes, this is ready for a first review. I wanted a sanity check before committing to the approach and doing the slightly more time consuming bit of adding a bunch of SVG elements and attributes.

dimsuz:
This PR was closed with no reason specified - I'm curious what was the fate of SVG (WIP) support?..

jdegoes:
See discussion here: slamdata/purescript-halogen#133

Issue #135, opened 2015-06-08
Missing requirement to build examples README.md


jmatsushita:
As discussed [here on IRC](http://ircbrowse.net/browse/purescript?id=106201&timestamp=1432599960#t1432599960)

It would be good to add this requirement to the purescript-halogen examples:
```npm install virtual-dom```

stkb:
If I may, there's a couple of other easy-fix points that I'd like to tack on to this issue, that I also ran across when trying the examples.

1. The `pulp build` step isn't necessary, since the browserify step does a compile first anyway.
2. In the **intro** example the `index.html` file looks for the .js file at `js/index.js`; needs to be changed to `dest/index.js`.
3. The **ajax** example doesn't work in IE, because it's requiring an outdated version of purescript-aff. This needs updating in the `bower.json`.

codygman:
I was affected by @jmatsushita's issue as well. Here's a full log of what happend following the current instructions:

```shell
~/source/unsorted/purescript-halogen/examples/counter $ pulp browserify
* Browserifying project in /home/cody/source/unsorted/purescript-halogen/examples/counter
* Building project in /home/cody/source/unsorted/purescript-halogen/examples/counter
* Build successful.
* Browserifying...
events.js:85
      throw er; // Unhandled 'error' event
            ^
Error: Cannot find module 'virtual-dom/create-element' from '/home/cody/source/unsorted/purescript-halogen/examples/counter/output/Halogen.Internal.VirtualDOM'
    at /home/cody/.nvm/versions/node/v0.12.5/lib/node_modules/pulp/node_modules/browserify/node_modules/browser-resolve/node_modules/resolve/lib/async.js:46:17
    at process (/home/cody/.nvm/versions/node/v0.12.5/lib/node_modules/pulp/node_modules/browserify/node_modules/browser-resolve/node_modules/resolve/lib/async.js:173:43)
    at ondir (/home/cody/.nvm/versions/node/v0.12.5/lib/node_modules/pulp/node_modules/browserify/node_modules/browser-resolve/node_modules/resolve/lib/async.js:188:17)
    at load (/home/cody/.nvm/versions/node/v0.12.5/lib/node_modules/pulp/node_modules/browserify/node_modules/browser-resolve/node_modules/resolve/lib/async.js:69:43)
[...]

PR #143, opened 2015-06-28
fix issues with examples (#135)

stkb:
(fixes issue #135)
Adding instructions to the readme to make sure virtual-dom is installed
Fixing the path of the index.js file in the into example
Updating purescript-affjax dependency in the ajax example, to make it work in IE


jdegoes:
:+1: Thanks!

PR #144, opened 2015-07-08
Update build and core library for purescript 0.7

garyb:
The examples haven't been updated yet as they depend on other projects that need updating too, and I guess will be needing other updates soon anyway.

Any preference on how we handle that @jdegoes?


jdegoes:
Let's leave the examples non-functioning just so we can get this in. They can be updated later, after their dependencies are updated, and of course again once halogen gets the component updates.

garyb:
Ok, this is good to go as it is now then.

PR #147, opened 2015-07-15
New components

### LINKED ISSUE #133: Attributes v. Properties, looking toward SVG support

### LINKED ISSUE #138: New Halogen Components
# Introduction

Halogen components currently suffer from a variety of drawbacks:

 * **Queries**. It is not possible to extract information from components. Since components close over their state, this means it is not possible to read component state. For foreign components, it is not possible to effectfully extract foreign state. 
 * **Localization**. Every detail about even the smallest component somewhere deep inside the application must propagate all the way to the top, even for pure components. Parent and child components cannot communicate in a principled or localized way.
 * **Invariance**. Components are invariant in their input type: they must accept the same type they produce in response to user input (at least in practice, if not in theory). This leads to a single type representing commands, requests for information, and responses. As a result, the types alone do not precisely model information flow, and the types permit many undesirable flows with poorly-defined semantics.

To work around the first two issues, the emerging anti-pattern involves passing around `Driver` so that components may, on user input, effectfully inject information into the top of the program, where it can then trickle down to where it is required. This effectful style is difficult to reason about and leads to tangled programs.

This proposal introduces a new design for components, based on many iterations and much discussion, which is not yet fully realized, but which has the ability to address all these drawbacks while still permitting classic Halogen components and combinators as a special case of a more generic abstraction.

## Component

```purescript
newtype Component s f g p = Component 
  { render  :: s -> HTML p (f Unit),
    query   :: forall i. Free f i -> StateT s g i }
    
instance functorComponent :: Functor (Component s f g)
```

A component is a bundle of two functions:

1. `render` — This is a pure function that renders the component to HTML based on the component's state. Unlike prior Halogen components, this rendering function may not perform any effects on user input.
2. `query` — This is a function to query the component, which may or may not modify the state or have other effects up to `g`.

The type parameters have the following meanings:

 * `s` - The type of the component state. A component is invariant in `s`.
 * `f` - The component's query algebra, whose interpretation has effects up to `StateT s g`.
 * `g` - The monad describing the component's extra (non-`State`) effects.
 * `p` - The type of placeholders in the component, which can be thought of as addresses for child components that will be installed into the HTML of the parent component. Components form a `Functor` in `p`, allowing easy manipulation of placeholders.

### Rendering

A component can render itself, given the state. As part of this rendering, it may specify the event listeners to be called on user input. These event listeners may produce commands to the query DSL accepted by the component. 

By design, commands generated from user input return `Unit`, because the `HTML` rendering must neither know nor care what the query processor does with them. Rather, the `HTML` rendering must always be strictly a function of component state.

User input may generate only one command per user input event. This requirement could be loosened but only by allowing the `HTML` rendering to interact with the query processor.

### Querying

A component also exposes a query DSL, which is a `Free` algebra in `f`. Classic Halogen components consist purely of "write-only" commands, which return `Unit`, and which are designed to be called in response to user-input.

The query DSL is implemented through a query processor, which is a function that lifts the query DSL into a monad `g`, with stateful effects on the component's state layered on using `StateT`. Thus, the query processor may implement the query DSL using both the component state (reads and writes), as well as additional effects captured by `g`.

### Composition

Child components can be installed into parent components. Parent components can communicate with their children, as well as all descendants. Sibling components can communicate only through the facilitation of a parent component.

This communication architecture ensures that parent components remain in control of information flow and promotes modular architecture.

### Asynchronous Behavior

By design, components cannot block rendering, because they are split into pure rendering and possibly effectful query processing parts.

In addition, there is no easy way for race conditions to accidentally overwrite state, because all query processing occurs in a `StateT` monad, and query processors do not have direct access to state (and therefore cannot easily capture over it). This does not prevent such conditions from occurring, but it does make it much harder for a user to accidentally squash state.

## Pure Component

Pure components are simply components whose effects are limited to component state.

```purescript
type ComponentPure s f p = forall g. (Monad g) => Component s f g p
```

## Classic Halogen Components

Classic Halogen components are simply a special case of normal components, where all inputs in the query DSL are commands that return `Unit`.

```purescript
module ClickComponent where
  import Control.Monad.State.Trans
  import Control.Monad.State.Class(modify)
  import Control.Monad.Free
  import Control.Monad.Rec.Class

  import Data.Inject
  import Data.Void
  import Data.Identity

  import Halogen  
  
  data Input a = ClickIncrement a | ClickDecrement a

  instance functorMyComponent :: Functor Input where
    (<$>) f (ClickIncrement a) = ClickIncrement (f a)
    (<$>) f (ClickDecrement a) = ClickDecrement (f a)

  clickIncrement :: forall g. (Functor g, Inject Input g) => Free g Unit
  clickIncrement = liftF (inj (ClickIncrement unit) :: g Unit)

  clickDecrement :: forall g. (Functor g, Inject Input g) => Free g Unit
  clickDecrement = liftF (inj (ClickDecrement unit) :: g Unit)

  counterComponent :: forall g. (MonadRec g) => Component Number Input g Void
  counterComponent = Component { render : render, query : query }
    where
      eval :: forall g a. (Monad g) => Input (Free Input a) -> StateT Number g (Free Input a)
      eval (ClickIncrement next) = do
        modify (+1)
        return next
      eval (ClickDecrement next) = do
        modify (flip (-) 1)
        return next
      
      render :: Number -> HTML Void (Input Unit)
      render n = todo
      
      query :: forall g i. (MonadRec g) => Free Input i -> StateT Number g i
      query = runFreeM eval

  test :: Free Input Unit
  test = do
    clickIncrement
    clickIncrement
    clickDecrement
```

By design, the commands that can be sent to the component (`clickIncrement`, and `clickDecrement`) work in any `Free` algebra which contains `Input` (including, of course, `Input` itself).

## Foreign Components

```purescript
module EditorComponent where
  import Control.Monad.State.Trans
  import Control.Monad.State.Class(modify)
  import Control.Monad.Free
  import Control.Monad.Rec.Class
  import Control.Monad.Eff
  import Control.Monad.Trans(lift)
  import Control.Apply((*>))

  import Data.Inject
  import Data.Void
  import Data.Identity

  import Halogen  

  data Input a = GetContent (String -> a) | SetContent a String | GetCursor (Number -> a)

  instance functorInput :: Functor Input where
    (<$>) = todo
  
  getContent :: forall g. (Functor g, Inject Input g) => Free g String
  getContent = liftF (inj (GetContent id) :: g String)
  
  setContent :: forall g. (Functor g, Inject Input g) => String -> Free g Unit
  setContent s = liftF (inj (SetContent unit s) :: g Unit)

  getCursor :: forall g. (Functor g, Inject Input g) => Free g Number
  getCursor = liftF (inj (GetCursor id) :: g Number)
  
  editorComponent :: forall eff. Component Unit Input (Eff (dom :: DOM | eff)) Void
  editorComponent = Component { render : render, query : query }
    where 
      eval :: forall eff a. Input (Free Input a) -> StateT Unit (Eff (dom :: DOM | eff)) (Free Input a)
      eval (GetContent f  ) = lift $       f <$> effectfulGetContent
      eval (SetContent n s) = lift $ const n <$> effectfulSetContent s
      eval (GetCursor  f  ) = lift $       f <$> effectfulGetCursor
      
      render s = todo
      
      query :: forall eff i. Free Input i -> StateT Unit (Eff (dom :: DOM | eff)) i
      query = runFreeM eval

  test :: Free Input Unit
  test = do
    cursor  <- getCursor
    content <- getContent
    setContent $ content ++ (show cursor)

  foreign import data DOM :: !
  
  foreign import effectfulGetContent :: forall eff. Eff (dom :: DOM | eff) String
  foreign import effectfulSetContent :: forall eff. String -> Eff (dom :: DOM | eff) Unit
  foreign import effectfulGetCursor  :: forall eff. Eff (dom :: DOM | eff) Number
```

## Composing Components

The `install` family of functions can be used to install a child component into a parent component at well-defined places inside the HTML of the parent component.

```purescript
  installR :: forall s f g pl pr s' f' p'. (Ord pr, Plus g) =>
    Component s f (QueryT s' f' pr p' g s) (Either pl pr) ->   -- parent
    (pr -> Tuple s' (Component s' f' g p'))               ->   -- factory
    Component (InstalledState s s' f' g pl p') (Coproduct f (ChildF pr f')) g (Either pl p')
  installR a f = todo

  installL :: forall s f g pl pr s' f' p'. (Ord pl, Plus g) =>
    Component s f (QueryT s' f' pl p' g s) (Either pl pr) ->   -- parent
    (pl -> Tuple s' (Component s' f' g p'))               ->   -- factory
    Component (InstalledState s s' f' g pr p') (Coproduct f (ChildF pl f')) g (Either pr p')
  installL a f = todo

  installAll :: forall s f g p s' f' p'. (Ord pr, Plus g) =>
    Component s f (QueryT s' f' p p' g s) p               ->   -- parent
    (p -> Tuple s' (Component s' f' g p'))                ->   -- factory
    Component (InstalledState s s' f' g p p') (Coproduct f (ChildF p f')) g p'
  installAll a f = todo

  type ComponentState s f g p = Tuple s (Component s f g p)

  type ComponentState s f g p = Tuple s (Component s f g p)

  type InstalledState s s' f' g p p' = 
    { parent   :: s, 
      children :: Map.Map p (ComponentState s' f' g p'), 
      factory  :: p -> ComponentState s' f' g p' }

  data QueryT s' f' p p' g s a = QueryT (StateT (InstalledState s s' f' g p p') g a)
```

The installation happens automatically in response to changes in the parent component's HTML. Thus, the number and locations of the child components may change dynamically in response to user input or queries to the parent component.

The output monad of every parent component is `QueryT` for some base monad `g`. This monad transformer allows the parent to manipulate its own state, and send and received typed messages to its children. After installation using `installR`, `installL` or `installAll`, the communication between parent and child, as well as the children placeholders, are completely hidden, leaving only a simple component with composite state, composite query DSL, and "leftover" placeholders.

```purescript
  -- queries a particular child from the parent:
  query :: forall s s' f' p p' g. p -> (forall i. Free f' i -> QueryT s' f' p p' g s (Maybe i))
  query p q = todo

  -- lifts an effect into the QueryT monad:
  effect :: forall s' f' p p' g s a. (Monad g) => g a -> QueryT s' f' p p' g s a
  effect ga = todo

  -- MonadState for QueryT so parents can manipulate their own state
  instance monadStateQueryT :: MonadState s (QueryT s' f' p p' g s) where
    state f = todo

  instance functorQueryT :: Functor (QueryT s' f' p p' g s) where
    (<$>) f fa = todo

  instance applyQueryT :: Apply (QueryT s' f' p p' g s) where
    (<*>) f fa = todo

  instance applicativeQueryT :: Applicative (QueryT s' f' p p' g s) where
    pure a = todo

  instance bindQueryT :: Bind (QueryT s' f' p p' g s) where
    (>>=) fa f = todo

  instance monadQueryT :: Monad (QueryT s' f' p p' g s)

  instance functorChildF :: (Functor f) => Functor (ChildF p f) where
    (<$>) f (ChildF p fi) = ChildF p (f <$> fi)
```

Two functions `query` and `effect`, as well as standard instances for `QueryT`, may be used to easily build parent component query processors:

```purescript
  test = do 
    state <- get
    effect $ doEffect
    set state
    resp <- query editor getContent
    return $ fromMaybe "" resp
```

## Running Components

To run a component, you provide the component and an initial state.

The component has an arbitrary input DSL `f`, and runs effects in `Aff`. By design, components cannot ever block rendering.

```purescript
  type Driver f eff = forall i. f i -> Aff (HalogenEffects eff) i

  runComponent :: forall eff s f. (Functor f) =>
    Component s f (Aff (HalogenEffects eff)) Void ->
    s                                             ->
    Aff (HalogenEffects eff) (Tuple HTMLElement (Driver f eff))
  runComponent c s = todo
```

garyb:
:boom:

Resolves #138, resolves #133.

I can squash the commits before merging, and probably should add some example/tests for installed components/child queries etc. but the general API is good to review. Perhaps needs some more doc comments too.

There are a few other changes in here that I probably should have done separately afterwards:

- `Attr` representation has been overhauled to distinguish between `Attr`s and `Prop`s. Also `Key` is a special value now also. The type is called `Prop`, as all the constructors do all represent some kind of property, `Meta` seemed like it might be a bit confusing.
- Not every property has a corresponding attribute, and for those that do the property name and attribute name are not necessarily the same, so the class for rendering properties in the string renderer was adjusted slightly to accommodate this.
- `Element`s can now be namespaced
- Removed all the empty element constructors that previously accepted children (for example, `input_` has been renamed to `input` and there is no equivalent to the old `input` now - `<input>` isn't allowed to have content inside it).
- The form event helpers were renamed to be more consistent with the events (`onValueChanged` -> `onValueChange` corresponding to `onChange`, `onInput` -> `onValueInput` and the "normal" form of `onInput` was added to `Halogen.HTML.Events` as it previously was missing).
- The form event helpers only require an `Applicative f` rather than `Alternative f` now, as we can return `unit` instead of using `empty` (I think this behaviour is ok? as generally it will be `Pure unit` as `f` will be some `Free`).

I removed much of the machinery that we no longer require to make things work, but that include signals, do we want to keep those still? Maybe they should be moved into their own project. Phil also has a version of signal functions implemented on top of `purescript-signal`, so maybe that's enough.

I think that covers it.


jdegoes:
:fireworks: :beers: Will review everything in detail tomorrow! Very excited for these changes to land.

puffnfresh:
Would love to read a summary of how the changes contribute to component querying and localisation.

src/Halogen.purs
```diff
@@ -1,159 +1,78 @@
...
+
+type DriverState s =
+  Maybe { node :: HTMLElement
+        , vtree :: VTree
+        , state :: s
+        }
+
+type Driver f eff = forall i. f i -> Eff (HalogenEffects eff) i
+
+runUI :: forall eff s f. Component s f (Eff (HalogenEffects eff)) Void
```
jdegoes: This does not allow asynchronous effects. The type should probably be `Aff`, both here and in `Driver`.

jdegoes: Or actually, one could generalize to `MonadAff`, which is the asynchronous equivalent of `MonadEff`.

src/Halogen.purs
```diff
@@ -1,159 +1,78 @@
...
-        Just { process: process, node: node } -> do
-          let work = runSF process (Tuple req node)
-          node' <- head work
-          writeRef ref $ Just { process: tail work, node: node' }
-        Nothing -> log "Error: An attempt to re-render was made during the initial render."
+  driver :: Ref (DriverState s) -> Driver f eff
+  driver ref q = do
+    refVal <- readRef ref
+    case refVal of
+      Nothing -> throwException $ error "Error: An attempt to re-render was made during the initial render."
```
jdegoes: With `Aff`, these would likely change to `AVar`, which would have the nice side-effect of eliminating this bit of unpleasantness.

src/Halogen.purs
```diff
@@ -1,159 +1,78 @@
...
+      Just { node: node, vtree: prev, state: s } -> do
+        Tuple i s' <- queryComponent c q s
+        case renderComponent c s' of
+          Tuple html s'' -> do
+            let next = renderHTML (driver ref) html
+            node' <- patch (diff prev next) node
+            writeRef ref $ Just { node: node', vtree: next, state: s'' }
+            pure i
+
+actionF :: forall f g. (Functor f, Functor g, Inject f g) => (forall i. i -> f i) -> Free g Unit
```
jdegoes: This function and the others could use some documentation. It's not immediately clear what purpose they serve.

src/Halogen.purs
garyb: Ah right, yes, I wanted to talk about this (`Aff` vs `Eff`, I'll reply to MonadAff separately)

One slightly unfortunate thing about using `Aff` over `Eff` is since Halogen then has to run the `Aff`s it means that if there's a failure and `attempt` wasn't used the error kinda disappears (in the current Halogen we just log it to the console).

If you use `Eff` instead, it means one has to `runAff` themselves, and there's no way that error can disappear, unless `launchAff` was used, which is the expected behaviour anyway.

I'm totally fine with using `Aff`, just wanted to discuss this tradeoff and forgot to mention it in the PR.

src/Halogen/Component.purs
```diff
@@ -1,64 +1,214 @@
...
+queryComponent :: forall s f g p i. Component s f g p -> f i -> s -> g (Tuple i s)
+queryComponent (Component c) q = runStateT (c.query q)
+
+instance functorComponent :: Functor (Component s f g) where
+  map f (Component c) =
+    Component { render: lmap f <$> c.render
+              , query: c.query
+              }
+
+component :: forall s f g p. (s -> HTML p (f Unit)) -> (forall i. f i -> StateT s g i) -> Component s f g p
```
jdegoes: This one is not using the type aliases like the other ones, which makes it inconsistent and its signature overly complex.

src/Halogen.purs
garyb: Regarding `MonadAff`/`MonadEff`, I was really hoping that approach would work, as it makes for much nicer types, but the problem is we'd have to provide an `m a -> Eff _ a` function anyway to make it work, as we need a way of running the effect when calling the driver from the event handlers: https://github.com/slamdata/purescript-halogen/blob/c7ecfc88efe71da246bf45b79a69bfe66ae11fa2/src/Halogen/HTML/Renderer/VirtualDOM.purs#L40-L41

src/Halogen/Component.purs
garyb: Oops thanks, I only came up with those synonyms today when working on the examples, so forgot about this one.

src/Halogen.purs
garyb: Ah! I think there's nothing to discuss after all, I'm pretty sure it has to be `Aff` rather than `Eff`, as we don't have access to the driver in `eval` for a component.

src/Halogen/HTML/Core.purs
```diff
@@ -0,0 +1,209 @@
...
+-- | - A initializer that is triggered once the element for the property has
+-- |   been added to the DOM.
+-- | - A finalizer that is triggered when the element for the property is
+-- |   removed from the DOM.
+data Prop i
+  = Prop (Exists PropF)
+  | Attr (Maybe Namespace) AttrName String
+  | Key String
+  | Handler (ExistsR (HandlerF i))
+  | Initializer (HTMLElement -> i)
```
jdegoes: I like the new initializers and finalizers, should eliminate the need for some post render hooks.

src/Halogen/HTML/Events.purs
```diff
@@ -1,170 +1,129 @@
...
+
+import Data.Coyoneda (Coyoneda(), liftCoyoneda)
+import Data.Inject (Inject)
+
+import Halogen (actionF, actionFC)
+import Halogen.HTML.Events.Handler (EventHandler())
+import Halogen.HTML.Events.Types (Event(), MouseEvent(), FocusEvent(), KeyboardEvent())
+import Halogen.HTML.Core (Prop(), handler, eventName)
+
+inputF :: forall f g a. (Functor f, Functor g, Inject f g) => (forall i. a -> i -> f i) -> a -> EventHandler (Free g Unit)
```
jdegoes: Are these all useful functions? Would be worthwhile explaining what they're used for.

src/Halogen/HTML/Events/Forms.purs
```diff
@@ -1,48 +1,34 @@
...
 
-import Control.Plus
-import Control.Alternative
+import Halogen.HTML.Core (Prop(), eventName, handler)
+import Halogen.HTML.Events.Handler (EventHandler())
 
-import Halogen.HTML.Events.Handler
+-- | Attach event handler to event `key` with getting `prop` field as an
+-- | argument of `handler`.
+addForeignPropHandler :: forall f value. (Applicative f, IsForeign value) => String -> String -> (value -> EventHandler (f Unit)) -> Prop (f Unit)
```
jdegoes: This is so terrible, hopefully we can eliminate this someday.

src/Halogen/HTML/Events.purs
garyb: Yeah, these along with the `actionF`/`requestF` and `FC` variants all need some explanatory comments. I'm pretty sure they are useful though, as they essentially let you use DSL constructors directly: https://github.com/slamdata/purescript-halogen/blob/5b817583266b03e32622f749dcd6a1173e0d7ed6/test/Test/Main.purs#L43

src/Halogen/HTML/Events/Forms.purs
garyb: At least it's private :smile:

test/Examples/Intro.purs
```diff
@@ -0,0 +1,43 @@
...
+  where
+
+  render :: RenderFC State p Input
+  render (State s) =
+    H.div_ [ H.h1_ [ H.text "Toggle Button" ]
+           , H.button [ E.onClick (E.inputFC_ ToggleState) ]
+                      [ H.text (if s.on then "On" else "Off") ]
+           ]
+
+  eval :: Eval Input State g
```
jdegoes: Thank god for type aliases. :smile:

src/Halogen.purs
garyb: I'm working on this now, but it's a bit tricky as we need a `MonadRec` instance for `Aff` (because of `runFreeM`/`runFreeCM`).

jdegoes: `Aff` should have a `MonadRec`, it should probably let the stack grow for a bit before jumping off using `later`.

jdegoes: Let me know if you need me to add `MonadRec` to `Aff` or if you have that (it is something I have been meaning to do, since a valid implementation is possible; a performant one is trickier since you want to stay on the stack for some time before bouncing off using `later`).

garyb: I'm just about to test one now, I let it run 50 times before bouncing off. It's hard to know what a sensible value is!

jdegoes: 50 times should be fine (don't want to get too aggressive). I've done call stacks nested hundreds of layers deep in JS.

garyb: Yeah, that's what I figured - 50 seemed conservative yet a reasonable maximum that we'll probably never reach in the Halogen case, or certainly not often enough for it to be a problem.

jdegoes: Consider that Promises/A+ bounce on every bind. :smile:

garyb: Wow, I did not know that! :scream:

We were also talking about using [asap.js](https://github.com/kriskowal/asap) in #purescript, but that would make using aff a bit more complicated with the extra dependency.

natefaubion: Most fast promises library use something similar to asap, which is essentially bouncing like what monad rec does. The promises spec only requires that it happens after the current call stack, not that it happens on a subsequent tick. I've modified some hot algorithms in sweet.js to use Bluebird as an expirement, and only saw about a 10-15% decrease in performance. The raw part of asap.js is small enough to just inline as well. It's certainly not any longer than other FFI code in Aff :)

jdegoes:
> The form event helpers only require an Applicative f rather than Alternative f now, as we can return unit instead of using empty (I think this behaviour is ok? as generally it will be Pure unit as f will be some Free).

Yes, that's fine, in fact, the only type safe way to implement `query` is with `Free`, since we don't have GADTs. You can hide data constructors and (pseudo-safely) take advantage of that knowledge by write `query` without `Free`, but there's going to be coercion involved.

I am glad to see `Event` etc. go, this all seems a lot simpler.

> include signals, do we want to keep those still? Maybe they should be moved into their own project.

Yeah, that should be its own project, it doesn't really depend on Halogen. Signal functions (of HTML) can be compiled into views, so the concept is still useful.

I'll take another look through, but I think most of my (unmade) suggestions can and should be done in separate PRs (for which I'll create separate tickets). For this PR, I'd just like to see a demo of Ace that utilizes nested components. Not only is this one of the most requested examples, but it will prove out the concepts.

Is there any other cleanup-type tasks you passed on for now, or did you get most of your wish list into this PR?

garyb:
Great, once the `MonadRec Aff` stuff is in I can finish switching to `Aff` and finish up the ajax example I'm half way through converting, once that is done, I'll do the Ace one.

And nope, I think this is it. Pretty sure I covered all the major things I wanted to clean up in here, aside from maybe the px/% for `width` and `height` values.

jdegoes:
@puffnfresh This is basically #138, which explains how the design achieves querying, localization and modularity. However, I think when @garyb does a real Ace example that utilizes local querying and component nesting, it will be much more tangible. If there are any tricky parts you'd like to see in that example, let us know!

garyb:
I've switched over to `Aff` from `Eff` now, the `AVars` certainly cleaned up that stuff in `runUI`/`driver`!

What do you think about this? https://github.com/slamdata/purescript-halogen/commit/831681c8624b071d164572b39bba19c3239eeb49#diff-5e398ef5650dd9aaccecdf2f73089d81R49

In the current halogen we just log when there are uncaught errors from an `Aff` that is being run, but I think re-throwing might be more appropriate.

I've also updated the ajax example, and it's revealed an interesting maybe-problem with the new model. When a request is made to the server we actually want to trigger multiple inputs - one to set the UI state to "busy", and then another one that updates it after the `Aff` completes. My first attempt doesn't work, as [I change the busy state during the handler that runs the `Aff`](https://github.com/slamdata/purescript-halogen/commit/831681c8624b071d164572b39bba19c3239eeb49#diff-07dd2ef337cf3a169ba61a5a0f46996eR100), but that doesn't trigger a re-render. I think I know the workaround (combine two inputs), but this wasn't something I entirely expected.

The build is failing right now as I need something to be merged into `purescript-nonempty` and to restructure things a bit so that the examples have their own dependencies.

garyb:
Hm, no, "combine two inputs" doesn't help out here either actually, as they still get run "together".

jdegoes:
> When a request is made to the server we actually want to trigger multiple inputs - one to set the UI state to "busy", and then another one that updates it after the Aff completes.

Yeah, this is something I have been thinking about (basically the reason we had `Event` in the last model, to emit multiple events). While you can emit multiple commands via `Free`, they will be treated as atomic and interpreted "at the same time" (up to `g`), so it doesn't really suit the above use case. Let me think on this one some more and maybe we can chat later today about solutions.

jdegoes:
> In the current halogen we just log when there are uncaught errors from an Aff that is being run, but I think re-throwing might be more appropriate.

That's probably the right thing to do in Javascript, as then the window main uncaught error handler can do whatever it does. Perhaps we should allow users to provide an "uncaught error handler" when they run the UI, we could utilize that to log errors back to SlamData or provide a bug report interface or something like that.

zudov:
I see the mixins were removed in https://github.com/slamdata/purescript-halogen/commit/cd7eab837e66aa373653ab1a0cdbd80bf2dbe8cd . What is the plan about those?

garyb:
I think something similar should still be possible, perhaps using functor coproducts to do the mixing-in, but I'm not exactly sure yet.

jdegoes:
> I think something similar should still be possible, perhaps using functor coproducts to do the mixing-in, but I'm not exactly sure yet.

Yes, you should be able to write combinators that take a component and return another one with a richer query language allowing more commands. 

Although, the old router mixin was not that useful (see purescript-router), and the undo / redo can be done a bit differently, you can actually snapshot & restore the whole app without needing a mixin, since the state is exposed. 

But I'm sure there are plenty of other cases where mixins will be the best solution to some problem.

src/Halogen/HTML/Events/Forms.purs
paf31: Curious: what don't you like about it?

jdegoes: Not this function, specifically (since it's private), just the fact that `IsForeign` and friends show up in the public API. :cry:

paf31: Yeah it's unfortunate. Problem is, you can guess at the fields the event might have, but browser differences etc. mean I wouldn't want to trust what I get back as the event, and want to validate it. This sort of thing probably should be hidden in some low level DOM events library somewhere.

jdegoes: That's what I'm thinking, it would be nice if there were a `purescript-events` library that has an on-opinionated, low-level abstraction across all browsers, which unifies the common stuff and in some type safe fashion supports the useful browser-specific stuff. Halogen and other libs could depend on it.

paf31: I was working under the assumption that `target.value` gave values of different types based on the input type: checkbox, radio, text, etc., but that's not true. According to the spec, it's always a string: http://www.w3schools.com/jsref/prop_attr_value.asp

So if the implementation is sound and the `value` property isn't set artificially, by a plugin for example, there's no need for a polymorphic `value`. Validating the value is probably still a good idea though, in case of bad plugins or whatever.

PR #152, opened 2015-08-13
New components

### LINKED ISSUE #133: Attributes v. Properties, looking toward SVG support

### LINKED ISSUE #138: New Halogen Components

garyb:
Resolves #138, resolves #133.

Things to do before a release: update/add remaining documentation. Adopt the "phantom rows" type safe  HTML DSL.


jdegoes:
:fireworks: :fireworks: :fireworks:  Can you squash the commits?

garyb:
Oops, certainly... the commit log got for this so long I forgot about it at a certain point. Done!

paf31:
:fireworks:

PR #154, opened 2015-08-13
Extract CSS module, remove unused dependencies

garyb:


bower.json
```diff
@@ -19,19 +19,16 @@
...
-    "purescript-maybe": "^0.3.3",
     "purescript-nullable": "^0.2.0",
-    "purescript-prelude": "^0.1.1",
-    "purescript-profunctor": "^0.3.0",
     "purescript-simple-dom": "^0.1.0",
     "purescript-unsafe-coerce": "^0.1.0",
     "purescript-void": "^0.3.0"
+  },
+  "resolutions": {
+    "purescript-strings": "^0.7.0"
```
garyb: This resolution is because of a dependency in `simple-dom`, but I'm hoping to replace that with `purescript-dom` before we release this anyway - I've been doing some work on it.

`simple-dom` will be broken by psc 0.7.3 as it has orphan instances, so that's another reason to switch over asap.

PR #160, opened 2015-08-23
Fix due to purescript-dom effect type change

nwolverson:
Effects in type of event listener stuff used in `onload` changed in [0.2.3](https://github.com/purescript-contrib/purescript-dom/releases/tag/v0.2.3)


garyb:
Thanks!

PR #162, opened 2015-08-28
Import re-exported HTML modules

passy:
Required for PSC >= 0.7.4.0

[Review on Reviewable](https://reviewable.io/reviews/slamdata/purescript-halogen/162)

PR #164, opened 2015-08-31
Update dependencies, fix warnings, better re-exports, clean up some other stuff

garyb:
Resolves #162

Also:
- Removed `Inject` mechanisms (they were basically unnecessary, can build `HalogenF` values manually with `left` and `right`, a user of the library can always handle their own `inj`ecting if they want composite query algebras)
- Removed a bunch of explicit typing that was required while relying on `Inject`
- Simplified some examples to be more focused on what they're demos of
- Now re-exports are properly usable `import Halogen` gets you all the useful stuff aside from things like element and property smart constructors, event types, etc.


garyb:
/cc @puffnfresh 

Hopefully my changes here won't interfere with your materials too much :smile:

PR #165, opened 2015-09-01
Update README

garyb:
The new README isn't complete yet, but perhaps better to have this attached to `master` rather than a completely inaccurate version?


puffnfresh:
Can we release this!? :+1:

PR #166, opened 2015-09-02
WIP multi-install components

garyb:
OK, so this approach is the one I've had most success with so far when it comes to having a sum of different types of component that want combining into a single parent. See the `examples/multi-component` stuff for a reasonably comprehensive example of how it might work (although I didn't include `peek` usage).

Of all the things I tried, "manually" `transform`ing the components to introduce `Either` and `Coproduct` into their types was by far the most successful, by which I mean something like this:

``` purescript
-- child components are defined as a normal standalone components, no either or coproduct stuff required:
componentA :: Component StateA InputA g p
componentB :: Component StateB InputB g p
componentC :: Component StateC InputC g p

-- we then use transformR to give us something that can be installed into the parent container:
transformR componentC :: Component (Either sl StateC) (Coproduct fl InputC) g p

-- or chain transforms together if we have a nested either:
transformL (transformR componentB) :: Component (Either (Either sl StateB) sr) (Coproduct (Coproduct fl InputB) fr) g p
```

I ended up making a `Data.Functor.Coproduct.Nested` module that is basically a clone of the equivalent `Data.Either.Nested` lifted up to the level of `Coproduct` to help make some of the types and resulting transforms a bit more manageable, perhaps introducing a collection of `component1of3` style functions would also be useful to avoid having to explicitly chain `transformL` and `transformR` together.

By "manually" up there I mean this happens within the `install` combinator, and each case is dealt with explicitly by doing the `transform`, rather than using something like `installL`. However, this means you do need to install all of the components together in one go, and I'm not sure if that's too restrictive. I can definitely think of cases where it would be useful to leave placeholders open in a parent while installing some components, and then later filling in the remaining placeholders.

One potential workaround would be to create a component which only consists of a placeholder, that way it can be installed in the slot of an existing placeholder to keep the install function total, and then later another `install` can fill in the re-substituted placeholder, but perhaps it would be better to encode this directly somehow instead? Maybe we could also have `install1of3` style functions to aid this, although I'm not sure if we'll run into problems making the types work there again.

The `transformL` and `transformR` functions are implemented in terms of `transform` which I've also exposed in case more flexibility than the L/R varieties provide is required. `transform` takes a bunch of functions which could be expressed in a lens-like form, but I chose not to do this as even though they look very much like `Iso` (and the combinators work beautifully), they're not really isomorphisms, as the paired functions usually involve an unsafe partial function (like `Left` and `fromLeft`), so I just left it as accepting a bunch of separate functions.

I was going to write much more about the things I've tried on the way to get here, but since I have something that actually works now it's probably not worth outlining the various dead ends I encountered. Instead I'd like to see what you all think of this approach so far, whether it's "too much", overly restrictive, any other concerns, and also if you have any ideas for improvements - to the design, combinators that would be helpful, anything!

/cc @jdegoes @cryogenian @jonsterling


cryogenian:
Is it possible to override child `eval`?

jonsterling:
This looks reasonable, though it's hard for me to predict whether or not it will be too restrictive. My preference in this case would be to risk over-restriction if it lets us keep the interface from getting more complicated...

The other thing I am not totally in love with is the amount of boilerplate that's required to get everything wired up properly, but it does at least look like the boilerplate is "canonical" (i.e. you can just follow the types and it will work). So I think it is probably fine...

garyb:
> Is it possible to override child `eval`?

No, although you can still use the `peek` mechanism to observe a child component's queries - this doesn't change any of the capabilities of parent/child components, it's just providing a way of installing children with different state/query algebra types.

cryogenian:
@garyb can I access parent's state during install?

garyb:
@jonsterling yeah, I'm with you there - there is an unfortunate amount of boilerplate, but I think some more `thingNofX` functions could help a little with that:

``` purescript
-- we could replace the either/coproduct1of3 here:
query (either1of3 PlaceholderA) (coproduct1of3 $ request GetStateA)

-- with something like:
query1of3 PlaceholderA (request GetStateA)
```

I've made some progress with an `installNofX` too, which means we might well be able to install selectively.

Something that would potentially cut down on the need for these combinators is making use of `Inject` again, with the downside that to do that we'd need to prefer the right rather than the left when combining `Coproduct`s.

garyb:
@cryogenian install doesn't have direct access to state, but you can achieve the same thing by passing it via the placeholder for a child component.

jdegoes:
> I ended up making a Data.Functor.Coproduct.Nested module that is basically a clone of the equivalent Data.Either.Nested lifted up to the level of Coproduct to help make some of the types and resulting transforms a bit more manageable, perhaps introducing a collection of component1of3 style functions 

I'm not sure this is a good approach. I think turning `Inject` into a `Prism`-like data type might be useful:

```purescript
data Inject f g = Inject { inj :: forall a. f a -> g a, prj :: forall a. g a -> Maybe (f a) }
```

Then defining a range of combinators on `Inject`. The composition will be explicit, but it's far saner to use such a mechanism than `query5of14` or whatever. :smile: 

> However, this means you do need to install all of the components together in one go, and I'm not sure if that's too restrictive. 

I'm not sure either. In many examples I can think of, this is the case. But still not sure.

> The other thing I am not totally in love with is the amount of boilerplate that's required to get everything wired up properly, but it does at least look like the boilerplate is "canonical" 

Yeah. I'm fine with "boilerplate that can't go wrong", i.e. brainless type massaging to prove something to the compiler. The boilerplate I hate is the kind that can go wrong...

I'll take a look at the PR in depth probably over the weekend. Unfortunately not much time for this right now.

garyb:
Ok, separating the `install` stuff out into separate combinators is problematic I think, I managed to implement them, but attempting to use them result in an infinite type... maybe this is rescuable by some way, but since we probably don't want to do `installNofX` anyway. Thinking more about it, I don't think the "install all at once" restriction is a fatal flaw, as we could use a component constuctor function to provide the unspecified child component stuff anyway.

I'll take a look at the `Inject` data constructor approach. I had also been considering that before - ideally they should compose so you can create a "path" out of them, also I think we'll need something more specific than `Inject` as we want to be able to build up and tear down both `Either` and `Coproduct` values at the same time.

garyb:
Following the `Inject` method, I came up with some combinators and things that lets us write the example like this:

``` purescript
module Example.MultiComponent where

import Prelude

import Control.Apply ((*>))
import Control.Monad.Aff (Aff(), launchAff)
import Control.Monad.Eff (Eff())
import Control.Monad.Eff.Exception (throwException)
import Control.Plus (Plus)

import Data.Const (Const())
import Data.Either (Either(..))
import Data.Either.Nested (Either3(), either1of3, either2of3, either3of3, either3)
import Data.Functor (($>))
import Data.Functor.Coproduct.Nested (Coproduct3(), coproduct1of3, coproduct2of3, coproduct3of3)
import Data.Maybe (Maybe(..))
import Data.Tuple (Tuple(..))
import Data.Void (Void())

import Halogen
import Halogen.Component.Inject (inl, inr, (:>))
import Halogen.Util (appendToBody)
import qualified Halogen.HTML as H
import qualified Halogen.HTML.Events as E

import Example.ComponentA
import Example.ComponentB
import Example.ComponentC

newtype State = State { a :: Maybe Boolean, b :: Maybe Boolean, c :: Maybe Boolean }

initialState :: State
initialState = State { a: Nothing, b: Nothing, c: Nothing }

data Input a = ReadStates a

type ChildState = Either3 StateA StateB StateC
type ChildInput = Coproduct3 InputA InputB InputC
type ChildPlaceholder = Either3 PlaceholderA PlaceholderB PlaceholderC

injA = inl :> inl
injB = inl :> inr
injC = inr

parent :: forall g p. (Functor g) => ParentComponent State ChildState Input ChildInput g (Const Void) ChildPlaceholder p
parent = component render eval
  where

  render :: Render State Input ChildPlaceholder
  render (State state) = H.div_
    [ H.div_ [ H.placeholder' injA PlaceholderA ]
    , H.div_ [ H.placeholder' injB PlaceholderB ]
    , H.div_ [ H.placeholder' injC PlaceholderC ]
    , H.div_ [ H.text $ "Current states: " ++ show state.a ++ " / " ++ show state.b ++ " / " ++ show state.c ]
    , H.button [ E.onClick (E.input_ ReadStates) ] [ H.text "Read states" ]
    ]

  eval :: Eval Input State Input (QueryF State ChildState ChildInput g ChildPlaceholder p)
  eval (ReadStates next) = do
    a <- liftQuery $ query' injA PlaceholderA (request GetStateA)
    b <- liftQuery $ query' injB PlaceholderB (request GetStateB)
    c <- liftQuery $ query' injC PlaceholderC (request GetStateC)
    modify (const $ State { a: a, b: b, c: c })
    pure next

ui :: forall g p. (Monad g, Plus g) => InstalledComponent State ChildState Input ChildInput g (Const Void) ChildPlaceholder p
ui = install parent (either3 installA installB installC)
  where
  installA PlaceholderA = createChild injA componentA initStateA
  installB PlaceholderB = createChild injB componentB initStateB
  installC PlaceholderC = createChild injC componentC initStateC

main = launchAff $ do
  app <- runUI ui (installedState initialState)
  appendToBody app.node
```

Definitely an improvement I think?

jdegoes:
Definitely an improvement. Though we can use a `query`-ish combinator to turn this:

```purescript
liftQuery $ query' injA p i
```

into something more like this:

```purescript
query injA p i
```

which in turn could be made a bit prettier using a local `where`

```
res <- queryAce p i
where queryAce = query aceInject
      ...
```

jonsterling:
@garyb pretty slick

PR #169, opened 2015-09-23
Multi-install components

garyb:
/cc @jonsterling @cryogenian @beckyconning

### DIFF
--- a/docs/Data/Injector.md
+++ b/docs/Data/Injector.md
@@ -1,0 +1,52 @@
+## Module Data.Injector
+
+#### `Injector`
+
+``` purescript
+data Injector a b
+  = Injector (a -> b) (b -> Maybe a)
+```
+
+##### Instances
+``` purescript
+instance semigroupoidInjector :: Semigroupoid Injector
+instance categoryInjector :: Category Injector
+```
+
+#### `inj`
+
+``` purescript
+inj :: forall a b. Injector a b -> a -> b
+```
+
+#### `prj`
+
+``` purescript
+prj :: forall a b. Injector a b -> b -> Maybe a
+```
+
+#### `injLE`
+
+``` purescript
+injLE :: forall a b. Injector a (Either a b)
+```
+
+#### `injRE`
+
+``` purescript
+injRE :: forall a b. Injector a (Either b a)
+```
+
+#### `injLC`
+
+``` purescript
+injLC :: forall f g a. Injector (f a) (Coproduct f g a)
+```
+
+#### `injRC`
+
+``` purescript
+injRC :: forall f g a. Injector (f a) (Coproduct g f a)
+```
+
+
--- a/docs/Data/Injector.md
+++ b/docs/Data/Injector.md
@@ -1,17 +1,20 @@
 ## Module Data.Injector
 
+#### `Prism`
+
+``` purescript
+type Prism s t a b = forall p f. (Choice p, Applicative f) => p a (f b) -> p s (f t)
+```
+
+Compatible with `Prism` from `purescript-lens`.
+
 #### `Injector`
 
 ``` purescript
-data Injector a b
-  = Injector (a -> b) (b -> Maybe a)
+type Injector s a = Prism a a s s
 ```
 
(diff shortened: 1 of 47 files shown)


cryogenian:
I like it :+1:

docs/Data/Injector.md
```diff
@@ -0,0 +1,52 @@
+## Module Data.Injector
+
+#### `Injector`
+
+``` purescript
+data Injector a b
+  = Injector (a -> b) (b -> Maybe a)
```
jonsterling: Is this isomorphic to a prism?

src/Halogen/Component.purs
```diff
@@ -44,19 +41,22 @@ import Control.Monad.State (State(), runState)
...
 import Control.Plus (Plus, empty)
 import qualified Control.Monad.State.Class as CMS
+import qualified Control.Monad.State.Trans as CMS
 
-import Data.Bifunctor (lmap, rmap)
+import Data.Bifunctor (bimap, lmap, rmap)
 import Data.Const (Const())
 import Data.Either (Either(..), either)
-import Data.Functor.Coproduct (Coproduct(), coproduct, left, right)
+import Data.Functor.Coproduct (Coproduct(..), runCoproduct, coproduct, left, right)
```
jonsterling: `runCoproduct` is not used in this file

docs/Data/Injector.md
garyb: Yeah, that's exactly what it is really - we don't have a dependency on any lens/optic stuff at the moment, but maybe we should just use that instead?

src/Halogen/Component.purs
garyb: Good spot, thanks. I need to get that unused imports warning working :)

docs/Data/Injector.md
jonsterling: Fair enough... I don't think I'd advocate adding a dependency on the lens/optics stuff, unless it would provide significant value.

Another option would be to use profunctor prisms here, so that it would be type-compatible with other implementations. I suppose that would still require imposing a dependency on `purescript-profunctors`; anyway, I don't have a strong opinion about any of this, and I tend to bias toward local re-implementation rather than dependency bloat.

What do you think?

garyb: I think profunctors is a transitive dependency (via coroutines or freet) so if it is, sounds like a plan!

jonsterling: Looks like it `purescript-profunctor` is indeed already included; if you like, I can look into making the switch.

garyb: That'd be great thanks. I'm working on a thing to make more efficient child updates currently so if you have time that'd be awesome.

jonsterling: no problem! Will give me a chance to remember how profunctors work.

garyb: I've just been failing miserably to explain them to @beckyconning :smile:

jonsterling: Since I don't think I have push access to your fork, here's a patch that you can apply: https://gist.github.com/jonsterling/21dc46bc4dff553878ea

garyb:
@jonsterling  Addressed those comments, I'll merge tomorrow if there are no other objections. Thanks for the review / lens work!

src/Data/Injector.purs
```diff
@@ -0,0 +1,64 @@
...
+import Data.Maybe.First (First(..), runFirst)
+import Data.Profunctor (Profunctor, dimap)
+import Data.Profunctor.Choice (Choice)
+import qualified Data.Profunctor.Choice as PF
+
+-- | Compatible with `Prism` from `purescript-lens`.
+type Prism s t a b = forall p f. (Choice p, Applicative f) => p a (f b) -> p s (f t)
+
+-- | Compatible with `PrismP` from `purescript-lens`.
+type Injector s a = Prism a a s s
```
puffnfresh: Nice! :+1:

jonsterling:
@garyb Excellent, everything looks good to me!

Issue #146, opened 2015-07-15
Integer attributes


cryogenian:
`rowSpan`, `colSpan`, `width` and `height`

garyb:
I've revised `rowSpan` and `colSpan` in #147, but left `width` and `height` for now, as there's also #141 for those.

jonsterling:
Resolved by #172

Issue #141, opened 2015-06-18
Add type for size value attributes


garyb:
Something like:
``` purescript
data Size = Pixels Number | Percent Number
```

Unfortunately a more general one for CSS including `em` sizes and the like wouldn't work here, as `width` and `height` attributes only support numbers and numbers followed by %.

garyb:
(Of course, using CSS is the better way, but there are some things that don't behave correctly unless sized with attributes - `canvas` for example, `iframe` in older browsers).

jdegoes:
:+1:

jonsterling:
Resolved by #172

Issue #93, opened 2015-04-24
Make `type` type safe?


garyb:
I started working on this, adding a sum type that covers all the valid `type` attributes for `input`, but then I realised that isn't really going to work unfortunately, as the `type` attribute is also usable with `button`, `command`, `embed`, `object`, `script`, etc.

Do we have any ideas for how something like this might work? Or is it not really worth the pain to make a "truly" type safe representation of HTML?

jdegoes:
I think there are lots of ways to make this work, what's the goal? A single `type` attribute which can cover all possible "types" of types (input, script, object, etc.)?

garyb:
I guess ideally there would be different `type`s - one covering everything would be a bit strange as you'd have mime types (for `script`) mixed in with `checkbox` and the like for `input`.

garyb:
Well, not full mime types exactly, but you know - `text/css`, `javascript`, and so on.

puffnfresh:
We should use different functions for each of these cases. We could possibly use a mutli-param type-class but that would be gross.

jdegoes:
> We should use different functions for each of these cases. 

Yeah, `mimeType`, `inputType`, etc. Doesn't have to match HTML naming convention exactly.

jonsterling:
Resolved by #172

PR #172, opened 2015-09-24
Add Halogen.HTML.{Elements,Properties}.Indexed

### LINKED ISSUE #151: use rows to improve attribute type safety
This might be worth exploring:

``` purescript
data InputTypeAttr
data ValueAttr
data ForAttr

data Attr (foo :: # *) = Attr
data Element = Element

input :: Array (Attr (type :: InputTypeAttr, value :: ValueAttr)) -> Element
input _ = Element

type_ :: forall a. Attr (type :: InputTypeAttr | a)
type_ = Attr

value :: forall a. Attr (value :: ValueAttr | a)
value = Attr

for_ :: forall a. Attr (for :: ForAttr | a)
for_ = Attr

-- type checks:
someInput1 = input [type_, value]
someInput2 = input [value]

-- fails type checking as `for_` is not an acceptable attribute for `input`
someInvalidInput = input [type_, value, for_]
```

Thoughts?

jonsterling:
I've added refinements (indexed versions) of properties and elements that can be used to ensure grammatical HTML by construction. I did this on top of the "untyped" versions rather than replacing them, because it is not possible for us to fix in advance the grammar and vocabulary of every conceivable HTML document.

Resolves #151, #141, #93, #146.


jonsterling:
Hmm, I probably should switch over some of the examples to use this.

paf31:
:+1: :smile: Very cool!

src/Halogen/HTML/Elements/Indexed.purs
```diff
@@ -0,0 +1,497 @@
...
+type Node r p i = Properties r i -> Array (HTML p i) -> HTML p i
+
+-- | An HTML element that does not admit children.
+type Leaf r p i = Properties r i -> HTML p i
+
+erases :: forall r i. Array (IProp r i) -> Array (Prop i)
+erases = Prelude.map erase
+
+a :: forall p i. Node (download :: I, href :: I, hreflang :: I, mediate :: I, rel :: I, target :: I, type :: I) p i
+a = E.a <<< erases
```
garyb: Perhaps we could use `unsafeCoerce` instead of using the erases no-op? It would mean there's zero overhead at all using `Indexed` over the non-`Indexed` types then:
```
a :: forall p i. Node (download :: I, href :: I, hreflang :: I, mediate :: I, rel :: I, target :: I, type :: I) p i
a = unsafeCoerce E.a
```
I think that should work anyway...

(The compiler isn't clever enough to inline things like `map id` yet)

jonsterling:
Just a quick update—I'd like to add some more stuff to this PR before it's merged.

- [ ] add refined versions of the event handler properties
- [ ] convert the examples to use this kit

docs/Halogen/HTML/Properties/Indexed.md
```diff
@@ -0,0 +1,236 @@
...
+#### `selected`
+
+``` purescript
+selected :: forall ρ i. Boolean -> IProp (selected :: I | ρ) i
+```
+
+#### `slot`
+
+``` purescript
+slot :: forall ρ i. String -> IProp (slot :: I | ρ) i
```
jonsterling: By the way, what is this attribute for? I wasn't aware that it was a standard HTML attribute.

garyb: Argh! It should be `placeholder`... I must have done this with an overzealous search and replace.

jonsterling: Aha!! That makes so much more sense :smile:

I'll switch it over.

jonsterling:
I think this may be ready for review now; it's a big changeset with a lot of (mostly) mindless code added. Please let me know if there's anything that I should change! Thanks :smile:

jonsterling:
By the way, one thing I noticed was kind of frustrating when using this was that only the versions of the element builders are exported from the "refined/indexed" modules which take an array of properties. This makes sense from one perspective, since there is nothing to refine about the `underscore_`'d ones, but it means that you end up needing to import both `*.Elements.Indexed` and `*.Elements`, and presumably qualify one of them, in order to get anything done.

I think it may make sense to export the underscore-suffixed ones from the indexed module as well, so that users who want that extra layer of confidence can use this version exclusively without getting mixed up.

What do folks think?

garyb:
That's a great idea actually, I'd forgotten about the fact it would be unnecessary to declare those varieties in the `Indexed` modules.

This all looks good to me so far. One other thing I'd like to address with this approach is there are a few attributes that could have some additional type-based help: `width` and `height` (as they can either accept pixel or percentage values, it would be nice to not have them as just stringly typed), #141, and there are several different types of `type` attribute, depending on the element, #93.

jonsterling:
@garyb Great points, thanks for the feedback!

With regard to `type`, indeed, the phantom key for `type` should be broken up into several different ones: `inputType`, `mediaType`, etc. It might be a nice idea also to use sum types for these!

jonsterling:
@garyb All right; I have made the modifications we discussed, I think. Should I squash these commits or do you prefer them separate?

garyb:
:tada:

Issue #168, opened 2015-09-16
Halogen performance (and new components)


themoritz:
I modified the components (ticker) example to show a grid of tickers. It turned out that for a 30x30 grid, increasing the counter of one ticker takes about 2s because every time (I guess?) the whole app is re-rendered.

I'm not very familiar with the internal workings of halogen, but my intuition was that if I only modify the local state of a component's child, I don't need to re-render the whole component.

In general, I'm wondering what I can expect performance-wise when using halogen as compared to similar UI frameworks (e.g., React).

The code can be found here:
https://github.com/themoritz/halogen-components-performance

jdegoes:
Halogen is definitely not optimized. The parent components know when their children change state, so it should be possible to be _very smart_ about caching the views. Currently that's not done.

Beyond just view caching, I'm sure general profiling & optimization could increase performance by at least 10x.

garyb:
It certainly shouldn't be re-rendering the whole DOM, as Halogen is backed by `virtual-dom` which operates on the same principle as React... so yeah, I guess it must be the internal rendering that is currently not smart. Caching of child components should be quite easy to implement so I'll definitely take a look at this once the composition PR is resolved.

Thanks for constructing the handy example too!

themoritz:
That makes sense. I think overall halogen performance is not a big concern, but the reconstruction of the whole virtual-dom after every input is what's causing our current project to be quite laggy.

Unfortunately at the moment I don't have the resources to look into caching myself. When would you estimate to have time working on this (given your other priorities)?

garyb:
Sorry I didn't get back to you about this earlier... I'm looking into it now though :)

kristianmandrup:
I guess this one can be closed now due to #173 ?

garyb:
Almost! The performance is significantly better, but I want to make some improvements to the internal `virtual-dom` rendering, so I'll close this once that is done too - as I mentioned in the other issue, this will be a post-release patch though as it won't affect the API.

garyb:
#196 eliminated the state monad from `virtual-dom`'s rendering so that should further improve the performance.

The next version of halogen may have its own layer of diffing when rendering HTML to virtual-dom, but for 0.5.x the obvious improvements are now complete.

themoritz:
I just checked out the performance of version 0.12, and I'm impressed. Lag in the 30x30 grid is virtually unnoticable, and with 2400 cells it's still totally fine to work with it. Amazing work guys!

PR #173, opened 2015-09-30
Memoize children and batch state updates

garyb:
Two changes here:

1. State modifications are now batched together so that rendering occurs either after the `Free HalogenF` has been fully processed, or when we switch from processing `StateF` to running the inner `Aff`. This means we render much less frequently now while preserving the property that running the inner `Aff` does not block rendering.
2. Child component HTML is now only re-rendered if the child has received a query since it was last rendered.

1 actually made the biggest difference here by far. 2 gives a micro improvement in the example I was testing with, but should have a more pronounced effect when children have complex HTML.

### DIFF
--- a/docs/Halogen/Component.md
+++ b/docs/Halogen/Component.md
@@ -46,10 +46,16 @@ A type alias for a component `eval` function that takes a value from the
 component's query algebra and returns a `Free` monad of the Halogen
 component algebra.
 
+#### `EvalP`
+
+``` purescript
+type EvalP i s s' f f' g p p' = Eval i s f (QueryF s s' f f' g p p')
+```
+
 #### `Peek`
 
 ``` purescript
-type Peek s s' f f' g p p' = PeekP s f (QueryF s s' f' g p p') (ChildF p f')
+type Peek s s' f f' g p p' = PeekP s f (QueryF s s' f f' g p p') (ChildF p f')
 ```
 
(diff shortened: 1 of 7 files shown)


jdegoes:
Has this been tested with #168?

garyb:
Yep, that was the main case I was using to test. It's still a little slow at large sizes but _much_ better than it was. It's about as slow with 2500+ components now as it was when there were 900 before. The delay with 900 now is a little random thanks to GC, so sometimes is still quite slow but sometimes almost instant.

After I fix the problem with the build here (assuming I can fix up the current method - adding the additional `f` type to `InstalledState` may cause some issues with composition for the multi-component example, but I've not looked deeply into it yet) I'm going to look at improving the construction of the virtual-dom tree to use ST or something to avoid allocating tons of objects when accumulating properties and building child node arrays.

After that, we're a little stuck for optimizations - from profiling there is no one thing that is slow, so really it's time to turn to the compiler then and look at that issue we discussed a while ago where we can improve codegen to prevent much unnecessary recomputation of pure values (purescript/purescript#687).

garyb:
Fixed up the types now /cc @jonsterling, @jdegoes

jonsterling:
@garyb Cool! I guess I'll say for the record that the number of things that have to be typed in that `eval` declaration in `ace/src/Main.purs` is bordering on the comedic, but I guess it is not too much more than before :wink: 

:+1:

PR #174, opened 2015-10-01
Make use of the EvalP synonym where appropriate

garyb:
/cc @jonsterling 

Marginally better I guess :smile:


jonsterling:
:+1: !!

PR #175, opened 2015-10-01
Fix some missing & incorrect refinements

jonsterling:
- add value property to select element
- add missing alt attr to img
- correct refinement of alt attr


garyb:
:+1:

PR #178, opened 2015-10-06
A few minor things to clean up

garyb:
- The change to `runUI` to only accept `Component` is okay because `ComponentP` is only used by components-with-children that have not yet had their children installed.
- Renaming `ρ` to `r` for the phantom rows is unfortunate, but seems like the best option currently to ensure "everyone" can build Halogen
- Forgot to add the `interpret` example to the stuff travis runs when I did that PR

### DIFF
--- a/src/Halogen/Component.purs
+++ b/src/Halogen/Component.purs
@@ -76,7 +76,7 @@ import Halogen.Query.SubscribeF (SubscribeF(), subscribeN, remapSubscribe, hoist
 newtype ComponentP s f g o p = Component
   { render :: State s (HTML p (f Unit))
   , eval   :: Eval f s f g
-  , peek   :: PeekP s f g o
+  , peek   :: PeekP o s f g
   }
 
 -- | A type alias for Halogen components where `peek` is not used.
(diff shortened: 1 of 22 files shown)


src/Halogen/Component.purs
```diff
@@ -232,7 +232,15 @@ install :: forall s s' f f' g p p'. (Plus g, Ord p)
         => ParentComponent s s' f f' g p p'
         -> (p -> ChildState s' f' g p')
         -> InstalledComponent s s' f f' g p p'
-install c f = Component { render: render c f, eval: eval, peek: const (pure unit) }
+install c f = installWithState c (const f)
+
+-- | A version of [`install`](#install) that gives us access to the parent's
+-- | state while installing children.
+installWithState :: forall s s' f f' g p p'. (Plus g, Ord p)
```
cryogenian: 'installWitState' is not exported

garyb: Oops! Thanks, I'll put a fix in later today.

garyb:
Does anyone have any other comments or can I merge this now? I've added a few more changes since the original PR, outlined in the commit comments.

/cc @beckyconning @cryogenian @jonsterling

jonsterling:
I think it looks good

garyb:
Cool, thanks :smile:

PR #182, opened 2015-10-09
Make `key` a global attribute

passy:
It's entirely possible that I'm misinterpreting something here, but it seems to me that it should be possible to set a `key` for every attribute. Is that assumption correct?

[Review on Reviewable](https://reviewable.io/reviews/slamdata/purescript-halogen/182)


garyb:
Ah right, yes I think so - `key` is a magic `virtual-dom` property, so it probably should be usable for every element.

Any thoughts/comments @jonsterling?

jonsterling:
I think it's fine...

garyb:
Thanks!

PR #187, opened 2015-10-13
Add explicit purescript-profunctor dependency, fix #186

### LINKED ISSUE #186: Missing purescript-profunctor dependency
It may be that the Halogen build ended up somehow having profunctors transitively, but when using purescript-halogen as a dependency elsewhere, the profunctor dependency does not appear to get installed.

There's a trade-off here... If you take advantage of a transitive dependency, you may be sorry when your immediate dependency changes its own dependencies. I think it is necessary to specify as a dependency anything which you will import, since transitive dependencies are not part of the signature / interface of a library, and may be changed out from under you (for instance, there is no versioning policy for when I remove a dependency from a library, but that will break any other package that relies on the transitive dep).

The trade-off is of course that it makes upgrading the ecosystem a bit more involved. However, I think that this is less undesirable than builds seemingly randomly going wrong.

jonsterling:


jonsterling:
:+1: from @garyb, so merging

PR #188, opened 2015-10-13
Fix rowspan/colspan prop indices / refinements

jonsterling:
There was a difference of case between the `Elements.Indexed` and `Properties.Indexed` modules.


garyb:
:+1:

jonsterling:
thanks @garyb!!

PR #189, opened 2015-10-13
Remove unnecessary `p`

garyb:
@jonsterling Here's the removal of the slot type I mentioned that can only be `p` or `Void`.

Cleans up the types a bit, but means I had to introduce `parentComponent` and `parentComponent'` to make the types work... kinda. `component'` can construct any kind of component, but by using the other 3 varieties you get better assistance when it comes to type errors.

Another thought I did consider is doing away with all the varieties and not even distinguishing between peeking and non-peeking components, but instead provide a helper `nopeek = const (pure unit)`. I'm not sure which approach is better.

### DIFF
--- a/docs/Halogen/Component.md
+++ b/docs/Halogen/Component.md
@@ -23,18 +23,24 @@ instance functorComponent :: Functor (ComponentP s f g o)
 #### `Component`
 
 ``` purescript
-type Component s f g = ComponentP s f g (Const Void)
+type Component s f g = ComponentP s f g (Const Void) Void
 ```
 
-A type alias for Halogen components where `peek` is not used.
+A type alias for self-contained Halogen components.
 
 #### `Render`
 
 ``` purescript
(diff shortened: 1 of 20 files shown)


jonsterling:
Nice, this does seem to clean things up a bit...

Regarding getting rid of the distinction between peeking & non-peeking, is it the case that this is currently part of the types?

garyb:
Yeah, so, `parentComponent'` is the only "specific" constructor that allows you to provide a `Peek` function when constructing a component, and it returns a corresponding `ParentComponentP` synonym.

(By "specific" I mean aside from `component'`, as that can do it, but it's the most generically typed constructor).

This distinction will vanish from the types if I get inline installation working, as there will be no need for the `ParentComponent` and `ParentComponentP` synonyms.

beckyconning:
:+1:

garyb:
Closing in favour of #191

PR #191, opened 2015-10-14
Inline component installation

### LINKED ISSUE #180: Persist child state (or access state during install)
What is the best way to persist child state, even when a particular slot isn't active? I'm thinking of the tabs use case, where you swap out a slot based on the currently active tab, but you want to persist state between tabs. I know you could do this by hiding tabs with CSS, but then everything is being rendered all the time.

Right now, if you were to swap out say 3 installed components based on a tab, the state is reset each time. If you want to persist that state, I'd think you have to move it into the parent state when swapping (so you'd query the child for its state). But you don't have access to the current state when installing, so you have no way of getting the old state when you swap it back in. The only thing I can think of is to put the state of the component into the slot, so you can thread it through when rendering (this seems problematic). Would it make sense to just give the installer access to the current state?

### LINKED ISSUE #181: Components that take children
What would be the best way to write a component that takes components as parameters? For example, a generic modal or popover that has its own state and lets you embed a form component or something (or even just HTML). With the slot scheme, because of the addressing, you want them to have as little information as possible to lookup the component state, so this means you can't really pass in children to the slot in `render`. I'm sure its likely possible to do when installing if the state is available, but it doesn't seem desirable to split up rendering like this.

garyb:
:boom:

This also subsumes #189.

Resolves #180. Resolves #181 (I think).

### DIFF
--- a/docs/Halogen/Component.md
+++ b/docs/Halogen/Component.md
@@ -23,18 +23,24 @@ instance functorComponent :: Functor (ComponentP s f g o)
 #### `Component`
 
 ``` purescript
-type Component s f g = ComponentP s f g (Const Void)
+type Component s f g = ComponentP s f g (Const Void) Void
 ```
 
-A type alias for Halogen components where `peek` is not used.
+A type alias for self-contained Halogen components.
 
 #### `Render`
 
 ``` purescript
-type Render s f p = s -> HTML p (f Unit)
+type Render s f = RenderP s f Void
 ```
 
 A type alias for a component `render` function.
+
+#### `RenderP`
+
+``` purescript
+type RenderP s f p = s -> HTML p (f Unit)
+```
 
 #### `Eval`
 
@@ -49,7 +55,7 @@ component algebra.
 #### `EvalP`
 
 ``` purescript
-type EvalP i s s' f f' g p p' = Eval i s f (QueryF s s' f f' g p p')
+type EvalP i s s' f f' g p = Eval i s f (QueryF s s' f f' g p)
 ```
 
 A convenience variation on `Eval` for parent components.
@@ -57,7 +63,7 @@ A convenience variation on `Eval` for parent components.
 #### `Peek`
 
 ``` purescript
-type Peek i s s' f f' g p p' = PeekP i s f (QueryF s s' f f' g p p')
+type Peek i s s' f f' g p = PeekP i s f (QueryF s s' f f' g p)
 ```
 
 A type alias for a component `peek` function that observes inputs to child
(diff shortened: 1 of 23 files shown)


garyb:
This may be of interest to you too @natefaubion

natefaubion:
:fireworks: :clap:

This looks really neat. I feel like it made a big difference in the examples (perceived complexity vs actual complexity). Can't wait to try it out.

jonsterling:
I'll review this within the next two hours

garyb:
No rush, I'm going to be out for that long at least.

One question I thought of: currently the initial state part of the slot is thunked, I wonder if it would be better to include the child component definition in there too. I can think of a hypothetical case where constructing the child component def might be a little expensive...

jdegoes:
On the thunking, any reason not to use `Lazy` just to help convey the intention?

jonsterling:
So, one question I had regarding inline component installation is whether the same level of modularity is preserved. That is, in the old system, you were essentially defining a component that has a bunch of holes in it (a "hypothetical component"), and then filling them in separately ("cut").

Is anything lost in the new approach? My instinct is that I should be able to recover the old way of doing things if I need to by turning the component into a function of components. Is that the right instinct?

beckyconning:
@jonsterling seems right to me : )

beckyconning:
:+1:

natefaubion:
I think there are two kinds of "slots". There's a slot to put in an addressable component (like for 3rd party widgets), and there's slots to be filled because they don't exist yet. In my tab example in https://github.com/slamdata/purescript-halogen/issues/181#issuecomment-147237331, it takes advantage of both. I wanted to inline-`install` the actual tab component, but then I needed to install the various placeholder slots (ie, `TabLabel`, `TabPanel`) after-the-fact because the number of each aren't necessarily static. I don't think this can be accomplished with purely inline install, but it would be great if you could do both in a single go.

garyb:
@jdegoes it hadn't occurred to me but sounds good :+1:

As for the "holes" approach that is less obvious now, the function approach  @jonsterling mentioned was exactly what I had in mind. Perhaps I should add an example for that.

natefaubion:
@garyb Does this only compile with psc master, now? (you aren't exporting RenderP, PeekP)

jonsterling:
OK; :+1:  from me! Nice work, @garyb.

garyb:
@natefaubion Ah, it's in a precarious position where it doesn't compile properly with either of them just now :cry:

It won't compile in master due to purescript/purescript#1526, and with 0.7.4.1 it fails when recompiling due to a bug in the externs generation that Jon uncovered last night. It should compile from clean with 0.7.4.1 though... I'll fix it so it always works that way for now.

garyb:
Fixed the synonyms now, I'll merge once Travis passes.

PR #193, opened 2015-10-15
Make slot initialisation lazier

garyb:
I tried to use `Lazy` here rather than a naked thunk, but unfortunately the type checker couldn't deal with that when you have a `ParentComponent` with a polymorphic `g`:

```
Cannot unify constrained type
    (Prelude.Functor g0) => Halogen.Component.Component<Example.Components.Ticker.TickState,Example.Components.Ticker.TickInput,g0>
  with type
    Halogen.Component.ComponentP _1 _2 _3 (Data.Const.Const Data.Void.Void) Data.Void.Void
```

A shame... but aside from that I prefer this tweaked approach for two reasons: 

1. It means we don't have to pay when defining the `HTML` if a child component is constructed rather than being a reference.
2. Making it a record with a field named `initialState` reinforces the fact that you cannot change the child component's state by altering this value alone.

I think this is the last change I want to make to the code for 0.5. I just need to update the documentation now, and rework some sections of the README to reflect these install changes. I'll make one last PR with all that stuff in together.

@jonsterling @beckyconning @cryogenian

### DIFF
--- a/docs/Halogen/Component.md
+++ b/docs/Halogen/Component.md
@@ -74,7 +74,7 @@ A variation on `Render` for parent components.
 
 ``` purescript
 data SlotConstructor s' f' g p
-  = SlotConstructor p (Component s' f' g) (Unit -> s')
+  = SlotConstructor p (Unit -> { component :: Component s' f' g, initialState :: s' })
 ```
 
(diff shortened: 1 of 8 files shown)


jonsterling:
Looks good to me! :+1:

PR #196, opened 2015-10-17
Remove memo key from lifecycle props

natefaubion:
The cleans up the lifecycle hooks so they don't need a memo key.

Virtual-dom provides a third parameter to `hook`/`unhook` that tells you which hook it is transitioning from or to. When it has not transitioned from another hook, we can call the initializer. And when it is not transitioning to another hook, we can call the finalizer.

As far as I can tell, this works fine. Both the ace example, and a new example I added, work as expected and you don't get the infinite recursion.

I also added `initializer`/`finalizer` to `Properties` and created `Indexed` variants of them. Previously you couldn't use them as-is with the `Indexed` HTML elements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -483,7 +483,7 @@ The [Ace editor exampe](examples/ace) illustrates the creation of a basic widget
 
 A widget is constructed using the standard `component` function but taking advantage of some features that are not used when defining a normal component.
 
-Even though we’re embedding an external component in our page we still need to render a `HTML` value, so the suggested way of doing this is to generate a single empty element such as a `div` and then attach the special `Initializer` property to it to handle setup once the element has been rendered to the page. From the Ace example:
+Even though we’re embedding an external component in our page we still need to render a `HTML` value, so the suggested way of doing this is to generate a single empty element such as a `div` and then attach the special `initializer` property to it to handle setup once the element has been rendered to the page. From the Ace example:
 
 ``` purescript
 ace :: forall eff. String -> Component AceState AceQuery (Aff (AceEffects eff))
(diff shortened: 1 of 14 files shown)


garyb:
Interesting! And very cool. This was one of the wart-iest things left in Halogen, so eliminating it is fantastic - the removal of the state monad during rendering should give a nice boost to performance too, one of the major reasons I wanted to get rid.

So if I understand rightly, with this approach after you set an initializer it can never be changed once that element has been rendered to the page? I think that's a perfectly reasonable constraint, I just want to make sure I understand what's going on here.

natefaubion:
> So if I understand rightly, with this approach after you set an initializer it can never be changed once that element has been rendered to the page? I think that's a perfectly reasonable constraint, I just want to make sure I understand what's going on here.

Yes that's correct. An initializer is only fired on an element once. If you change it after it has fired, it has no effect. I think this is reasonable. 

Personally, I don't think lifecycle handlers belong in the virtual DOM as they are a property of the component, and diffing them doesn't really make sense. For the `virtual-dom` library itself, this means that each Halogen component would translate to `Widget` that controls rendering of it's children, and calling it's lifecycle methods. I think something like that would be reasonably achievable, but not necessary for now.

I also might like to see more fine-grained lifecycle handlers. What we have now is basically `componentWillMount` and `componentWillUnmount` from React. In the initializer, I don't think you're guaranteed that the node is already in the DOM (due to how `Aff` may be synchronous up to a point, but I may be wrong here). I did some work on this but backed it out because it was pretty hideous (passing a mutable array around, and having the hooks push handlers to it during patching).

aspidites:
@natefaubion, it appears that in making these changes, you have removed various bits of documentation. While I respect that the documentation as it was would no longer be accurate with your changes, I think there is value in replacing it with up-to-date documentation.

garyb:
@natefaubion great! And yes that's how the keyed initializer would behave if removed and restored too.

I agree the lifecycle handlers should not be diffed, and now you've shown this technique I have an idea about how to make that happen (kinda - it'll still happen in virtual-dom, but now I can do something with Halogen to make them more integral to a widget component definition). I'll open a PR for discussion.

natefaubion:
@aspidites Sure, I'll update the relevant portions of the docs. I think there are a few unused imports to clean up now as well due to the removal of `State`.

garyb:
Actually let's just stick with this for now :smile: I'd like to get this change in asap before people start relying on the old behaviour, as this is going to be released as `v0.5.1` even though it's a breaking change.

If you're in there updating things could you rename `Input` to `Query` in the lifecycle example too please?

natefaubion:
I've tidied it up and made the changes to the examples and docs. Let me know if there's anything else.

garyb:
Looks good, thanks!

PR #201, opened 2015-10-19
Added appendTo to Halogen.Util

zudov:
Users often might want to append to something else than `body`.

Of course the problem of "Get an element and append another element to it", should be solved by some other libraries (purescript-simple-dom, purescript-jquery), but it seems that at the moment we end up with `purescript-dom` which is a bit unwieldy for such a simple task.


garyb:
May as well re-implement `appendToBody` as `appendTo "body"` if we're doing this? :)

Regarding your second point, I agree, that's why I only provided `appendToBody` originally - but `querySelector` is a perfectly valid thing to appeal to here, I didn't think of it for some reason, so :+1:

zudov:
True

garyb:
Thanks!

PR #202, opened 2015-10-20
Helper functions for component slots

parsonsmatt:
In the process of getting my tutorial posts up to speed, I found myself wanting a few helper functions in just about every component I was making.

If these are :+1: to your tastes, then I'll be glad to extend them to the `slot'` functions and write up corresponding documentation :)

### DIFF
--- a/examples/todo/src/Component/List.purs
+++ b/examples/todo/src/Component/List.purs
@@ -41,7 +41,7 @@ list = parentComponent' render eval peek
            , H.p_ [ H.button [ E.onClick (E.input_ NewTask) ]
                              [ H.text "New Task" ]
                   ]
-           , H.ul_ (map (H.indexedSlot TaskSlot task initialTask)) st.tasks)
+           , H.ul_ (map (H.indexedSlot TaskSlot task initialTask) st.tasks)
            , H.p_ [ H.text $ show st.numCompleted ++ " / " ++ show (length st.tasks) ++ " complete" ]
            ]
 
--- a/examples/todo/src/Component/List.purs
+++ b/examples/todo/src/Component/List.purs
@@ -41,12 +41,9 @@ list = parentComponent' render eval peek
            , H.p_ [ H.button [ E.onClick (E.input_ NewTask) ]
                              [ H.text "New Task" ]
                   ]
-           , H.ul_ (map renderTask st.tasks)
+           , H.ul_ (map (H.indexedSlot TaskSlot task initialTask)) st.tasks)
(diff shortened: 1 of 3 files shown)


garyb:
I'm not too sure about these - I definitely see the motivation, I just worry a bit that they will be misused.

The idea behind using the thunk is it means you don't have to pay for constructing the component and initial state every time the parent is rendered, and also using the name `initialState` in the record reinforces the fact that the value is only used for the slot's initial state and that changing it will have no effect on the child component.

I'll have a think! Any thoughts from @beckyconning / @cryogenian / @jdegoes / @jonsterling would also be appreciated.

beckyconning:
Maybe this is more of a problem with the lack of something like

```
mapWithIndex :: forall f a b. (Functor f) => (Int -> a -> b) -> f a -> f b
```

rather than an issue with Halogen?

beckyconning:
I agree that there is a real concern here though :+1:

jonsterling:
I appreciate that these may have cleaned up some code in an application project, but I think that the convenience functions are not of broad enough applicability to include in the Halogen project itself. (Just my 2c.)

cryogenian:
I like `indexedSlot` and don't like `makeSlot` :smile: 

+ Generic way to make slots indexed is cool feature. But probably we should make something like `Halogen.Component.List` where this helper function will live (with generic `List` component) 
+ As far as I know `slot` signature (thunk and object) is made this way to make obvious that changing `initalState` won't cause rerender.

parsonsmatt:
I was under the impression that the calls delegating to `slot` would preserve the performance benefits of the thunk -- but I'm probably thinking about it with Haskell's evaluation strategy on the mind, and could certainly be wrong. If there's no way to avoid the `\_ -> { component: ....` line noise then that does restrict the utility of the `mkSlot` helper as something to be included in the library. I do feel like it will be reimplemented many times across many components and projects -- I've already implemented something similar at least three times, and there's one in the todolist example as well.

`mapWithIndex` would be nice, but the general use case I had in mind is the `Array Int` used as a means of keeping track of slots (like `renderTask` in the todolist), where the index is potentially less important than the actual item in the array. I imagine that will be a pretty common use case for this sort of thing. The name is striking me as poorly chosen now -- it's not the index wrt array, it's the index wrt child component! Could probably do away with that entirely if the `slot` function had the arguments in a different order: `map (slot component initialState <<< Slot)` accomplishes essentially the same thing.

jdegoes:
@parsonsmatt Thanks for the PR, btw. :smile:

PR #204, opened 2015-10-20
Finer grained synonyms for better reusability

garyb:
This keeps the old `Render` etc. synonyms but introduces new synonyms that are helpful for cases where you have `render`, `peek`, or `eval` split over multiple functions.

### DIFF
--- a/examples/todo/src/Component/List.purs
+++ b/examples/todo/src/Component/List.purs
@@ -9,6 +9,7 @@ import Data.Array (snoc, filter, length)
 import Data.Functor.Coproduct (Coproduct())
 import Data.Generic (Generic, gEq, gCompare)
 import Data.Maybe (Maybe(..), fromMaybe)
+import Data.NaturalTransformation (Natural())
 
 import Halogen
 import qualified Halogen.HTML.Indexed as H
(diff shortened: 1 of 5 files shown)


beckyconning:
:+1: Exciting! This will help identify "Render functions" and make writing subexpressions of render functions easier. It will also help people new to Halogen understand the subexpressions of a component by including `->` in their types.

People asking what the synonyms are happens a lot when I am showing Halogen apps to people and showing the full synonyms as they were previously can be confusing to people. Now I expect people will be perfectly happy with the understanding that for example `eval` is a function from some state to some `CompoentHTML` or `ParentHTML`.

jonsterling:
wow, this is way better!

examples/todo/src/Component/List.purs
```diff
@@ -48,12 +49,12 @@ list = parentComponent' render eval peek
   renderTask :: TaskId -> HTML (SlotConstructor Task TaskQuery g TaskSlot) ListQuery
```
beckyconning: Should this type signature use one of the new synonyms?

garyb:
@beckyconning updated TODO example to use the new synonyms everwhere.

PR #205, opened 2015-10-20
Add liftEff and liftAff variants for parent components

garyb:
/cc @cryogenian 

These types are nasty, but we can't refer to `QueryF` in here without having a circular dependency in the modules.

### DIFF
--- a/docs/Halogen/Query.md
+++ b/docs/Halogen/Query.md
@@ -173,6 +173,15 @@ A convenience function for lifting an `Aff` action directly into a
 `Free HalogenF` when there is a `MonadAff` instance for the current `g`,
 without the need to use `liftH $ liftAff $ ...`.
 
+#### `liftAff''`
+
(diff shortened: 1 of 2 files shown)


garyb:
After chatting on Slack looks like these functions aren't too useful as inference can't use them without explicit typing, which pretty much defeats the point of these helpers.

Constraints being the problem as ever:

``` purescript
No instance found for

  Control.Monad.Eff.Class.MonadEff _1822
                                   (Aff ( now :: Now
                                        , file :: READ_FILE
                                        , err :: EXCEPTION
                                        , zClipboard :: ZCLIPBOARD
                                        , console :: CONSOLE
                                        , ref :: REF
                                        , ajax :: AJAX
                                        , avar :: AVAR
                                        , random :: RANDOM
                                        , dom :: DOM
                                        )
                                   )
```

PR #211, opened 2015-10-24
Component transforms

garyb:
- Exposes the `transform` function that allows mapping over a component's state and query algebra
- Hides the internal representation of `SlotConstructor`

This second change is to enable us to perform a trick to allow us to hide the state and query algebra of child components in certain conditions:

``` purescript
module PrivateComponent where

data PrivateState
data PrivateQuery a

type State a = { ... private: a ... }
data Query a = ...

mkPrivateComponent :: forall s f g
                    . Component (State s) (Coproduct Query f) g
                   -> s
                   -> SlotConstructor (State PrivateState) (Coproduct Query PrivateQuery) g
```

This is possible by using transform on the component to pack/unpack `PrivateState` and `PrivateQuery` values. The unpacking of these values is partial, but as long as we only export the types and not the data constructors for `PrivateState` and `PrivateQuery` this is safe as we can ensure that there is no way for these values to originate from anywhere other than within the component itself, therefore we know we have the right type inside `PrivateQuery`.

The change to hide `SlotConstructor`'s representation ensures that we can't make two `PrivateComponent`s with different internal state values and then extract their `initialState` values and swap them over or something. Basically ensuring we keep the guarantee that there is no place a `PrivateState` value is accessible except during the transformation of the private component.

@jonsterling I decided tweaking Halogen probably was worth it to achieve this - it seems likely we might want to do this existential children trick in other places too, and since this doesn't involve breaking `H.slot`'s interface or introducing yet another variant we _almost_ had the means to it already anyway (`H.slot` creates a `SlotConstructor`, so we'll just bypass that when making a parent component for children using this trick and use `HTML`'s `Slot` constructor directly).

### DIFF
--- a/docs/Halogen/Component.md
+++ b/docs/Halogen/Component.md
@@ -238,6 +238,12 @@ liftQuery :: forall s s' f f' g p. (Functor g) => EvalParent (QueryF s s' f f' g
 Lifts a value in the `QueryF` algebra into the monad used by a component's
 `eval` function.
 
+#### `transform`
+
+``` purescript
+transform :: forall s s' f f' g. (Functor g) => (s -> s') -> (s' -> s) -> Natural f f' -> Natural f' f -> Component s f g -> Component s' f' g
+```
+
 #### `transformChild`
 
 ``` purescript
--- a/docs/Halogen/Component.md
+++ b/docs/Halogen/Component.md
@@ -96,10 +96,18 @@ same form but the type representation is different.
 
 ``` purescript
 data SlotConstructor s' f' g p
-  = SlotConstructor p (Unit -> { component :: Component s' f' g, initialState :: s' })
 ```
 
 The type used for slots in the HTML rendered by parent components.
+
+#### `slotConstructor`
+
+``` purescript
+slotConstructor :: forall s f g p. p -> (Unit -> { component :: Component s f g, initialState :: s }) -> SlotConstructor s f g p
+```
+
+Creates a `SlotConstructor` from a slot address value and thunked
+constructor.
 
 #### `ParentDSL`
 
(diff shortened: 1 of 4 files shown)


garyb:
Actually, we could probably provide a module which does exactly what I outlined above, that way neither we nor nobody else will have to do the work to make these partially-private components.

garyb:
@jonsterling could you review please? I think you're probably best placed to review given our discussions up to this point.

jonsterling:
I'm happy to do so

--jon

garyb:
Sigh. Thought of another case that makes this unsafe: `peek` lets us get hold of a `PrivateQuery` value in a way we can do heinous things with it, like storing it in the state and/or using it to query another child, so scratch this too. Will have to go with the `FibredEq` approach I think.

jonsterling:
@garyb Damn it... ugh.

garyb:
I guess we can probably still provide something like this private module to takes care of the plumbing in a generic way though. We can't do it via `transform` due to being unable to provide a safe `(s -> s')` and `(s' -> s)` (no `Plus`), but there's probably some core piece we can extract that means we don't have to implement the large chunk of code we have in `eval` currently. And the user would have to provide their own φ of course.

garyb:
The only problem being I'm not sure about some of those coercions/failure cases we have in there either...

jonsterling:
@garyb I think that the coercions in that nasty code I wrote are safe, and with some effort, could probably be replaced with actual uses of the passed leibniz identification. It may help if we try and disentangle the `CellState` and `Def` types a little bit.

garyb:
Yeah, definitely - I had been looking at that too. We can pass `Def` into the component instead, although we need to add `stateTag` to the the `CellState` then, I think.

PR #212, opened 2015-10-24
Alter IProp representation

### LINKED ISSUE #206: unecessary IProp type synonym?
In `Halogen/HTML/Properties/Indexed.purs`, the following type alias exists:

```
-- | The phantom row `r` can be thought of as a context which is synthesized in the
-- | course of constructing a refined HTML expression.
type IProp (r :: # *) i = Unsafe.IProp r i
```

`Unsafe.IProp` is defined in `Halogen/HTML/Properties/Indexed/Unsafe.purs` like the following:

```
-- | The phantom row `r` can be thought of as a context which is synthesized in the
-- | course of constructing a refined HTML expression.
newtype IProp (r :: # *) i = IProp (Prop i)
```

I'm not sure I understand why IProp type synonym exists.  Couldn't `Halogen/HTML/Properties/Indexed.purs` just re-export the `Unsafe.IProp` type (and not export the the data constructor)?  Then `Halogen/HTML/Properties/Indexed/Unsafe.purs` could export both the `Unsafe.IProp` type constructor and data constructor.

Maybe this wouldn't work for some reason?

garyb:
This eliminates the need for the `Properties.Indexed.Unsafe` module defining `IProp` separately from `Properties.Indexed` and also uses switches from using `refine <<< x` in favour of directly coercing `x` so that there is no runtime overhead (we made the same change to `Elements.Indexed` a while back).

Resolves #206


jonsterling:
looks good to me!

PR #213, opened 2015-10-24
Add style to global attrs, add autocomplete property

### LINKED ISSUE #190: "autocomplete" property
The form and input elements make mention of the "autocomplete" option but there is no corresponding property in purescript-halogen/src/Halogen/HTML/Properties/Indexed.purs.

garyb:
Resolves #190 and makes it possible to use the `style` attribute! There's a corresponding PR in `purescript-halogen-css`.


jonsterling:
:+1:

PR #215, opened 2015-10-26
Add no-op to HalogenF

jonsterling:
@garyb I forgot to open a PR for this; anyway, this is just a preliminary implementation to see if it solves the problem you were having. If we actually want to use this, I can go and make the other changes we discussed (namely, making `HalogenF` a proper sum type and collapsing all the coproducts into it).

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -30,6 +30,7 @@
     "purescript-dom": "^0.2.6",
     "purescript-foreign": "^0.7.0",
     "purescript-free": "^0.9.0",
+    "purescript-inject": "^0.3.0",
     "purescript-maps": "^0.5.0",
     "purescript-nullable": "^0.2.0",
     "purescript-profunctor": "^0.3.1",
--- a/docs/Halogen/Component.md
+++ b/docs/Halogen/Component.md
@@ -130,7 +130,7 @@ components.
 #### `parentComponent`
 
 ``` purescript
-parentComponent :: forall s s' f f' g p. (Plus g, Ord p) => RenderParent s s' f f' g p -> EvalParent f s s' f f' g p -> Component (InstalledState s s' f f' g p) (Coproduct f (ChildF p f')) g
+parentComponent :: forall s s' f f' g p. (Functor g, Ord p) => RenderParent s s' f f' g p -> EvalParent f s s' f f' g p -> Component (InstalledState s s' f f' g p) (Coproduct f (ChildF p f')) g
 ```
 
 Builds a component that may contain child components.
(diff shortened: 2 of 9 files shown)


garyb:
No problem! I'd started using the branch directly from your repo anyway. It's looking good so far!

jonsterling:
Oh dear lord, I think I killed it... :wink: Will need to add some `Inject` instances or something.

garyb:
Restarted the build as it wasn't obvious what the error was to me - just a hiccup?

And actually this isn't quite what I had in mind either! I was thinking we could just squash everything into a single `HalogenF` type:

``` purescript
data HalogenF s f g a
  = Get (s -> a)
  | Modify (s -> s) a
  | Subscribe (EventSource f g) a
  | Empty
  | Run (g a) -- ok, not sure what this should be called... QueryHF in your version I guess?
```

Not sure if this would make the implementation any simpler though, so no biggie if not.

jonsterling:
@garyb Oh, sorry, I had misunderstood! I could certainly try something like this if you want. Do you think it will result in a performance improvement?

garyb:
I'm not really sure it will make that much of a difference - just thought it might have made the implementation of interpreters/natural transformations for `HalogenF` a bit simpler.

jonsterling:
@garyb Ah, I see. I think that the baby step that I took has simplified the interpreters and transformations a fair bit (or, if not simplified, at least clarified). I can see if the further-squashed version you propose would make things even clearer—it seems like it might, at the expense of a bit of implementation modularity.

jonsterling:
@garyb I thought about it a little more, and I think that squashing this further into a single type will complicate things more than simplify them; for instance, currently we are able to hide the implementations of various natural transformations inside modules like `*.State` and `*.Subscribe`, and just use them as we please... 

Can you think of an example of where squashing it into the single type would make things easier or simpler? I probably missed a number of thing, just since I'm not super familiar with this codebase.

jdegoes:
> Can you think of an example of where squashing it into the single type would make things easier or simpler? 

It would make things faster. If there are modularity tradeoffs, however, it may not be worth it.

jonsterling:
@jdegoes If it's a significant performance improvement, I think it's probably worth it. Halogen's internals are gnarly enough that I won't cry too many tears about losing a little bit of implementation modularity (which is not as important as interface modularity anyway).

garyb:
I didn't have anything in mind in particular, just wondered if it would help. It sounds like this is probably the right level of fusion to stop at by the sounds of it :smile: (unless we have some serious performance issue it would help with, but I doubt that, queries are never going to be super fast given that we're in Aff, etc.)

jonsterling:
All right; I'll leave it up to @garyb :smile:

jdegoes:
By the way, I suggest instead of naming the empty constructor `Empty`, you name it `Halt` (since that's effectively what it is).

jonsterling:
@jdegoes Good idea, I suppose it is terminal...

garyb:
Once this passes :+1: for jondentation, :+1: for the `HalogenF` changes.

jonsterling:
thanks @garyb ! :smile:

PR #216, opened 2015-10-28
Update docs and examples

garyb:
- Switched all the examples over to use the new type synonyms and jondentation
- Updated the docs to explain the new synonyms a bit
- Relaxed `Plus` constraints for `Functor` where possible now we can do that
- Tweak to the wording about HalogenF to remove mention of coproduct
- `Halogen` re-exports `Natural` as per @natefaubion's suggestion - I wasn't sure about this idea at first, but it is pretty useful and does no harm, it won't break anything that explicitly imports `Natural` alongside `Halogen`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,17 +42,21 @@ data Query a
   | GetState (Boolean -> a)
 
 -- | The component definition
-myComponent :: forall g. Component State Query g
+myComponent :: forall g. (Functor g) => Component State Query g
 myComponent = component render eval
   where
 
-  render :: Render State Query
+  render :: State -> ComponentHTML Query
   render state =
-    H.div_ [ H.button [ E.onClick (E.input_ ToggleState) ]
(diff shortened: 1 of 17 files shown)


jonsterling:
This looks great to me, thanks for doing it! Nice catch on removing those `Plus` constraints too.

garyb:
Cool, thanks for the review!

natefaubion:
I'm a little late on this, but would it make sense to change `Eval i s f g` to just `Eval s f g`? I'd think if `i` and `f` were to differ, you'd be using `Natural` anyway, and it seems like the vast majority of `Eval` types are going to agree on `i` and `f`.

garyb:
Yeah, probably, and same goes for `Peek` too. I considered removing them entirely, but I think they are helpful from a "documentation" point of view for the arguments to the component construction functions, and then just didn't really look any further after that. Will roll that into the next release perhaps!

PR #218, opened 2015-10-28
Deep-peeking example

### LINKED ISSUE #217: Communication from grandchild to grandparent
It is not clear to me how communication from a grandchild component to a grandparent component is working/will be working in halogen.

Communication from grandparent to grandchild (both query and action) is not a problem, since, when the parent is evaluating the actions that are called from the grandchild, it can always call actions on the grandchildren.

However, when peeking (this is how currently communication from child to parent is implemented), the parent has no way of issuing actions in its own query algebra that can then be peeked into by the grandparent, and therefore, messages from the child cannot "bubble" up the hierarchy.

What are the options at the moment?

* @garyb mentions in #184 that components can peek arbitrarily deep, but I cannot figure out how this would look like. If this will be the best practice way of doing it, adding an example would be great.

* Maybe there is some way to feed input into the root component (similar to the driver, but from any descendant component in the app), so that it can then trickle down the hierarchy? Independent of the technical possibilities, this would be a nice feature, since it would cover many use cases without the need to carry messages up to the root and then down to other components.

garyb:
Resolves #217 /cc @themoritz

If things are still unclear please ask away and I'll try to improve the example to better illustrate.

### DIFF
--- a/examples/deep-peek/.gitignore
+++ b/examples/deep-peek/.gitignore
@@ -1,0 +1,3 @@
+/dist/example.js
+/output
+/bower_components
--- a/examples/deep-peek/README.md
+++ b/examples/deep-peek/README.md
@@ -1,0 +1,13 @@
+## Deep peeking example
+
+This example illustrates a parent component peeking on a 3-level deep child component child component.
+
+### Building
+
+From the root of the Halogen project:
+
+```
+$ npm run example-deep-peek
+```
+
+The code will be built as `example.js` in the `examples/deep-peek/dist` directory within the example, runnable by opening the corresponding `index.html`.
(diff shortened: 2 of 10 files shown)


beckyconning:
:+1:

themoritz:
Thanks! Makes all sense to me. The interesting thing here is to go into the `q` using `coproduct`.

garyb:
Yep! It's a little obscured by using the `ListQueryP` synonym in the argument to `peek`, but if you look it up, it makes things much clearer:

``` purescript
type ListQueryP = Coproduct ListQuery (ChildF TickSlot TickQuery)
```

This additional coproduct appears any time you have a child component that is also a parent component, as it's the query algebra type you get back from a usage of `parentComponent'`.

PR #219, opened 2015-10-28
Clean up unnecessary stuff

garyb:
Surprise! :tada: 40% of `Component`'s type variable are unnecessary.

### DIFF
--- a/src/Halogen/Component.purs
+++ b/src/Halogen/Component.purs
@@ -1,6 +1,5 @@
 module Halogen.Component
-  ( ComponentP()
-  , Component()
+  ( Component()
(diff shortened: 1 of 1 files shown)


beckyconning:
😍

jonsterling:
wow!

jdegoes:
:gift:

garyb:
This happened in two phases:

1. When we removed `installL`/`installR` etc. storing `peek` in the component became unnecessary, and therefore the `o` type variable.
2. When we moved to inline installation the `p` type variable became unnecessary, as the `render` type's `p` is always `Void` after install.

PR #223, opened 2015-10-29
Fix behaviour of `subscribe`

### LINKED ISSUE #222: Can only subscribe to one event
If I subscribe to a second event in the ace example, only the first handler is attached.

```purescript
subscribe $ eventSource_ (Editor.onFocus editor) $ do
  pure $ action (ChangeText "focus")
subscribe $ eventSource_ (Session.onChange session) $ do
  text <- liftEff $ Editor.getValue editor
  pure $ action (ChangeText text)
```

In the above example, the `onFocus` handler works, but if I put `onChange` first, only that one works.

garyb:
Resolves #222 

Without the `forkAff`, using `subscribeN` this way will block forever. The hint being in the name of the `forever` combinator :wink:.


beckyconning:
:+1:

PR #229, opened 2015-11-01
Update README.md

aspidites:
Typo? This project itself doesn't appear to have a project.json file, and I can't seem to find any reference to `project.json` anywhere except in asp.net projects. `npm init`, however, does create a `package.json` file, which is what I assume is being referred to here.


garyb:
You're quite right. Thanks!

PR #230, opened 2015-11-02
Provide a more useful transform function

garyb:

This works the non-cheating partial component transfom I demonstrated the other day into Halogen itself. The reason I chose to add it here is we can also use it to make the `ChildPath` transforms safe, as previously this was not the case (although it would have involved some work to make them fail).

I mean, it's still not super awesome, but at least we have `HaltHF` as our failure case now rather than a runtime error from `fromJust`... even though generally `HaltHF` gets interpreted as Aff's `empty` which means it's still a runtime error, but it's "more right" than what we had before.

We could switch the functions out for `PrismP`, but we don't have a lens dependency yet, so maybe the plain functions are better?


garyb:
There are a bunch of less relevant changes in here too (unused imports, type variables), so here's the actual important part: https://github.com/slamdata/purescript-halogen/pull/230/files#diff-526e403138d983b7d4d7baf496755cb6R342

jonsterling:
Thanks Gary! I'll review this this morning.

jonsterling:
So, if we are certain that Halogen itself will never try sending the wrong (state, query) to a child (do we know this?), the only way for a user to trigger a halt is to perform some kind of "switcharoo", right?

garyb:
I'm not sure it's quite as solid as that unfortunately...

Providing an invalid `initialState` is probably the worst thing that can happen, as the component will not be rendered at all (`H.text ""`), and then attempting to query it will result in a halt, however we can take steps to reduce the likelihood of that happening (when using `transform`, return a `SlotConstructor` as this will have the `initialState` and `component` values baked in).

Querying a child component with an invalid query (something the prism returns `Nothing` for) will also result in a halt. I think this may be fix-upabble into a `Nothing` result (querying component externally always returns a `Maybe`'d result), but I need to look into that a bit more. This would improve the situation greatly though, so I'll see what I can do.

Neither of those cases can arise from Halogen itself though.

jonsterling:
Ok; I think I'm OK this, since Halogen itself apparently cannot such cause a fault. Would you mind adding to the doc-comment that this is the case (it may seem redundant, but it could be helpful to folks who are looking at using it)?

It would be really nice to fix this so that a bad query turns into `Nothing`, but we can probably look into that later on.

garyb:
Good thinking, how's this? https://github.com/garyb/purescript-halogen/commit/40617e946eb9f3d02fbe680d9361b0270b462bde#diff-931cc7fe853afe2f231a64bef049aaf8R238

jonsterling:
@garyb Looks great, that's just what I wanted! There's a comma splice though; could you replace the comma after `to halt` with a semicolon?

garyb:
Oops :smile:

jonsterling:
Merging now, :+1: . Separately, I will investigate interpreting bad queries as `Nothing` (or perhaps, a more informative thing).

PR #231, opened 2015-11-02
Fix typo (?) in the README examples

jonsterling:
There is no longer anything called `actionF`; I think that `action` is meant.

(This is unrelated to his PR, but I wonder whether the `action` thing is a good idea, since it is just a synonym for `\f -> f unit` IIRC; it kind of increases the perceived surface area of Halogen that needs to be learned. What do you think?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -586,7 +586,7 @@ Both of these have a rather difficult looking type signatures but are fairly str
 data Query a = TextCopied String a
 
 eventSource (Editor.onCopy editor) \text -> do
-  pure $ actionF (TextCopied text)
+  pure $ action (TextCopied text)
 ```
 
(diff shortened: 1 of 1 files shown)


garyb:
:+1: `actionF` is ancient history from when we still had `F` and `FC` versions of everything for `Coyoneda` :smile:

I'm not sure on the `action` thing. It just exists as a counterpart to `request` really, which is `\f -> f id` which is a little trickier to explain. From a mini DSL side of things I think maybe there's some utility in it, `query (action Go)` / `query (request Get)` is perhaps more helpful than `query (Go unit)` / `query (Get id)`? I don't know.

jonsterling:
Ah, keeping it for symmetry makes sense, I think.

PR #238, opened 2015-11-04
Synchronous variant of `appendTo`.

fisx:
See commit message.

I did this to solve an immediate issue I had; if you don't like the code as is, but have ideas what you would like instead, I am happy to give this a bit more work.

### DIFF
--- a/docs/Halogen/Util.md
+++ b/docs/Halogen/Util.md
@@ -7,7 +7,7 @@ appendTo :: forall m eff. (MonadEff (dom :: DOM | eff) m) => String -> HTMLEleme
 ```
 
 A utility for appending an `HTMLElement` to the element selected using querySelector
-element once the document has loaded.
+element (synchronously).
 
 #### `appendToBody`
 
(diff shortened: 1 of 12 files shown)


zudov:
Heh. Initially we only had a synchronous version, but changed it to asynchronous to make it smoother.

IMHO, it would be ideal to make `appendTo` synchronous and just provide an `onLoad` thingy.

fisx:
Done.  It's a bit more disruptive, but I like this solution better.

garyb:
Can we keep the `MonadEff` constraint on `appendTo` and `appendToBody` please?

fisx:
Yes, I wanted to, but I got an error that I don't remember.  I'll try again later and come back with more data if I fail again.

garyb:
Ok, thanks! I think all it needs is a sprinkling of `liftEff` to make it work out, although the `callback` will probably have to be `Eff` as I didn't use the constraint when defining `addEventListener`.

fisx:
Hm, that really wasn't that hard.  (-:

fisx:
(I looked at the travis log, but don't understand it.  It looks like `npm install` choked on installing...  what?)

zudov:
Try to restart the build (close pull request and then open it)

jonsterling:
No need to close the PR, I have restarted the build

fisx:
oops, sorry...

garyb:
Looks good, thanks! One last thing, can you squash the commits please?

fisx:
aw, i was being so tidy about the individual commits.  (-:

but it's your code, so here you go.  thanks for talking me through this!

garyb:
Thanks!

Issue #233, opened 2015-11-03
Event sources aren't filterable


jonsterling:
Event sources are producers from `purescript-coroutines`, but these cannot be filtered. A (pure) producer is guaranteed to always have a value; however, most browser-related event streams do not have this property, and so it appears that "waiting on" a value that may never come has been factored into `AVar` in the `Aff` monad. That is, emptiness for producer is implemented using an effect at present, it would appear.

That was a sufficient means to deal with the fact that (for instance) a JavaScript event listener will not always be firing, but I still think the use of `Producer` is kind of unsatisfactory, since it seems to mandate that *all* events from an `EventStream` be handled; in practice, this means adding no-ops to your query algebra for no good reason.

It may be possible to "filter" an event stream by again using `AVar` and `Aff` in the same way as was done above, but this seems kind of anticompositional and difficult, especially in a real world application, you almost *always* want to filter an event stream (or, at least, say "I won't respond to this event").

One option would be to add a new type like `Producer`, but which uses something with a `Stall` constructor instead of just `Emit`. I'm also open to other suggestions; what do people think?

@garyb @jdegoes @beckyconning @cryogenian

cryogenian:
I've had troubles with `Producer` in code that uses `forkAff`, not sure why. Probably it produced more events than `Consumer` can handle. :+1: for switching to something more `Aff`like or remake __consumer/producer__

jonsterling:
I wonder if it would be possible to just switch Halogen to use a stack-safe version of `ListT`. Anybody have any thoughts on this?

jonsterling:
Another option would be to have the event source use `Coproduct (Const Unit) f` instead of `f`. I am not sure what effects this would have on the rest of the apparatus, since the relationships between the various type parameters as enforced in the menagerie of Halogen types is a bit abstruse, or at least unclear to me.

jdegoes:
`ListT (Aff _)` would indeed be much simpler (and it should already be stack safe). We used that in an earlier version of Halogen. I'm not sure that's the right abstraction either, but I think finding the right one sort of depends on unifying how HTML elements produce streams of values with how user-land code (possibly using FFI) produces them, which seems to require vdom rethinking, which in turn is a longer-term project. 

Certainly, `ListT` would provide filtering and lots of other things out of the box; it's a finite or infinite stream, albeit with no mechanism to "unsubscribe" if it's an infinite stream (not that `Producer` is any different).

jonsterling:
[...]

PR #239, opened 2015-11-04
[Request for comment] Try switching event streams to something that can terminate

jonsterling:
@garyb 

This is a little experiment with switching `EventSource` from `Producer` to something that can terminate. The types seem to work; does it look like it does what we intend?

If we think that this is a viable path forward for #233, I can switch it to `ListT`, which should be equivalent, if people want. (Though the `FreeT` version seems so much cleaner & more composable.)

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -34,6 +34,7 @@
     "purescript-maps": "^0.5.0",
     "purescript-nullable": "^0.2.0",
     "purescript-profunctor": "^0.3.1",
+    "purescript-stalling-coroutines": "^0.1.0",
     "purescript-unsafe-coerce": "^0.1.0",
     "purescript-void": "^0.3.0"
   }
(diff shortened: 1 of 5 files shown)


jonsterling:
If we proceed in this direction, we can use the kit in this repository: https://github.com/jonsterling/purescript-stalling-coroutines/

garyb:
I've just tried to review this actually, but I'm too tired to make sense of it right now :wink: I'll revisit tomorrow.

jonsterling:
@garyb No worries :smile: Talk to you tomorrow!

garyb:
I suspect I'm missing the point, but how does this differ from the usual process in "haltability"? You can terminate a `Producer` [by returning a `Right` value](https://github.com/paf31/purescript-coroutines/blob/05ee8ec3b598fe775e4a74eb470da3b475470482/src/Control/Coroutine.purs#L83-L88) - is that not the same thing?

jonsterling:
A producer cannot be filtered.

garyb:
Ah okay, I see now - it makes more sense after looking at your other repository too. Seems like a sensible approach to me :+1:

PR #240, opened 2015-11-05
Switch EventSource to use purescript-stalling-coroutines

### LINKED ISSUE #233: Event sources aren't filterable

jonsterling:
This will allow users to filter event sources (see [`Control.Coroutine.Stalling.filter`](https://github.com/slamdata/purescript-stalling-coroutines/blob/master/docs/Control/Coroutine/Stalling.md#filter)). This is meant to resolve #233.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,3 +14,4 @@ script:
   - npm run example-intro
   - npm run example-multi-component
   - npm run example-todo
+  - npm run example-keyboard-shortcuts
--- a/bower.json
+++ b/bower.json
@@ -34,6 +34,7 @@
     "purescript-maps": "^0.5.0",
     "purescript-nullable": "^0.2.0",
     "purescript-profunctor": "^0.3.1",
+    "purescript-stalling-coroutines": "^0.1.0",
     "purescript-unsafe-coerce": "^0.1.0",
     "purescript-void": "^0.3.0"
   }
--- a/docs/Halogen/Component.md
+++ b/docs/Halogen/Component.md
@@ -131,7 +131,7 @@ descendant components have processed.
 #### `InstalledState`
 
 ``` purescript
-type InstalledState s s' f f' g p = { parent :: s, children :: Map p (Tuple (Component s' f' g) s'), memo :: Map p (HTML Void (Coproduct f (ChildF p f') Unit)) }
+newtype InstalledState s s' f f' g p
 ```
 
 The type used by component containers for their state where `s` is the
--- a/docs/Halogen/Query/SubscribeF.md
+++ b/docs/Halogen/Query/SubscribeF.md
@@ -6,7 +6,7 @@ listeners.
 #### `EventSource`
 
 ``` purescript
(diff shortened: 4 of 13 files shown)


garyb:
Perhaps we could include an example too?

jonsterling:
@garyb sure! Maybe I'll write up a quick keyboard shortcut example.

jonsterling:
@garyb @beckyconning When you get a chance, can you let me know what you think about this? I've added an example w/ keyboard shortcuts.

garyb:
Does `purescript-dom` not provide a way of doing the keyboard event stuff yet?

Either way, looks good to me!

jonsterling:
@garyb Oh neat! I didn't realize `purescript-dom` was as powerful as it is; I think it should be possible to reconstruct that example code using it.

garyb:
It might be missing a few things that would make it possible, but it's supposed to cover the majority of the DOM and HTML APIs, so hopefully it's doable! If not we can perhaps add whatever is missing.

jonsterling:
I'll switch it over to use `dom`, and if there's anything missing, I'll file a ticket and add it.

jonsterling:
@garyb Hmm, I just looked into it and actually, it will probably be more work to use `purescript-dom` than it's worth. Apparently, `keyCode` is not standard and `purescript-dom`'s README says that only standard properties will be supported. Some use of the FFI is, then, inevitable I think, so it might be worthwhile to just keep the example super simple.

What do you think?

garyb:
I remember now, yeah, seems the keyboard stuff is in a mess at the moment. Let's go with this for now then, when browsers are sorted out in 10 years time or something we can use the spec'd version then :wink:

jonsterling:
Cool, sounds good!

garyb:
I'll make a release when #238 is in, unless we need it before then. If so, please go ahead and make a new one yourself - I guess the menu/keyboard shortcut stuff requires this?

PR #244, opened 2015-11-14
Add subscribe' for parent components

### LINKED ISSUE #234: Subscriptions & parent components
I'm opening this ticket so that various discussions which have so far been in Slack or hangouts can be centralized here. Here are a few observations:

1. In the evaluator for a regular component (something of type `Component...`), it is already possible to use `subscribe` to translate events into queries on the component. Just so there is no confusion, this means: if component `C`'s query language contains a query `Q`, then `C`'s evaluator can interpret an event source into invocations of `Q` on itself.

2. In the evaluator for a *parent* component (something of type `ParentComponent...`), this does *not* appear to be possible. The best I was able to do was turn events into queries on a child component.

The above is pretty concerning, if true. If this is really not possible under the current design, then it smells like we have coded ourselves into a corner—ideally, the same code would work in both cases (!), but I'll grant that this is not necessary possible. I think that turning events into queries is a pretty standard use-case that we need to support, on both child and parent components.

If anyone has worked around this issue, please note it here! Or if anyone has ideas on how to change Halogen's architecture to support this, let's talk about it here as well.

### LINKED ISSUE #242: SubscribeF and parent components
@jonsterling 

Trying to free up `subscribe` / `EventSource` so we can use it in parent components (see #234) is proving troublesome.

Currently `EventSource` can only be used in a useful way when `g ~ Aff` partially due to `produce` from `Control.Coroutine.Aff`, but really there's not much else that makes sense if we build the coroutine here. This doesn't work for parent components as they force `g ~ QueryF`.

I thought one fix would be just to capture the arguments we construct the `EventSource` with and defer the coroutine construction until we're interpreting it, as then we don't force `Aff` everywhere - it happens at the end, which is fine, as it has to be `Aff` there anyway for `runUI`. The problem with that is the row of effects need to go somewhere:

``` purescript
data EventSourceF f eff a
  = EventSourceF ((a -> Eff eff Unit) -> Eff eff Unit) (a -> Eff eff (f Unit))

-- Also need to hide the `a`, but there's no problem here
newtype EventSourceE f eff = EventSourceE (Exists (EventSourceF f eff))
```

We don't want to add an additional type variable to `Component` to capture these effects as it'll make the types even worse than they already are, and there will be duplication as the row will usually be present in a component's `g` as well.

So I figured I'd use existential hiding (and `unsafeInterleaveEff` later... :cry:), but I _think_ this is safe, because we can ensure that the existential version it has to be constructed like this: 

``` purescript
subscribe :: forall s f g eff. (MonadEff eff g) => EventSourceE f eff -> Free (HalogenF s f g) Unit
subscribe = ...
  -- there's a mkExists thing in here that disappears the `eff`, but because of
  -- the constraint we know that `g` must be able to deal with them
```

However, I'm not sure this is safe: we can re-interpret a component's `g`, so there's no guarantee that future `g`s will represent the effects we hid here. But then maybe it's not a problem as `Eff`'s rows are additive (excluding ST and other FFI hacks to remove effects)?

garyb:
@jonsterling !!! 

You already discovered the magic invocation we needed to make all this stuff work without changing anything about `SubscribeF` :heart_eyes: 

We just need to `hoistFreeT liftH` an `EventSource` to make it compatible, you did this in the `FileInput` and it suddenly clicked that that's all we needed here too.

Resolves #242, resolves #234

### DIFF
--- a/docs/Halogen/Query.md
+++ b/docs/Halogen/Query.md
@@ -166,7 +166,16 @@ subscribe :: forall s f g. EventSource f g -> Free (HalogenF s f g) Unit
 ```
 
 Provides a way of having a component subscribe to an `EventSource` from
-within an `Eval` or `Peek` function.
+within an `Eval` function.
+
+#### `subscribe'`
+
+``` purescript
+subscribe' :: forall s s' f f' g. EventSource f g -> Free (HalogenF s f (Free (HalogenF s' f' g))) Unit
+```
+
+Provides a way of having a parent component subscribe to an `EventSource`
+from within an `Eval` or `Peek` function.
 
 #### `liftH`
 
--- a/docs/Halogen/Query/SubscribeF.md
+++ b/docs/Halogen/Query/SubscribeF.md
@@ -51,6 +51,15 @@ let onChange = eventSource_ (Session.onChange session) do
 ```
 (Taken from the Ace component example)
 
+#### `catEventSource`
+
+``` purescript
+catEventSource :: forall f g. (MonadRec g) => EventSource (Coproduct (Const Unit) f) g -> EventSource f g
+```
+
+Take an `EventSource` with events in `1 + f` to one with events in `f`.
+This is useful for simultaneously filtering and handling events.
+
 #### `SubscribeF`
 
 ``` purescript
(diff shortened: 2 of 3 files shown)


jonsterling:
holy crap!

PR #246, opened 2015-11-17
Subscriptions

garyb:
@jonsterling Here we go again...

- This time we keep the coroutine inside `EventSource`.
- `EventSource` can only be constructed or deconstructed with the `MonadAff` constraint so that we know `g` can somehow deal with the inner `Aff`.
- We allow hoisting of `EventSource` with a natural transformation. We don't need the `MonadAff` constraint as it's implied by the natural transformation. This allows us to change the `g` to `QueryF s s' f f' g p` when subscribing in a parent. Relying on natural transformation also means we don't need to impose a global `MonadAff` constraint throughout Halogen.
- We allow "contra-hoisting" of `EventSource` so we can return a hoisted `EventSource` to an earlier `g`. This allows us to get our `g` back out of `QueryF`.
- `g` is a phantom type for `EventSource`, so hoist/contrahoist actually do nothing at the value level. 

I'm pretty sure with the above rules this is all safe.

It's still a rather hacky solution really. This whole thing is to allow our `EventSource` to sail through from definition to `runUI` without having to hoist the coroutine directly, we keep the internal `Aff` so we can `forkAff` in `runUI`. The reason my first `subscribe'` blocked is this: https://github.com/slamdata/purescript-halogen/blob/75ee1641c34f555b34c4cc3165de8ee295772f67/src/Halogen/Component.purs#L330... we run the coroutine after hoisting it into `QueryF` rather than when it is `Aff`, so have no opportunity to `forkAff`. I first tried reworking this function to avoid the need for running the coroutine, but I can't see a way of doing it.


garyb:
Oh, to clarify one thing:

> This allows us to get our `g` back out of `QueryF`

It doesn't matter if this is actually impossible or would be partial, because this is just type juggling, the inner value is untouched due to the phantom type thing. We're mainly concerned about the ability to at some point to have a `g` that satisfies the `MonadAff` constraint directly again so we can `runEventSource`.

src/Halogen/Query/EventSource.purs
```diff
@@ -0,0 +1,70 @@
+module Halogen.Query.EventSource
+  ( EventSourceE()
+  , EventSource()
+  , makeEventSource
+  , runEventSource
+  , interpretEventSource
+  , hoistEventSource
+  , contrahoistEventSource
```
jonsterling: we need to export `catEventSource` here

src/Data/ExistsRE.purs
```diff
@@ -0,0 +1,20 @@
+module Data.ExistsRE where
+
+import Unsafe.Coerce (unsafeCoerce)
+
+-- | A variant of `Exists` that works for type constructors that accept a _row_
+-- | of effects.
+data ExistsRE (f :: # ! -> *)
```
jonsterling: I don't know if it makes any semantic difference in PureScript, but is there a way to export this abstractly so that it is not possible to pattern match on it? Otherwise, this looks like an empty type, whence we can prove `False` quite easily.

This is one of the reasons, by the way, I was so strongly in favor of changing PureScript to not export the identities of type synonyms unless you ask it to—always exporting the identity of a type synonym makes data abstraction very difficult, and in many cases impossible.

garyb: It's already not possible to pattern match on this - it has no constructor. Essentially this is the same as foreign data import.

Agreed :)

src/Halogen/Query/EventSource.purs
garyb: Oops, good spot.

src/Data/ExistsRE.purs
jonsterling: What's the difference between this declaration and an empty type (which also has no constructor), then?

garyb: I guess there is none really.  How do you prove an empty type false?

jonsterling: There are two ways, depending on your perspective. The first perspective is that every empty type literally *is* False, so if you can construct an inhabitant of it, you have a proof of falsehood.

Secondly, sum types are *always* meant to support pattern matching, even the special case of the nullary sum. So, even if your language doesn't have syntax for pattern matching on a nullary sum, the below is *always* a valid thing to do, assuming that you know that `Empty` is a sum type:

```
absurd :: forall a. Empty -> a
absurd x = unsafeCoerce "the pattern `x` is impossible!"
```

Now, on the contrary in (a hypothetical version of) ML (where you can declare a nullary sum) the following two signatures are quite different: 

```
signature SIG1 = 
sig
  type welp
end

sig SIG2 =
sig
  datatype welp = .
end
```

In `SIG1`, `welp` is a type about which you know nothing: you can neither construct it nor destruct it. On the other hand, in `SIG2`, `welp` is the empty type and is isomorphic to `false`. So, you know that it can have no values ever.

src/Halogen/Query/EventSource.purs
```diff
@@ -0,0 +1,71 @@
...
+newtype EventSourceE f eff = EventSourceE (SCR.StallingProducer (f Unit) (Aff eff) Unit)
+newtype EventSource f (g :: * -> *) = EventSource (ExistsRE (EventSourceE f))
+
+makeEventSource :: forall f g eff. (MonadAff eff g) => SCR.StallingProducer (f Unit) (Aff eff) Unit -> EventSource f g
+makeEventSource = EventSource <<< mkExistsRE <<< EventSourceE
+
+runEventSource :: forall f g eff. (MonadAff eff g) => EventSource f g -> SCR.StallingProducer (f Unit) (Aff eff) Unit
+runEventSource (EventSource xese) = runExistsRE (\(EventSourceE p) -> hoistFreeT coerceEffects p) xese
+  where
+  -- This coercion is safe because our `MonadAff` constraint implies `fff ~ eff`
```
jonsterling: Can you elaborate in this comment, on why `MonadAff eff g` implies `fff ~ eff`?

garyb: :boom::gun: it isn't actually:

Say we `makeEventSource` with `MonadAff (x, y, z) g` I guess a `MonadAff () g` still works for `runEventSource`.

jonsterling: aggh! this stuff is so hard :frowning:

garyb: I think if `makeEventSource` and `runEventSource` were specialised to `Aff` directly that would recover what I was trying to do though? As the row would be contained in the type then: `EventSource f (Aff eff)`. I'm not sure why I went to `MonadAff` actually.

jonsterling: I think that'd at least be an improvement; let's make the switch, and then I'll look at it again and see if this makes sense...

src/Data/ExistsRE.purs
garyb: Interesting. How is it different if we define a type with constructor but then hide the constructor export then? 

I'm actually not entirely sure why it's defined as this empty type, by the way. I think something like this would work too:
```
newtype ExistsRE (f :: # ! -> *) = ExistsRE (forall eff. F f eff)
```
I just copied `Exists`/`ExistsR` and tweaked the kind.

jonsterling: @garyb I don't think that hiding the constructor export would help in PureScript. In ML, the reason `SIG1.welp` isn't necessarily the empty type is that the identity of the type is literally not exposed, and so you really have no idea how it is implemented (i.e. is it a datatype? or a type synonym?). 

Haskell may have the same problem, fwiw, as PureScript here, but I'm not certain. I think the proper thing to do in the language is to introduce abstraction.

In the present case, if your `newtype` idea works, then let's do that! :smile: If it's necessary on the other hand to *implement* this using an empty type + coercions, it would better express the intent if we made `newtype Welp = Welp Empty`, and then hid the constructor; I'm not certain if this ends up meaning anything different in PureScript, but I think it would at least be more clear that it is not intended to support pattern matching.

garyb: I think Haskell does have the same problem, unless there's some type family kind of thing that can better represent it.

The `newtype` is definitely better, as that's how we're treating it, so I'll give it a go! It wouldn't matter whether we hid the constructor or not then too, as you wouldn't be able to do anything with it, it'd basically be the same as `exists eff. f eff` anyway.

garyb:
@jonsterling I've made those changes. The `newtype` exists thing is much nicer, as we don't need to cheat to "run" it now either.

src/Data/ExistsRE.purs
```diff
@@ -0,0 +1,20 @@
+module Data.ExistsRE where
+
+import Unsafe.Coerce (unsafeCoerce)
+
+-- | A variant of `Exists` that works for type constructors that accept a _row_
+-- | of effects.
+newtype ExistsRE (f :: # ! -> *) = ExistsRE (forall eff. f eff)
```
jonsterling: Is this actually correct? I believe the quantification is now in the wrong place. This is saying that the content of the existential is a global section / limit of `f`, but what you really want is a colimit. I'm not sure if the following is possible in PureScript, but what you need is something like:

```
data ExistsRE (f :: # ! -> *) = forall eff. ExistsRE (f eff)
```

Alternatively, if this is the only way to implement the existential in PureScript, you'll need to hide the constructor of `ExistsRE`.

garyb: Argh, you're right. I thought there was something not quite right.

garyb: Yeah, that's the best we can do I think. Another option would be to make it foreign explicitly, do you think that would better express what's going on? (In an "all bets are off" kind of way)

garyb: By which I mean:
``` purescript
foreign import data ExistsRE :: (# ! -> *) -> *
```

jonsterling: Yeah, I doing the foreign import makes sense in this cae.

src/Halogen/Query/EventSource.purs
```diff
@@ -0,0 +1,72 @@
...
+
+makeEventSource :: forall f eff. SCR.StallingProducer (f Unit) (Aff eff) Unit -> EventSource f (Aff eff)
+makeEventSource = EventSource <<< mkExistsRE <<< EventSourceE
+
+runEventSource :: forall f eff. EventSource f (Aff eff) -> SCR.StallingProducer (f Unit) (Aff eff) Unit
+runEventSource (EventSource xese) = runExistsRE (\(EventSourceE p) -> hoistFreeT coerceEffects p) xese
+  where
+  -- This coercion is safe because we can only construct an `EventSource`
+  -- with an `Aff`, and we can only transform `g` with natural tranformations,
+  -- so for any `Aff` we end up with the effect row must subsume the original.
```
jonsterling: OK; let's say this explicitly in the comment, in case it turns out not to obtain and we are left scratching our heads in the future :wink: 

However, I think this principle is false. Here's a possible counterexample. Suppose we have got an effect called `divergence :: DIVERGENCE` with a `diverge` operator. Now can construct a natural transformation `η : Aff (err :: EXCEPTION, divergence :: DIVERGENCE) => Aff (divergence :: DIVERGENCE)` by handling all exceptions with `diverge`.

garyb: Hmm, you're right, we kinda went over this before: we almost never remove effects, but that the possibility exists is problematic. Any ideas then? Or is this approach sunk as well?

jonsterling: If there is a possibility that the correctness of this approach only depends on a weaker principle than the one I laid out, maybe it can be salvaged...

But I am feeling a little bit pessimistic about it.

jdegoes: Not to be heretical or anything, but: :smile: 

```purescript
foreign import data IO :: * -> *

liftIO :: forall eff a. Aff eff a -> IO a
liftIO = unsafeCoerce

runIO :: forall a. (forall eff. Aff eff a -> z) -> IO a -> z
runIO f io = f (unsafeCoerce io)
```

jonsterling: @jdegoes Hahaha, this may actually be the simplest approach...

garyb:
@jonsterling I think... maybe... this time... etc. :wink:

`EventSource` is back to being just a synonym for a coroutine. Now there's a `ParentEventSource` wrapper that we use temporarily, that is interpreted into a normal `EventSource` during component installation without having to run the coroutine.

The `ParentEventSource` wrapper allow us to fudge the types a bit, so `ParentEventSource f (Free (h g)))` unwraps to `EventSource f g` (`h ~ HalogenF s f` for us, I left it generic to avoid a circular dependency between `HalogenF`, `SubscribeF`, `ParentEventSource`). The constructor is hidden so we can only create or unwrap a `ParentEventSource f (Free (h g)))` and there are no operations we can run to manipulate any of the type parameters after it has been constructed, preserving the isomorphism.

The main downside of this approach is now we have an extra case we have to deal with for `SubscribeF` in `runUI` and `transform` where the subscription is for a `ParentEventSource`: this will never arise from Halogen and cannot be sensibly interpreted, so we fall back to halt.

jonsterling:
OK, this looks more promising. I'll look at it in more depth shortly...

src/Halogen/Query/EventSource.purs
```diff
@@ -0,0 +1,91 @@
...
+catEventSource
+  :: forall f g
+   . (MonadRec g)
+  => EventSource (Coproduct (Const Unit) f) g
+  -> EventSource f g
+catEventSource es =
+  SCR.catMaybes $
+    SCR.mapStallingProducer (coproduct (const Nothing) Just) es
+
+foreign import data ParentEventSource :: (* -> *) -> (* -> *) -> *
```
jonsterling: Is there any reason for this not to be a `newtype` of `EventSource`?

garyb: I wasn't sure what was best to use here. `newtype` definitely works with a hidden constructor, but we still need to construct/deconstruct it with `unsafeCoerce` (I think), so I figured I'd make the whole thing foreign.

jonsterling: Oh, I had read the injection backwards; what you say makes sense, I think.

garyb:
Whipped up a little something on the train that means the halt stuff is gone too - added a parameter to `HalogenF` so we can provably know we're dealing with an `EventSource` rather than a `ParentEventSource`

jonsterling:
@garyb That's a cute fix! This is looking pretty solid, I think. I need to think for about 10-20 minutes about one aspect, but I'm tentatively hopeful about this approach.

src/Halogen/Query/EventSource.purs
```diff
@@ -0,0 +1,98 @@
...
+    SCR.catMaybes $
+      SCR.mapStallingProducer (coproduct (const Nothing) Just) es
+
+foreign import data ParentEventSource :: (* -> *) -> (* -> *) -> *
+
+toParentEventSource :: forall f g h. EventSource f h -> ParentEventSource f (Free (g h))
+toParentEventSource = U.unsafeCoerce
+
+fromParentEventSource :: forall f g h. ParentEventSource f (Free (g h)) -> EventSource f h
+fromParentEventSource = U.unsafeCoerce
```
jonsterling: So, my only concern in this PR now whether this `ParentEventSource` type is kosher. The safety obligation incurred by these uses of `unsafeCoerce` is essentially the following:

> There exists a solution to the isomorphism `ParentEventSource f (Free (g h)) ≃ EventSource f h` for all `g`.

Do we have any reason to believe that the above is true? Can you "solve for" `ParentEventSource` (informally)?

jonsterling: I have thought about it some more, and I think I was worrying for no reason. We can do the very same thing in Haskell using a GADT:

```haskell
data ParentEventSource f g where
  PES :: EventSource f h -> ParentEventSource f (Free (g h))

toParentEventSource :: EventSource f h -> ParentEventSource f (Free (g h))
toParentEventSource = PES

fromParentEventSource :: ParentEventSource f (Free (g h)) -> EventSource f h
fromParentEventSource (PES e) = e
```

So, I consider this matter resolved.

jonsterling:
OK; I *think* that this PR is good, modulo the build errors in the examples. Nice work, @garyb!!

src/Halogen/Query/EventSource.purs
garyb: Phew! That was the encoding I had in mind, glad you concur that it's safe. Do you think there's a better way to express it in PureScript?

jonsterling: I think it's probably fine as-is

garyb:
Tests are passing again now. I'll squash and merge if you're happy @jonsterling?

jonsterling:
@garyb Awesome! :+1:  from me

garyb:
Huzzah, built at last!

PR #247, opened 2015-11-18
Add autofocus property

spicydonuts:
Autofocus was already listed as a valid property on the `input` element function, but it was not yet defined in Halogen.HTML.Properties or Halogen.HTML.Properties.Indexed.

This PR adds the `autofocus` property definitions and adds the new property to the example `example-todo`.

Note: I'm not sure how valid the `autofocus` property is on a completely client-side rendered app, since the browser only seems to read it on initial page load.  It's still part of the spec, though, and there could be a future case like server-rendered html.


jonsterling:
:+1:

PR #248, opened 2015-11-18
Fix behaviour of length literals

### LINKED ISSUE #243: Setting `width` and `height` properties results in 0 width/height in rendered component
After setting the `width` and `height` properties of a `img` to a non-zero, the rendered component has the width and height set to zero. 

An example of this can be found [here](https://github.com/enetsee/halogen-gravatar/blob/master/src/Halogen/Gravatar/Component.purs#L27-L28).

garyb mentioned that he has also encountered this issue.

garyb:
Resolves #243 

Also switched pixels lengths to only accept `Int`, as I believe that is all that is valid for a non-CSS length (tested too - a float value is rounded to the nearest int when set via the property).

The problem here was down to `IsProp` being misleading, it's not used to render values for properties, but instead for attributes for string renderering. The values passed to `prop` are not processed, and are passed straight through to be set as properties on the `virtual-dom` nodes.

### DIFF
--- a/docs/Halogen/HTML/Properties.md
+++ b/docs/Halogen/HTML/Properties.md
@@ -6,13 +6,8 @@ This module provides `Prop` values for some common HTML attributes.
 
 ``` purescript
 data LengthLiteral
-  = Pixels Number
+  = Pixels Int
   | Percent Number
-```
-
-##### Instances
-``` purescript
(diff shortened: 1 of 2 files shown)


jonsterling:
This change looks good to me.

PR #249, opened 2015-11-18
Add missing onMouseEnter to indexed events

alidd:
It looks like purescript needs some metaprogramming thing for boilerplate like xxx.indexed


jonsterling:
:+1:

PR #254, opened 2015-12-13
Add `rows` and `cols` to (Indexed) Properties.

montanonic:
`Halogen.HTML.Element.Indexed.textarea` currently accepts the `rows` and `cols` properties, yet they are undefined in both standard and indexed forms. This branch adds them.

I tested these additions with Ajax.Example's `H.textarea`, and my inputs of `P.rows 10` and `P.cols 50` properly rendered as attributes in the `textarea` tag on the browser.


garyb:
Looks good, thanks! I'll wait for Travis to finish anyway, and then make a new release.

PR #256, opened 2015-12-15
Table of Contents

aspidites:
I kept losing my spot when browsing different parts of the readme. Generated a TOC with http://doctoc.herokuapp.com/


aspidites:
Yuck. Didn't do what I thought it would.

PR #262, opened 2015-12-27
Remove *most* warnings generated by purescript 0.8.0-rc.1

telser:
- Warnings about imports made redundant by whole module imports were left alone since it seemed that it often had to do with re-exports.
- Everywhere there was a whole module import I silenced the warning by using (..) rather than use the suggested explicit imports to keep semantics the same (this is a non-trivial codebase and I'm not going to pretend to have grokked all of the inner workings). 
- Fixed all of the missing class keyword warnings

Thanks for the awesome project everyone. This was motivated by using the rc and halogen at work. Saw all the warnings and thought I'd contrib something. Let me know if I should change anything!

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,8 @@
   "contributors": [
     "Phil Freeman <paf31@cantab.net>",
     "John A. De Goes <john@degoes.net>",
-    "Gary Burgess <gary.burgess@gmail.com>"
+    "Gary Burgess <gary.burgess@gmail.com>",
+    "Trevis Elser <trevis.elser@gmail.com>"
   ],
   "scripts": {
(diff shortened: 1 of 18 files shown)


telser:
Oh and obviously this can't be merged until 0.8.0 is out (unless someone changes the devDeps to be 0.8.0-rc.1)

garyb:
I ended up doing this in a newer PR, so closing, but thanks again for doing all these!

telser:
Absolutely! Sorry I let so many of these get stale. If there is anything else I can do let me know.

garyb:
No problem, it's more my fault for not merging them at an earlier point. :smile:

PR #265, opened 2016-01-05
Added aria properties

beckyconning:
This pr is pretty naive. There may be things things missing and I'm pretty sure that the indexed version could be more restrictive.


garyb:
I don't think any of these are properties, but are instead attributes. A "property" is something that exists in the prototype of HTML elements, and will often have an alternative javascript name to avoid name clashes or illegal identifiers. For example, the `class` attribute has a corresponding `className` property, `for` is `htmlFor`, etc.

As far as I can tell, there is no "built in" ARIA support at the javascript level, as none of these have associated property names if you start poking around the DOM with the web inspector.

Looks like we'd want to add the `role` attribute too? (Which again is an attribute, not a property).

As for the indexing, we'll definitely want to consider doing something else there. As it stands all those properties will be unusable :smile:. I'd suggest just using a type variable for the row, which will allow the aria attributes to be used on any element. This is perhaps not totally ideal, as we're basically throwing away what indexing gives us, but in this case the attributes are associated with particular _roles_ rather than elements, and we definitely don't have a way of modelling that with the current indexing.

I'd also suggest we define these in a `Halogen.HTML.Properties.ARIA` module perhaps, and not drop them in with the more common properties?

beckyconning:
That makes sense. I agree they aren't properties. 

I spent a little while reading `Halogen.HTML.Properties.Indexed` and couldn't work out exactly what I needed to do to make these restricted by their role so your suggestion of making them work for every element sounds sensible.

The two main problems I had were

* I haven't been able to work out how to make these specifically attributes. The`propName`s aren't wrapped in a `Maybe` like the `attrName`s are.

* It didn't seem quite right to add these to the `GlobalAttributes` synonym.

How would I go about "using a type variable for the row, which will allow the aria attributes to be used on any element"?

garyb:
Ah sorry, I should have explained more!

- To make them attributes we just use the [`Attr` constructor](https://github.com/slamdata/purescript-halogen/blob/b8353d53a536b9b9cebda6984956459a8a8afb17/src/Halogen/HTML/Core.purs#L95) rather than `prop`.
- The type variable row thing is just using a type like `ariaDescribedby :: forall r i. String -> IProp r i` - since `r` is left unspecified it will unify with any row, making it compatible with any element.

beckyconning:
Thanks thats super clear now : )

beckyconning:
@garyb not sure why travis took half an hour but this is ready for a second look when you are available : ).

garyb:
It just takes that long now unfortunately. Code good to me, can you regenerate the docs and I'll do one final check through?

beckyconning:
Docs regenerated.

garyb:
Sorry a few more things:

Can you add the new `ARIA` modules in here https://github.com/slamdata/purescript-halogen/blob/b8353d53a536b9b9cebda6984956459a8a8afb17/gulpfile.js#L46-L72 and then do the docs

For the new props, do you think maybe we should camelcase them? I'd probably expect things to be like `ariaValueMin` rather than `ariaValuemin`, `ariaLabelledBy` instead of `ariaLabelledby`, etc. 

What do you think about dropping the `aria` prefix too? As they're in their own modules they can be imported under a separate namespace to avoid confusion/collision with other properties/attrs.

beckyconning:
If you're happy with matching value names across these modules e.g. `selected` such that they should be used like

```
import Halogen.HTML.Properties as P
import Halogen.HTML.Properties.ARIA as PA
```

instead of 

```
import Halogen.HTML.Properties as P
import Halogen.HTML.Properties.ARIA as P
```

Then I'm happy to remove the aria prefix.

I totes prefer the camelcased one. I was trying to stay consistent with things like [`readonly`](https://github.com/slamdata/purescript-halogen/blob/master/src/Halogen/HTML/Properties.purs#L26 ) but perhaps "readonly" is a single word? The web lol!

garyb:
Yeah, importing as `PA` or `ARIA` was what I had in mind - also because importing multiple modules `as` the same thing will be a warning in 0.8 unless they're imported with an explicit list (`import X (y, z) as A`) so as to avoid problems with name collisions if modules are extended at a later date.

And that is an excellent point, just use your discretion as to where the humps should go. :laughing:

beckyconning:
I renamed the values. I chose to camelCase all of them to keep the module consistent.
I also updated the gulpfile and docs.

garyb:
All looks great, thanks!

PR #266, opened 2016-01-09
Add FunctorEff and FunctorAff classes

garyb:
@jdegoes I figured we could include these here for now, since it's the motivating case. Maybe they can be adopted elsewhere if they turn out to be generally useful - but this approach seems _great_ so far. I've managed to do away with all the `liftH $ liftAff'` type stuff I've tried it on so far in `slamdata`.

It's also a drop in replacement for any instance of the `Monad` version of `liftEff`/`liftAff` that I've found so far. Can you think of a case where the `Monad` constraint would actually be useful?

### DIFF
--- a/docs/Halogen/Query/HalogenF.md
+++ b/docs/Halogen/Query/HalogenF.md
@@ -15,6 +15,8 @@ The Halogen component algebra
 ##### Instances
 ``` purescript
 (Functor g) => Functor (HalogenFP e s f g)
+(FunctorEff eff g) => FunctorEff eff (HalogenFP e s f g)
+(FunctorAff eff g) => FunctorAff eff (HalogenFP e s f g)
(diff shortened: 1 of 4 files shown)


jdegoes:
I think this is a good idea (lessening constraints). One could even go further and drop the functor requirement, I suppose, in which case if you wanted a monad *and* `liftEff`, then you'd do something like `(Monad m, EffF f) => ...`. Sort of ventures into lawless territory, but then again, `liftEff` never had any laws to begin with.

garyb:
That's true. And really this is just a specialized form of `Inject` too, without projection, and no overlapping instances - if we manage to implement instance chains (or something like that) we could potentially safely go back to that approach.

I'll merge as it is for now though, as it's definitely useful in the current form.

paf31:
I thought the law was that liftEff is a monad morphism.

PR #268, opened 2016-01-15
Use MonadAff for eventSource constructors

garyb:


cryogenian:
Can we use `FunctorAff` from `slamdata` here? If so we could make foreign components with `Free`, right?

jdegoes:
@cryogenian I suggest we put the following into `purescript-aff`:

```purescript
class Affable e f where
  liftAff :: forall a. Aff e a -> f a

class (Functor f) <= FunctorAff e f

class (Apply f) <= ApplyAff e f

class (Applicative f) <= ApplicativeAff e f

class (Monad f) <= MonadAff e f
```

If it works out well, we could try to get it replicated for `eff`.

garyb:
@cryogenian that's definitely the goal eventually, but I didn't want to make it a breaking change by switching to a new version of `Aff` that has the `Affable` hierarchy outlined above (not that it exists yet, but I'd been talking to John about it).

garyb:
Oh, sorry, that doesn't explain why it's not `FunctorAff` though: that's because it's required to be a `MonadAff` by `produce'` which comes from `purescript-aff-coroutines`, and relaxing the rules there without `Affable` is very difficult, due to a `MonadRec` constraint that comes in via `FreeT`.

Issue #270, opened 2016-01-17
Perform query on every child


zudov:
In some cases it seem to be useful to perform a query on every child, getting a list of responses as a result. At the moment one can do that by keeping a list of all used placeholders and mapping over them, but it should be possible to provide an API which:

1. would do that more efficiently,
2. wouldn't require to somehow maintain a list of used placeholders.

(2) alone could probably be handled by allowing to query a list of placeholders (`InstalledState.children >>> M.keys` (that should be safe, right?)

What do you think about that?

zudov:
Solution was implemented [in](url) the PR referenced above.

PR #271, opened 2016-01-17
Queries on all children

zudov:
I decided to try implementing something for #270

- added `queries` and `mkQueries`
- extended 'todo' example with "Mark all as completed" button (not the best showcase though as task list component has all the `TaskId`s in its state so it could as well just traverse over them)

### DIFF
--- a/src/Halogen/Component.purs
+++ b/src/Halogen/Component.purs
@@ -19,9 +19,14 @@ module Halogen.Component
   , QueryF()
   , mkQuery
   , mkQuery'
+  , mkQueries
+  , mkQueries'
   , liftQuery
   , query
   , query'
+  , queryAll
+  , queryAll'
+  , childSlots
   , transform
   , transformChild
   , interpret
@@ -33,7 +38,6 @@ import Prelude
 
 import Control.Apply ((<*))
 import Control.Bind ((=<<))
-import Control.Monad.Eff.Class (MonadEff)
 import Control.Monad.Free (Free(), foldFree, liftF, mapF)
 import Control.Monad.Free.Trans as FT
 import Control.Monad.State (State(), runState)
@@ -42,16 +46,16 @@ import Control.Monad.State.Trans as CMS
 
 import Data.Bifunctor (bimap, lmap, rmap)
 import Data.Functor.Coproduct (Coproduct(), coproduct, left, right)
+import Data.List as L
 import Data.Map as M
 import Data.Maybe (Maybe(..), maybe)
 import Data.Maybe.Unsafe as U
 import Data.NaturalTransformation (Natural())
-import Data.Profunctor.Choice (Choice)
-import Data.Traversable (for)
+import Data.Traversable (for, traverse)
 import Data.Tuple (Tuple(..), snd)
 import Data.Void (Void())
 
-import Halogen.Component.ChildPath (ChildPath(), injState, injQuery, injSlot, prjState, prjQuery)
+import Halogen.Component.ChildPath (ChildPath(..), injState, prjState, injQuery, prjQuery, injSlot, prjSlot)
 import Halogen.HTML.Core (HTML(..), fillSlot)
 import Halogen.Query (get, modify, liftH)
 import Halogen.Query.EventSource (EventSource(..), ParentEventSource(), runEventSource, fromParentEventSource)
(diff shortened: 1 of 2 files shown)


jdegoes:
I like the idea of this, but should it not return `Map p i` instead? You can always convert that to `List i`, but you cannot go the other way around.

zudov:
@jdegoes `Map p i` makes sense. I didn't think about it.

zudov:
It seems that the name of existing `query` function is a verb ("to query") rather than singular noun ("a query"). Probably the plural version of `query` should be something like `queryAll` rather than `queries`.

Any better ideas for naming?

jdegoes:
`queryAll` seems like a good name.

garyb:
:+1: for `queryAll` and the `Map` return value.

zudov:
I think `mkQuery` and `queryAll` is a pretty self-contained addition and if that's good to go I can squash the commits so that it's ready to be merged.

There is still no way of querying all grandchild components, or querying all children of a top level component from the driver, but I don't think that such functionality is really needed. Also, adding something like that seem to require much more breaking changes to `ChildF` mechanism of pointing to the children.

src/Halogen/Component.purs
```diff
@@ -58,6 +60,8 @@ import Halogen.Query.EventSource (EventSource(..), ParentEventSource(), runEvent
 import Halogen.Query.HalogenF (HalogenF(), HalogenFP(..), hoistHalogenF, transformHF)
 import Halogen.Query.StateF (StateF(..), mapState)
 
+import Unsafe.Coerce (unsafeCoerce)
```
garyb: Wait, what now? :laughing:  (I assume this is leftover from hole driven development)

zudov: Uups

zudov: Fixed. thanks!

zudov:
Hmm, just ended up in a situation where I have multiple child types and I want to query all children of specific type.

Need to add `mkQueries'` and `queryAll'`

garyb:
Hah, yes. I did think about those cases, but I wasn't sure if the `ChildPath` prisms let you get out of it without yet more combinators... I guess it makes sense that that is not the case though, given we needed the `query'` / `mkQuery'` combinators.

zudov:
@garyb It seem to type-check, I'll also check it with my code now.

zudov:
I am getting some weird errors when using `queryAll'`, e.g. `queryAll cpR ...` works fine, but then if I stick `H.slot' cpL ...` in my dom this same `queryAll cpR` begins failing with "Error: Always Fails" (coming from Aff)

zudov:
The latest commit makes a modification of multi-component example which shows the failure, pressing "Read states" buttons causes the "Error: Always fails", I digged it with debugger, but didn't find the solution. Any insights?

zudov:
After further investigation, something like `query (Left FooSlot) (left (action FooAction))` works fine, but `queryAll (left (action FooAction))` fails. So it's not just about the primed functions.

garyb:
- `"Always fails"` is what happens when `empty` is used in `Aff`.
- This happens when the `HaltHF` action from `HalogenF` is evaluated.
- This happens when you send a query with the "wrong value" to a child component.

So to `queryAll` you'll have to filter the children so as to only send queries to the children "on the left", in this case. Have a look at `Halogen.Component.transform` for a bit more on this.

zudov:
@garyb Ha, that makes sense. I think I got to the second point by myself and was suspecting the third one.

I will try to implement it now. Thanks.

garyb:
Yeah, it's a little unfortunate, I tried many approaches to make this fully safe using the type system, but I think we'd need dependant types to do so really.

The next best thing I could come up with was to say "as long as you use the `ChildPath` stuff with multi-typed child components it will always work", but as you've encountered now that's not totally airtight :wink:

zudov:
@garyb Seems like I fixed the primed functions, now the return types are not confusing (`Map p i` instead of `Map p' i`), I should have noticed that just from types.

However, `mkQueries` and `queryAll` remain to be broken as I described previously. Looking at `mkQueries` context, it just seem to lack information needed to do such matching. Perhaps I miss something? As I see now, the only way to make `queryAll` not to fail is to require supplying a `ChildPath` which it would project with, and which would make it identical to `queryAll`.
```haskell
mkQueries
  :: forall s s' f f' p g i.
   . (Functor g, Ord p)
  => f' i
  -> QueryF s s' f f' g p (M.Map p i)
mkQueries ..
 where
  matchesQuery :: f' i -> p -> Boolean
  matchesQuery = ???
```

zudov:
I also need to come up with more meaningful examples.

garyb:
I'm not sure I understand how `mkQueries` and `queryAll` are broken, unless you mean when they're used in a situation where `mQueries'`/`queryAll'` should be used instead? I don't think there's a way to avoid that.

zudov:
@garyb Yes, I meant the `queryAll (left (action FooAction))` thingy

garyb:
Right... yeah, I think that's unfixable. We should just include a note in the documentation about those functions being unsuitable when there are mixed component types as children.

zudov:
I refactored code a little and squashed commits, it now should be ready for more-or-less final review. Commit message provides a summary part of which could go to release notes.

zudov:
Just wanted to ping here. Can we merge it, or we want to add something else? I've been using this patch for some time and it works great so far.

src/Halogen/Component.purs
```diff
@@ -186,6 +190,33 @@ query'
...
+-- | Queries every child component that is currently installed. For use within
+-- | a parent component's `eval` or `peek` function.
+queryAll
+  :: forall s s' f f' g p i
+   . (Functor g, Ord p)
+  => f' i
+  -> Free (HalogenFP ParentEventSource s f (QueryF s s' f f' g p)) (M.Map p i)
+queryAll q = liftQuery (mkQueries q)
+
+childSlots
```
garyb: Could you add a doc comment for this one?

zudov: Yes, just a moment.

zudov: Added "Returns slots of all currently installed child components."

garyb:
Thanks!

zudov:
@garyb @jdegoes Thanks for all help

Issue #272, opened 2016-01-18
Finalizers for components are useless (broken)


garyb:
The way things are currently, I don't think there's a case where a `finalizer` will ever work when used for a component.

The finalizer is triggered when an element is removed from the DOM, but because this happens when a component has been removed from the state, this means it is not possible to route the query to the component that was just removed.

cryogenian:
There is global state too. For example `purescript-halogen-echarts` uses `Ref` to handle state of `echart` instances, and `finalizer` prevents memory leaks.

garyb:
Yeah, I encountered this working on `new-halogen`: whenever you delete an Ace cell an "Always fails" error occurs from `Aff`, due to halogen being unable to route the `Quit` query back to the component so as to disable the autocomplete stuff... so there will be memory leaks all over the place right now :disappointed:

natefaubion:
The reason why https://github.com/slamdata/purescript-halogen/blob/master/examples/lifecycle/src/Main.purs#L40-L41 works is because the lifecycle events are routed to the parent component.

I don't see how you could have proper finalization without decoupling lifecycles from virtual-dom (putting them in the component spec) and making it part of the render logic.

garyb:
Yeah, I'm thinking that'll be the solution. I'd prefer initializers/finalizers to be associated with entire components rather than random HTML element(s) anyway too. Unless there are downsides to that I'm not seeing?

jdegoes:
Perhaps component acquiring something like this would work:

```purescript
initialize :: Maybe (f Unit),
finalize :: Maybe (f Unit)
...
```

i.e. these are the commands to generate when the component is created and destroyed, respectively.

The installation combinators need the ability to run them on creation / destruction based on the placeholders, which I think should be possible.

garyb:
Yeah, I'm already on it :smile:. Although `Maybe (HTMLElement -> f Unit)`, as we often need an element in these situations to perform the initialization of 3rd party components.

natefaubion:
> Unless there are downsides to that I'm not seeing?

I don't see any, as this is how React works as well. Not quite sure how you'll get a hold of the HTMLElement from virtual-dom though.

garyb:
I've spent some time looking into this now, and yeah there are a few tricky points. I don't really have the time to fix it right now, as the change is pretty big, so if anyone else wants to have a stab at it I'd greatly appreciate it!

I'd like to alter `Component`'s type to this:

``` purescript
[...]

Issue #221, opened 2015-10-28
Switching from composable to decomposable


jdegoes:
I think a fairly conservative change can make it easier to build applications under Halogen.

Currently, components are built by composing them from smaller components. This leads to some rather complicated intermediate structures, which are not very easy to work with (e.g. 4-way nested `Coproduct`). However, there's nothing in Halogen that prevents us from going the other way: that is, from decomposing components from larger ones.

First, define a `Nest` (like bird's nest), which represents the location a component sits inside a parent:

``` purescript
type Nest s s' f f' g g' =
 { state :: LensP s s', input :: Prism1P f f', output :: Prism1P g g' }
```

The `Nest` is a description of how to get and set the component's state, how to extract out the component's input from the parent's, and out to map the component's output to the parent (types are _approximate_).

Now, when defining the component, you actually define it as a function from its nest to the component:

``` purescript
aceEditor :: forall eff s f g. (Functor f) => Nest s AceState f AceIn g AceOut -> Component s f g
aceEditor self = ...
```

Then when accessing the sate through helpers, one must supply the nest:

``` purescript
getState self
```

Similarly, when processing an input or producing an output, one must supply the nest.

It would be a convention for every component `X` to define `XState`, `XNest`, etc.

Various lens-related machinery can be used to access children and so forth.

Needs more fleshing out, but what do people think? Decomposition a viable alternative to composition?

/cc @garyb @beckyconning @cryogenian @jonsterling

garyb:
This is very much related to the things Jon and I have been looking at, but the real sticking point is the inability to provide a non-partial function that can safely transform to and from `s` and `s'`.

The `f` and `g` cases are a bit easier as you can require a `Plus` constraint and fall back to `empty`, but I am yet to figure out how to handle the state mapping.

garyb:
Although our use case was perhaps a little different, as we were trying to simplify having large sums of children, whereas this isn't quite that - here we're just "zooming" in on a part of the parent's state for the inner component to use, so it can definitely work for that.

jdegoes:
> whereas this isn't quite that - here we're just "zooming in" on a part of the parent's state for the inner component to use, so it can definitely work for that.

That's correct. 

Parents would then explicitly contain the state of their children, e.g.:

```purescript
data MyState = MyState
  { children :: Map AceId AceState, ... }
```

When you query a child, you "lift" it to the parent, and because the lookup may fail, the helper which performs the query requires a failure state in `g`.

[...]

PR #275, opened 2016-02-04
Initializers/finalizers

natefaubion:
I thought I'd go ahead and put this in a PR. This only adds some internal plumbing, so everything should be backwards compatible.

* I didn't go the route of annotating trees, I was able to get there by adding a function that collects finalizers by providing the component state. The user-facing API just takes an `Action` though.
* Finalizers cannot be peeked, but initializers can. This is because finalizers are run after a component is removed from the main state, so its run in a "detached" state where state updates don't affect the main state. The only way around this would be to have a wrapper query functor for parent components https://github.com/slamdata/purescript-halogen/issues/272#issuecomment-177229862.
* Initializers currently cause rerendering. So when you load an app, if you have 100 initializers running at the same time, it will be rendered 100x. This is consistent with what happens with the current initializers, and I think the main driver would have to be refactored a bit to change _where_ it triggers rerendering. We could also just throttle rendering with `rAF`.
* I added a `ParentQuery` type synonym because I thought `Coproduct f (ChildF p f')` got a little tedious, but I can remove this.
* I haven't made any changes for getting HTMLElement refs. I figured that would be a different PR.


garyb:
This looks great! Sorry I've not been getting back to your comments in the other issue, was a bit busy getting the "new halogen" version of SlamData released :smile:

- Finalizers not being peekable is totally fine, and I'd say expected.
- Initializers causing re-rendering I think is okay, especially if it's no different to the current situation, but perhaps we should open an issue to look at it again in the future. `rAF` seems like a bit of a hacky way to batch them, but it would work!
- :thumbsup: `ParentQuery`

When we next make breaking changes we may as well do away with the distinction between `Component` and `ComponentSpec`, and just use the spec versions... assuming that doesn't cause type system issues (I'm thinking of things like where we need to use `ALens` in records).

Do you have something in mind for how to get the `HTMLElement`s? That's probably the most important part of why we have initializers :smile:.

natefaubion:
> Do you have something in mind for how to get the HTMLElements? That's probably the most important part of why we have initializers. :smile:

There are a couple of things you can do. The _easiest_ thing to do is just what we do now, but rename the `initializer` prop to `ref`, and just use that to get a reference to a DOM node as soon as its available (rather than as an initializer). This is in line with what React is doing now ("ref" takes a callback and gives you the DOM node, and you can `setState` with it).

The other thing you could do is wrap all component state in something like:
```purescript
newtype LocalState s = LocalState
  { state :: s
  , refs :: UnsafeRefMap
  }

type UnsafeRefMap = StrMap HTMLElement
```

Then we can go back to the annotated-tree-method, and tag trees with our dodgy Map. We'd then have a prop like `ref "my-dom-ref"`, and when rendering to vdom, we'd create a hook parameterized with the key and Map, and it would just update the map with the reference. Then you'd maybe add another command to HalogenF to dereference HTMLElements from this map.

I don't really think that the initializer _needs_ to take an HTMLElement, because most often you don't even want the root element, you really want something else inside it. I think something that lets you get arbitrary DOM nodes is more useful.

(As a future note, if you did the `LocalState` thing you could use that to store locally typed props that you could pass to components on each render.)

Another thing, which I didn't add, is I'd like to make a best effort to run the initializers after everything is in the DOM (so you don't have to do hacks to figure out when it's added to the DOM like you do now). This may mean just having another function like `runUIWithElement :: HTMLElement -> Aff _ Driver` that mounts the UI in the given element instead of returning the node.

garyb:
Interesting. So far the only time I've ever used an initializer is when I have a component that is pretty much this:

``` purescript
render _ = H.div [ P.initializer (action <<< Init) ] []
```

The entire use case being to get the current element and then initialize some 3rd part component in this place, and then `eval` is just implementing an interface to the underlying component. I didn't think there were any non-3rd-party-component uses for initializers that couldn't be achieved by other means.

As for detecting whether the element is in the DOM I didn't think that was a problem currently - as long as the root node returned by `runUI` is in the DOM then when an initializer fires you know the node it gives you is also in the DOM.

natefaubion:
I'm thinking of use cases where you aren't _necessarily_ just wrapping a 3rd part library. Say for example, a native Halogen charting component written with canvas. It might look like:
``` purescript
render state =
  H.div [ ... ]
   [ renderLegend state
   , H.div [ ... ]
      [ H.canvas [ ref CanvasRef ]
      ]
   ]

eval (CanvasRef canvasElem next) =
  modify _ { canvasElem = Just canvasElem }
```

> As for detecting whether the element is in the DOM I didn't think that was a problem currently - as long as the root node returned by runUI is in the DOM then when an initializer fires you know the node it gives you is also in the DOM.

But this isn't the case. The node returned by `runUI` is what virtual-dom gives you. By the time you get it, so you can add it to the DOM, all the initializers have already run. So on initial app render, it won't be in the DOM. We would have to defer running initializers by a tick or something, or also give return a `start` function that runs initializers.

garyb:
I guess I would probably wrap the canvas there in a component, but I see what you're saying. I wonder if initializers are finalizers are even useful if we have refs then? As I think in the example above you'd probably want to know when the element disappears too, so `CanvasRef` would probably want to take a `Maybe HTMLElement` instead.

Good point with the initializers on startup case too. I'm just lucky enough to have not encountered that situation yet. I'd be in favour of replacing `runUI` with a version that requires you to pass the parent element in.

natefaubion:
>>> I wonder if initializers are finalizers are even useful if we have refs then?

That's exactly what we have now :smile: (`initialize` is run when it first appears, `finalize` is run when it disappears, it's just not very useful if you have it on the root element).

The distinction that I might make, `ref` gives an input as soon as a DOM node becomes available, changes, or leaves (so `Maybe HTMLElement` as you said), so it's tied to the vdom. While the component lifecycles give you better guarantees about when things are run in the process of rendering.

Decoupling the two makes a difference so you don't have to have a "canonical" root element. For example if I render two different root nodes depending on the state, my initial `HTMLElement` would become invalid, whereas if it was a ref it would change.

garyb:
Haha, I realise that! But I meant `ref` could subsume both if we went with the `Maybe HTMLElement`? Although we'd still need to do something to make the `Nothing` come through when a component is removed from the tree.

Do you have an example case in mind where you might want to use the lifecycle methods? Although given the above I can see that maybe finalize is useful, if we had to handle it separately from the element-removed case.

natefaubion:
The problem with `intialize` as we have it, is that its handled by virtual-dom since it's a hook. Virtual-dom runs it during the process of rendering the VTree (which means our current renderHTML type is actually a lie since it can run these effectful hooks while rendering the tree). It's also subject to patching, which means you could technically swap out 1 component for another, and it wouldn't trigger because the patch process would determine it hasn't changed. I guess this is a problem for anything that relies on virtual-dom, since components are "erased" from the DOM representation.

natefaubion:
As far as the use case for the `initializer` I implemented, it's the difference of "this is run when the node is available, but before it has been patched", and "this is run after the patch has been applied".

garyb:
I realise that, but I'm just having a hard time thinking of a case where I'd actually want that behaviour.

garyb:
My thinking is this: if you just want to do something with a component as soon as it appears, you can do that by passing a query into it from the parent (or app root) immediately.

natefaubion:
> I realise that, but I'm just having a hard time thinking of a case where I'd actually want that behaviour.

So here's a use case: on initial app render, I want to setup this 3rd party editor to be available. If I initialize it using a virtual-dom hook, then it will get initialized before patching (so the DOM node it references won't exist anywhere other than as a fragment). This 3rd party editor reads things like clientHeight and clientWidth or whatever to position certain other elements it adds. Since I'm initializing it before the patch has been applied, it won't measure anything correctly. It needs to be initialized after the patch has been applied.

> My thinking is this: if you just want to do something with a component as soon as it appears, you can do that by passing a query into it from the parent (or app root) immediately.

This would require every component to have a canonical initializer input, and every component has to manually propagate that through the tree. My `initializer` implementation just does that automatically!

garyb:
Ok, apparently I have no idea what my own code does. I thought property patches were applied after the element had been patched, meaning it should exist in the DOM as long as the parent also does, but after experimenting that's clearly not the case. :wink:

What I was trying to propose way up there ^ is that refs should use the `PostRender` mechanism and be plucked out of the DOM, not the existing hook method. But after further thought, I do think arbitrary refs are problematic:
- They need to emulate some of the component lifecycle properties, in that they "initialize" and "finalize" when added to and removed from the DOM, but I'm not sure it's possible to guarantee we will always run the finalizer for them, unlike the component case. It's not strictly a requirement that refs behave this way, but I think it makes them easy to misuse.
- The `PostRender` method wouldn't necessarily always give us the same element back if we did something with a selector and `id`s or `data-` attributes, as virtual dom might shuffle things around (using a `key` in virtual dom would help with this, but generally the problem is still there: at least with hooks we're more certain about when an element is reconstructed).

I don't really know how to get out of this quagmire. Doing our own diffing rather than using virtual-dom perhaps. That's something we've discussed a lot, but it's obviously not a trivial undertaking.

natefaubion:
> Ok, apparently I have no idea what my own code does. I thought property patches were applied after the element had been patched, meaning it should exist in the DOM as long as the parent also does, but after experimenting that's clearly not the case.

I feel this all the time when trying to reason about virtual-dom.

> They need to emulate some of the component lifecycle properties, in that they "initialize" and "finalize" when added to and removed from the DOM, but I'm not sure it's possible to guarantee we will always run the finalizer for them, unlike the component case. It's not strictly a requirement that refs behave this way, but I think it makes them easy to misuse.

This may be desirable, but I don't think any framework out there gives you hooks like this, so this may be a case of just doing the best we can. For example, react only lets you deref the current value (https://facebook.github.io/react/docs/more-about-refs.html). Since components don't have any unique identity within the render tree, there's no way to tell when these underlying references need to change. So yeah, there's no good answer, but I guess we can do what we do now and hope for the best :laughing:

natefaubion:
But in any case you're right, the `ref` as an attribute thing doesn't work for the same reason `initalizer` as a hook doesn't really work (they are subject to patching).

natefaubion:
I think the only thing you could do is probably compose slots. Right now they are local to a component and disappear after installing. But if every component was globally addressable, you could compare for identity that way.

natefaubion:
On a completely unrelated note, if you use the `Spec` variants and provide your functions inline, you don't need any type annotations because they get generalized :P

garyb:
Globally addressable in what sense? That might be a bit tricky to achieve in that they'd need to have globally unique IDs too, wouldn't they? Or is that what you meant about composing them. If so, I think we're already thinking along similar lines.

(Speaking of slot addresses I started looking at your other branch today as well, by the way. It looks really promising, although I'm still wrapping my head around it. I'll take a stab at implementing `transform` to see if I understand it properly).

natefaubion:
> Globally addressable in what sense? That might be a bit tricky to achieve in that they'd need to have globally unique IDs too, wouldn't they? Or is that what you meant about composing them. If so, I think we're already thinking along similar lines.

Yeah, I mean each component can be addressed by its slot, and slots aren't erased, but propagated through another type variable. So a basic component would get a slot of `Unit`. And then a parent component would get a slot of `Either Unit (Tuple p cs)` where `cs` is the composite addresses of the children slots. It's the same composition that we do with queries and state. In the end you just get a big tree of Eithers that can refer to any individual component in the app. This can be compared for equality. So then in the render tree you can compare two different trees in the same location based upon what component built them.

Or something like that, at least. It doesn't help our type errors, that's for sure.

jdegoes:
Since only components can be addressed (and have initializers / finalizers), and since components already support `f` which is unique down to the component, it makes sense to reuse that for a "component address". Rather than introduce a separate mirror term (`Either`s).

natefaubion:
> Since only components can be addressed (and have initializers / finalizers), and since components already support f which is unique down to the component, it makes sense to reuse that for a "component address". Rather than introduce a separate mirror term (Eithers).

I would like to be able to reuse the `f` machinery, but I don't see how that could be used to tag a component. `f` isn't (can't) be `Eq` so I don't see how you could use it to compare which component generated which trees.

jdegoes:
@natefaubion If the blocker is _just_ equality, well that's easy to fix (e.g. `Equal1`), though I suspect there are other issues that need to be sorted out. Still, in principle, I'm convinced it is somehow possible.

natefaubion:
I think it's just that, you don't really _care_ about the `f` itself, all you care about is the path to get to that position. I'd assume an `Equal1` is just something where you have to provide some inhabitant of that functor for it to be equal? If that's the case, then you'd have to at least have some unit constructor in your query algebra that you'd have to provide for this, which is a little odd.

What I was thinking is that given `Coproduct f (ChildF p f')` as the "branching type", all you really need is something like:
```purescript
data Unit1 a = Unit1 -- basically Proxy

instance eqUnit1 :: Eq (Unit1 a) where
  eq _ _ = true
```

and then replace it with `Coproduct Unit1 (ChildF p ...)`, which is just what I was saying with the `Either`s. Or because you don't actually about the _value_, just the tag, use something like:
```purescript
data Path f = Here | There f

Coproduct f (ChildF p ...) -> Path (ChildF p ...)
```

natefaubion:
Hah, I realize that's just `Maybe`. Duh. I think you could actually just build that with `Maybe` and `Unit` while rendering. These types aren't part of the component spec, they are part of the annotation on the trees you render, which the user doesn't see, and we can build automatically. So I don't think there's any actual problem!

natefaubion:
Oh oops, you'd still need to propagate it in a type variable to get the Eq constraint, so I guess that doesn't quite work out as I was hoping :laughing:

natefaubion:
One idea, if we go back to the idea of every component being a virtual-dom widget (which is opaque to the diffing process), then you can propagate a local, existential equality via an annotation which compares the already provided `p` slots. Since the insides of widgets are opaque to diffing (each widget does its own diffing), you'd never be comparing the inner contents of different trees, and the equality will hold. The downside of this is that you'll have more DOM updates since you don't get to diff a global tree.

jdegoes:
> I think it's just that, you don't really care about the f itself, all you care about is the path to get to that position. 

Yeah, exactly. 

> I'd assume an Equal1 is just something where you have to provide some inhabitant of that functor for it to be equal?

It'd be:

```purescript
class Equal1 f where
  eq :: forall a. (Eq a) => f a -> f a -> Boolean
```

Or something equivalent. Though I'm not sure it helps any.

It may help to approach from a slightly different perspective, which is: if I had the address of a component, what would I do with that address? Well, most likely, send the component a message (an `f`).

Which means, the address can be viewed as merely a way of sending a message. 

The main problem with this notion of "address" is that no component really knows its global address. Everything is always relative. However, that's fixable: in particular, in #221, I propose introducing some concept of "nest", and components are always functions from the nest (or, equivalently, and somewhat more nicely, just inside `Reader (Nest ...)`).

If a component knows its nest, then it also has direct access to the globally unique "addresses" of both itself, and its children. "Addresses" then are merely input functors or maybe even whole natural transformations.

natefaubion:
I've implemented the idea above https://github.com/slamdata/purescript-halogen/pull/275#issuecomment-180434754 and it works :+1: 

https://github.com/natefaubion/purescript-halogen/commit/cf7a3b9c140aa3aab35ceec2ba0f13c6710bd230

I'd like to know what you think about this approach. This also potentially lets us have completely local updates so you don't need to rerender the whole tree.

natefaubion:
This approach also takes care of the thunking, so you only do the `HTML -> VTree` rendering on trees that have changed.

natefaubion:
Also, forgot to mention that this change should make the arbitrary `ref` approach above reliable since you aren't diffing trees from different components.

jdegoes:
> I'd like to know what you think about this approach. This also potentially lets us have completely local updates so you don't need to rerender the whole tree.

@garyb has the final say in such things, but from my perspective, it looks like a relatively conservative change that, assuming it's safe and minimal, could be a useful stepping stone. (My longer-term goal with Halogen is to make things simpler, more principled, and less like React, but it's hard to escape the React mold until it's "turtles all the way down", which is closer to a 1.0 goal.)

I believe the main concern will be that it does not address the number one use case for an initializer: which is getting access to the HTML element of the component. Is there some simple way to extend the approach in this direction?

natefaubion:
> I believe the main concern will be that it does not address the number one use case for an initializer: which is getting access to the HTML element of the component. Is there some simple way to extend the approach in this direction?

This change would make the `ref` approach I mentioned earlier reliable https://github.com/slamdata/purescript-halogen/pull/275#issuecomment-179908414

There are a few problems with getting the element as an argument to the initializer: there is no canonical root element for a component (you can completely swap the element for something different in render), the element can change due to reordering in a list (if you don't use a key, which reminds me that there needs to ne

natefaubion:
> I believe the main concern will be that it does not address the number one use case for an initializer: which is getting access to the HTML element of the component. Is there some simple way to extend the approach in this direction?

This change would make the ref approach I mentioned earlier reliable https://github.com/slamdata/purescript-halogen/pull/275#issuecomment-179908414

There are a few problems with getting the element as an argument to the initializer: there is no canonical root element for a component (you can completely swap the element for something different in render), the element can change due to reordering in a list (if you don't use a key, which reminds me that there needs to needs to be a `slotWithKey` combinator). The `ref` approach is more verbose but it lets you get an arbitrary DOM node, and it would fire whenever the element changes.

garyb:
I had a thought last night about the `ref` stuff, it may be redundant with the other changes in here, but perhaps there's still something in it: could we make use of widgets to help there too? If I understand rightly, at least then we could guarantee that we have a constant element for the `ref`?

natefaubion:
No, it doesn't quite guarantee that there is a constant element. In `virtual-dom` if you patch diff two trees that have different roots, and then apply the patch on an element it will swap out that element (which is why `patch` always returns a "new" element). This means in `render` if you do `if foo then H.div_ [] else H.p_ []` then the element in the DOM will not be constant. Most people don't do this though, so its not really a big deal.

The other way you can trigger a new element is when reordering a list of components. I haven't tested what happens when you use a `key` on the widget, but right now if you reorder, and the `eq` check in `update` fails, then it will render a new root element (because it calls `init`). So in this case, the component initializer has already fired, but the root DOM node is changing. Another way to look at it is that you can't necessarily guarantee that a DOM node's lifecycle is tied 1:1 with the component lifecycle.

garyb:
Using `key` will definitely prevent element swapping when applied at the list level, we used it to fix a bug where we had multiple Ace editor components in a list.

What I was thinking with the suggestion here though, is we do some (black) magic during rendering, so that anything with a `ref` attribute is rendered as a widget instead of a vnode, as in that case it's the widget's responsibility to create the element that is rendered, and handle the internal patching of the thing.

Maybe I am barking up the wrong tree though!

natefaubion:
> What I was thinking with the suggestion here though, is we do some (black) magic during rendering, so that anything with a ref attribute is rendered as a widget instead of a vnode, as in that case it's the widget's responsibility to create the element that is rendered, and handle the internal patching of the thing.

That's basically what the component stuff is doing in the widget branch. I don't think there's a need to do that on each `ref`, unless I'm not quite understanding what you are getting at. By making each component a `widget`, a component's tree is _never_ diffed against a different component's tree (it's always local to the instance as defined by it's slot). This would make refs safe since you can't "lose" a ref change by transitioning to a different component in a diff. So as far as I can reason about `virtual-dom`, this should be enough, and I don't think you gain anything by using another layer of widgets. If we aren't trying to provide an element to the initializer, then I think a hook is all you need.

natefaubion:
_Update:_ I've got a stable render cycle without using `rAF`. I just put in the render state a boolean `renderPaused` that is true when running all the lifecycle callbacks. Then after they've all been forked, I have a `flushRender` function that recursively renders (via `tailRecM`) until `renderPending` is false. I also updated eval so that it should only flag `renderPending` when the state is modified (rather than on any `StateF` command). I've fixed a bug in the `Widget` which was causing more rerendering than needed, and extended the lifecycle example with the `ref` method above. As long as you use keys, the DOM for a component is stable. If you leave out a key, then it might get reinitialized (this is expected). Some time this weekend I'll try to submit a patch for `Aff` with a `forkAll` function, and I think we should be ready to go.

The only other thing is what do you want to do about `slot`s that need keys. I tentatively added a `slotWithKey` function (but I don't really like this). Other options:
* If you want a key on a slot, just wrap it in another element that has a key
* Provide a general `withKey` combinator that can work on any `HTML` (I like this more than `slotWithKey`)

natefaubion:
Just pushed a `runUI'` variant which takes a selector to https://github.com/natefaubion/purescript-halogen/commits/widget-trees. If you're OK with this direction then I'll move it to this branch.

PR #276, opened 2016-02-13
Add runChildF to unwrap ChildF queries

garyb:
/cc @jonsterling

### DIFF
--- a/docs/Halogen/Component.md
+++ b/docs/Halogen/Component.md
@@ -173,6 +173,14 @@ algebra with the slot the component was installed into.
 (Functor f) => Functor (ChildF p f)
 ```
 
+#### `runChildF`
+
+``` purescript
+runChildF :: forall p f i. ChildF p f i -> f i
+```
+
+Extracts the query part from a ChildF value, discarding the slot value.
+
 #### `query`
 
 ``` purescript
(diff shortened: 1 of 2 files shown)


jonsterling:
Definitely :+1: for this, assuming Travis succeeds.

garyb:
Do you think the name makes sense? Usually I'd just use `runX` for newtypes, but I couldn't think of anything snappy (`getQueryFromChildF` :wink:).

jonsterling:
Well, I have never liked the `runX` pattern from Haskell in all these years—it always seemed to be a sort of conceit of Haskellers, as if projecting something from a unary product would "run" it.

I sort of prefer the `getX` convention, though even that is not perfect. Maybe `unChildF` could work, but that looks a little weird.

With that said, I am not too worried about whether this is a `newtype`, or indeed, even a unary product, since it is the query that is considered the "main" part of this object in my view, not the slot. Anyway, all this is to say that I don't know of a better name, and I'm fine with whatever you choose.

garyb:
That's pretty much the thought process I ran through (aside from the `runX` musing which I've not considered for a while, I used to wonder why it wasn't `unX` by convention... it sometimes is in Haskell, but I have no idea why one is used over the other), so thanks. Just wanted to check it wasn't totally weird at least.

PR #280, opened 2016-02-22
Initializers/Finalizers squashed

natefaubion:
#275 squashed and rebased on master


garyb:
Just as an update - I am reviewing and trying this out now. Alas it doesn't drop in entirely, SlamData has a couple of problems when built against this branch, so I'm trying to figure out what has changed to cause these issues.

It type checks though :wink:

garyb:
I think I've finally managed to isolate the problematic behaviour: If you have a child component and you use `liftAff`, any changes you make to the state are ignored until the next render cycle.

https://gist.github.com/garyb/def457598f9ba6662f5a

In the example in the gist, the change to the state in `Init` is not rendered until you click the counter button. But this isn't just with initializers, as you'll see the first time you click the counter button its count will stay at 0 even though the message changes to "The initializer is working".

garyb:
I've not looked into the cause yet as I need sleep now :wink: but I suspect it'll be something in the driver: either an AVar read/write mismatch, leaving a hanging read at some point, or equally likely  something to do with `renderPending`/`renderPaused`.

natefaubion:
It's interesting to note, if you just mount the InnerComponent directly, it works as expected. So I'd actually be surprised if it was the driver then. I know you said everything looked fine, but I think something could still be weird in the install machinery, mapping child actions to parent actions or something.

garyb:
Ah, excellent point. And yeah I did notice it only applies when you have `InnerComponent` as a child component... it took me quite a while to boil down this minimal example!

natefaubion:
Or maybe it's just the memo logic? Since that wasn't working at all before, I could just be giving back an old render or something.

natefaubion:
Ah yes, I think that's it (the memo logic)! We only invalidate the `memo` for a component if it receives an input. But we need to invalidate it if there is a state change as well.

garyb:
Taking out the memo lookup does fix it, so yep, I think you're exactly right there with what's happening.

natefaubion:
It's always cache invalidation, isn't it.

garyb:
:smile:

garyb:
I think that's probably the only problem with this too, by the way - SlamData seems to be working as it should with memo disabled. :confetti_ball:

natefaubion:
Actually, it _only_ needs to be invalidated if there is a state change. This will actually prevent unnecessary rerenders, because right now, just receiving an input on a child component requires a rerender as it affects the parent's memo state.

garyb:
Yeah. Ideally we would only re-render when the component's actual state is changed too, not its `children`, `memo` (obviously), etc.

natefaubion:
Off topic, but one idea for that would be to pipe a coroutine to the backing widget. And whenever a component needs to be rerendered, just push a new state to the coroutine.

natefaubion:
I changed the memo behavior and it seems to work as expected now.

garyb:
Yep, thanks! I've been looking into another problem with a `Finalizer` that isn't firing, but I've not found the minimal case for it yet.

garyb:
Actually it just seems like finalizer properties aren't working, although the `componentSpec` version does.

natefaubion:
Yes, I didn't remove it or `initializer`, even though its still just as broken as ever. `initializer` and `ref` are the same, and still work as they did. But I guess we should just remove `finalizer`.

garyb:
Ahh ok. Interesting, it doesn't error because of a `HaltHF` now.

garyb:
Ok, I think I'm happy with this at last! I'll probably make a few more changes before making a release, removing `Initializer` and `Finalizer` and just exposing `Ref` instead, replacing `runUI` with `runUI'`, etc. and then release it as 0.6, since we have a breaking-change dependency bump also.

Thanks a ton for all your work on this @natefaubion!

natefaubion:
:fireworks: 

Docs could use updates as well...

natefaubion:
Oh, one other thing. I had written `runUI`' to take a selector out of convenience (since we had the code already in `Halogen.Util`, but maybe it should take an `HTMLElement` instead?

garyb:
I've made that change locally already :smile:

PR #282, opened 2016-02-24
readonly -> readOnly

cryogenian:
@garyb Is this ok? I've left function name same with attribute not property because
+ I don't want to change API
+ `class_` uses `className` property name.


garyb:
Yep, absolutely. We don't use the exat property name in a bunch of places. :+1:

garyb:
I'll merge/release once Travis completes.

cryogenian:
Thank you

PR #284, opened 2016-02-26
Add drag event handlers

zudov:


src/Halogen/HTML/Events/Types.purs
```diff
@@ -33,6 +36,23 @@ type MouseEvent =
...
+  , clientY :: Number
+  , screenX	:: Number
+  , screenY	:: Number
+  , ctrlKey	:: Boolean
+  , shiftKey :: Boolean
+  , altKey :: Boolean
+  , metaKey	:: Boolean
+  , which :: Number
+  , dataTransfer :: Nullable DataTransfer
+  )
```
zudov: `DragEvent` inherits from `MouseEvent`. I've just copied the fields since it's the most straightforward thing to do. We probably could make it nicer, e.g. by adding `fields` parameter to `MouseEvent` and other events.

src/Halogen/HTML/Events/Types.purs
```diff
@@ -33,6 +36,23 @@ type MouseEvent =
...
+  , clientX :: Number
+  , clientY :: Number
+  , screenX	:: Number
+  , screenY	:: Number
+  , ctrlKey	:: Boolean
+  , shiftKey :: Boolean
+  , altKey :: Boolean
+  , metaKey	:: Boolean
+  , which :: Number
+  , dataTransfer :: Nullable DataTransfer
```
zudov: I am a little bit not happy with `Nullable` here, but it seems that they are supposed to map directly. Is there a nice way to transform it to `Maybe`?

garyb: There's `Data.Nullable.toMaybe`, but we'd have to do something to the event before it reaches the handler to make that work.

zudov: Yes, the question is where to put `toMaybe`. We could just define the handlers like:
```haskell
onDrop f =
  handler
    (eventName "drop")
    \ev ->
      f { button: ev.button
        , detail: ev.detail
        ...
        , dataTransfer: Nullable.toMaybe ev.dataTransfer
        } 
```

garyb: That seems reasonable to me... it does mean we can't coerce the events to an actual `DragEvent` though, so perhaps we should include the underlying event in the record too. I don't know if it would ever be useful, but maybe?

zudov: Yes, that's the thing that I don't like. With other event handlers you can `unsafeCoerce` it to some other record if the halogen's type doesn't mention a field that you need. A nice property when you need a temporary working solution. But constructing the records fairly (providing a minimal value which would satisfy the type) would break that implementation detail.

I am not worried about not having the possibility for hack (it's only better to provide some `self` property with original event rather than coercing the thing), but I would hate inconsistency, so if we do that for `drop` events I think we should do something like that for the rest of them (e.g. make them into newtypes with foreign instances).

zudov: I think it is okay if we merge it as is, and try to come up with a nicer solution in a separate PR.

There is nothing wrong with having `Nullable` there, it's just a small inconvenience of having to use `Nullable.toMaybe`.

garyb:
Sorry I left this hanging. Thanks!

PR #287, opened 2016-03-07
Updates for 0.6 release

garyb:
0.6 will incorporate the lifecycle stuff @natefaubion kindly worked on along with a few other things.


garyb:
Ok, got a few things to update in the README still, and maybe a few other changes to go in if we sort `Affable` out (slamdata/purescript-aff#49), but I'm done messing around with it now.

Changes involved:

Build stuff:
- Use `pulp` rather than `gulp` to build everywhere
- Include a `package.json` in each of the example projects so the `virtual-dom` dependency is present for each, if someone tries to run one without using the `npm run example-blah` approach.
- Did some horrible things to make the travis build "fast" (now it's just really slow, not excruciatingly slow)

Breaking changes:
- Dropped the old `Render`, `Eval` synonyms, since the `s -> ComponentHTML f` / `Natural f (ComponentDSL s f g)` are less opaque and have basically been unused within Halogen since `ComponentHTML`/`DSL` were introduced
- All the component construction functions use a record now, partially to help with the comprehensibility of the `component` type signature now the old synonyms are gone, but also there's no `parentComponent'` now. Instead the expected record for a parent component has a `peek :: Maybe ...`. It's probably 50/50 as to whether a parent component uses peek or not, so it didn't seem bad to do that rather than having two version of `parentComponent`.
- @natefaubion's `componentSpec`/`parentComponentSpec` have been renamed to `lifecycleComponent` / `lifecycleParentComponent`
- `initializer` and `finalizer` properties have gone, in favour of `ref` which fires in the same situations the old properties did, but with a `Maybe` to distinguish between the element appearing and disappearing.
- `runUI` now requires an element rather than providing one that can be attached later. This avoids some issues with `ref`/`initializer` firing before the element is actually in the DOM.
- The stuff in `Halogen.Util` has been reworked to make it more useful with the new `runUI`.

I think that covers it. I'm not sure who should review this really... it's probably more a case of discussing the breaking change choices I made, as the code changes are mostly trivial, so any feedback is welcome @natefaubion / @jdegoes / @jonsterling etc.

natefaubion:
If we are doing breaking changes, how do we feel about changing `InstalledState` to `ParentState` to match the other synonyms (`ParentQuery`, `ParentHTML`, etc)?

natefaubion:
@garyb I know in IRC you mentioned getting rid of `f` in `ParentState`. I think the only thing you could do is use `memo :: M.Map p (Tree f' Unit)` rather than `memo :: M.Map p (Tree (ParentQuery f f' p) p)`. If we do the map fusion on trees or thunk the html on the tree, this is mostly a no op. You'd only pay for it if virtual-dom absolutely had to rerender the memoed tree (which would only be the case if you were moving around a component without using keys).

natefaubion:
OTOH, there's something nice about having `s s' f f' g p` everywhere. It's a bit tiresome but easy to remember.

garyb:
That's true, yeah. I'm fine with leaving it that way too.

I made the `ParentState` change by the way. Do you think it makes sense to call the `s -> ParentState ...` function `parentState` too? Previously was `installedState`.

Just waiting on an update to aff-coroutines now, then I can remove the local `FunctorAff` stuff in favour of `Affable`.

natefaubion:
Yeah, I think `parentState` is the logical choice.

natefaubion:
FYI I've got a branch off of this with a _massive_ performance improvement for render. When do you want to do a release of 0.6? I still need to do some testing, so it's probably best to do it in a follow up.

garyb:
I'm going to finish the README updates and release it tomorrow. Looking forward to the followup :)

natefaubion:
Bug in the lifecycle example: the `Main.purs` file does not define a `Main` module so it doesn't build correctly.

natefaubion:
Right now, the `unhook` for `ref` has the same issue as the old finalizers. Is this expected (ie, you should not use it to do guaranteed cleanup)?

garyb:
> Bug in the lifecycle example

Thanks, I'll fix that.

> Is this expected (ie, you should not use it to do guaranteed cleanup)?

Yep, I'll make that clear in the docs.

garyb:
Ok, I've updated the docs now too. I've separated out a lot of stuff into a separate [GUIDE.md](https://github.com/garyb/purescript-halogen/blob/0.6/GUIDE.md) file now rather than having the massive doc as the README.

@natefaubion could you do a quick scan of the README if you get a minute please? Mainly to check that I've characterised the behaviour of `ref`s and lifecycle handlers accurately, but I've also made a bunch of other changes where things had fallen out of date so hopefully everything should be good now.

natefaubion:
Looks good to me. I noticed a small typo: you mention `eventSource'` when it should be `eventSource_`.

garyb:
Good catch, thanks!

PR #289, opened 2016-03-11
Switch to more reliable Travis build (trusty)

jonsterling:
We've found with the `slamdata` build, as well as `purescript-markdown`, that Travis's `trusty` image is much faster and more reliable for PureScript builds than the default.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
 language: node_js
-sudo: false
+dist: trusty
+sudo: required
 node_js:
-  - 4
+  - 5
 install:
   - npm install
 script:


garyb:
Ha, ridiculous. 1hr 15mins for the last build on `master` and 4mins with this.

paf31:
Is this related to https://github.com/purescript/purescript/issues/1880
?

garyb:
Yes, it's basically a workaround for that. It would still be good to know why it slowed down so drastically between 0.7.5 and 0.7.6 though I think? Since there's a fix now it's significantly less of a concern for me though.

paf31:
Yes, I have a separate issue for the compiler slowing down for the regular (multithreaded) builds.

jonsterling:
@garyb wow!! I had no idea the difference would be so pronounced... it's kind of hilarious.

jonsterling:
you might even say that the build is more *trusty* now...

PR #291, opened 2016-03-14
Render rewrite

natefaubion:
This tracks from @garyb's 0.6 branch. My unscientific perf test is to create a 30x30 grid of counter components (900 child components), and then click a button 10 times. In the old code, I was measuring ~2100ms in garbage collection, and ~1400ms in executing the click handler (so around 350ms per click or so). In this updated code I'm measuring 200-400ms in garbage collection, and ~650ms in the click handler (so around 85ms per click). It's still slower than I'd like, but I don't know how much faster we'll get. At this point the majority of the time is spent in `Data.Map` inserts and deletes, and I don't know how I can minimize those anymore than I have.

### DIFF
--- a/src/Halogen/Component.purs
+++ b/src/Halogen/Component.purs
@@ -42,29 +42,30 @@ import Prelude
 
 import Control.Apply ((<*))
 import Control.Bind ((=<<))
+import Control.Monad.Eff (runPure, foreachE)
 import Control.Monad.Free (Free(), foldFree, liftF, mapF)
 import Control.Monad.Free.Trans as FT
-import Control.Monad.State (State(), runState)
-import Control.Monad.State.Trans as CMS
-import Control.Monad.Writer.Trans (WriterT(), runWriterT)
-import Control.Monad.Writer.Trans as CMW
+import Control.Monad.ST (runST, newSTRef, readSTRef, writeSTRef)
 
-import Data.Bifunctor (lmap, rmap)
-import Data.Foldable (foldMap, for_)
+import Data.Array (cons) as A
+import Data.Array.ST (emptySTArray, pushSTArray, STArray()) as A
+import Data.Bifunctor (lmap)
+import Data.Foldable (foldMap)
 import Data.Functor.Coproduct (Coproduct(), coproduct, left, right)
+import Data.Lazy (defer)
 import Data.List as L
 import Data.Map as M
 import Data.Maybe (Maybe(..), maybe)
 import Data.Maybe.Unsafe as U
 import Data.NaturalTransformation (Natural())
 import Data.Traversable (for, traverse)
-import Data.Tuple (Tuple(..), snd)
+import Data.Tuple (Tuple(..))
 import Data.Void (Void())
 
 import Halogen.Component.ChildPath (ChildPath(..), injState, injQuery, injSlot, prjState, prjQuery, prjSlot)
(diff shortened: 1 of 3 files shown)


natefaubion:
With https://github.com/purescript/purescript-maps/pull/59, garbage collector time down to 150ms, and event handler time down to 470ms :open_mouth:

natefaubion:
If `Data.Map` were to get https://github.com/purescript/purescript-maps/issues/58 then we could cut this even more. Right now installing a component requires a `lookup`, `insert`, and `delete`. If we had `pop` then we could cut out that down to 2 ops instead of 3.

garyb:
Have you tried implementing `pop` locally to see what kind of improvement it gets? I assume there's no way to get clever with `alter` to do what we need either?

Awesome work as usual anyway!

natefaubion:
No, `alter` doesn't work. What you do is take the old children, pop off the ones that are slotted, and add them to the new children. You then have a map leftover with all the removed children.

Using `pop`, garbage collection goes down to 130ms, and < 400ms for the handler. From a cold start, Chrome optimizes a single handler from 75ms to 40ms, and Firefox optimizes it from 150ms to 18ms :open_mouth:

natefaubion:
I think this is in a good state now. I've updated `renderParent` to use `ST` and instead of `fillSlots` and a `State` applicative. It appears to have a significant impact. I'm on a different computer, so I don't have apples to apples comparisons (Chrome won't even profile the current version of Halogen for some reason 😆), but its around 30ms optimized for Chrome, and 17ms optimized for FF. Chrome registers a GC time of around 50ms total.

natefaubion:
Also this should make rendering as stack-safe as it can be, since it's not using `State`.

src/Halogen/Component.purs
```diff
@@ -206,30 +220,33 @@ parentFinalizers
...
-  , children :: M.Map p (Tuple (Component s' f' g) s')
-  , memo :: M.Map p (Tree (ParentQuery f f' p) p)
+  , children :: M.Map p (ChildState s' f f' g p)
+  }
+
+type ChildState s' f f' g p =
+  { component :: Component s' f' g
+  , state :: s'
+  , memo :: Maybe (Tree (ParentQuery f f' p) p)
   }
```
garyb: For maximum performance we'd probably want to avoid records and use data constructors too, as then we benefit from "hidden class" optimisations in Chrome at least, and if I remember rightly they perform better or no worse than anonymous objects in other browsers too.

natefaubion: The reason I haven't done this is a clarity/performance balance. Some of this stuff can get hairy as is, and I just didn't want to make it any more opaque. I'll try subbing in a constructor, but my hunch is that it might not be worth it.

natefaubion: I've tried using a constructor, and I'm not registering a measurable improvement, so I'm going to keep them as records.

natefaubion:
All tests on slamdata pass with this PR.

garyb:
I'll let you do the honours then :smile:

src/Halogen/Component.purs
garyb: Ok cool, I wasn't sure if it would help much, just a thought as switching to the pseudo class representation for the data constructors did wonders for PS performance generally, but that was at the "macro scale" I guess!

PR #295, opened 2016-03-28
Add events for selectedIndex change

jonsterling:
Does this look all right? I think I need this in the `purescript-markdown-halogen` updates.


cryogenian:
:+1:

PR #297, opened 2016-04-17
Added transitionend event

beckyconning:


garyb:
Looks good to me. I'm trying to work out whether it's a breaking change or not though! I think it might be due to the change in type synonym, but then it's a row type, so maybe not...

jonsterling:
@garyb It is a breaking change, I think; consider the case where someone had written code like `InteractiveEvents (onTransitionEnd :: I)`; I think this would be an error now.

PR #303, opened 2016-05-12
Automatically force rerenders

### LINKED ISSUE #302: Remove the need to ever force renders
Manual render-forcing is an unfortunate side effect of having state changes batched before rendering.

This is required when a parent component has modified its state in such a way that it will bring one or more children into existence, and then immediately wants to query the child(ren) in the same `eval` block, something like:

```
modify (addChild c)
forceRender
query c (action SomeQuery)
```

Ideally we can bake in a mechanism to render as needed before a query, or cause querying to wait until `PostRender` if a state change has been made to the child since the last render.

natefaubion:
Fixes #302 

I've added a couple of render-centric constructors to `HalogenF`. This let's us handle the current pending render status, and force one if necessary, when querying children. I think this is probably the least invasive way to do it for the moment, while still maintaining a somewhat well-defined algebra.

I've tested it manually, and it solves the bug case. I've also dropped it into SlamData, while replacing `forceRerender` with a noop that won't trigger rerendering behavior, and all tests pass.

### DIFF
--- a/src/Halogen/Component.purs
+++ b/src/Halogen/Component.purs
@@ -50,7 +50,7 @@ import Control.Monad.ST (runST, newSTRef, readSTRef, writeSTRef)
 import Data.Array (cons) as A
 import Data.Array.ST (emptySTArray, pushSTArray, STArray()) as A
 import Data.Bifunctor (lmap)
-import Data.Foldable (foldMap)
+import Data.Foldable (foldMap, for_)
 import Data.Functor.Coproduct (Coproduct(), coproduct, left, right)
 import Data.Lazy (defer)
 import Data.List as L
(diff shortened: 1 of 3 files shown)


garyb:
Nice! :tada:

A bit of a shame it needed integrating into `HalogenF` itself, but your point in the other issue about the flattening makes sense. We do have `HaltHF` in there too, which also isn't really supposed to be used by library consumers though, so it's not so bad. They can't break anything by using the constructors at least!

:shipit:

natefaubion:
Yeah, I just couldn't think of anyway to do it without adding something to HalogenF 😞. Maybe in a different version we can tease stuff like this out. This formulation of it is at least "safe", in that you can't really mess up anything.

jonsterling:
nice. having to do these manually always worried me...

natefaubion:
I don't know all the details of when we use this in SlamData, but I have a few thoughts:

1. Having to always imperatively update state like this is problematic. If this was something that could be propagated declaratively, it wouldn't be an issue.
2. Multi-staged, async transitions can be somewhat problematic. Not directly related, but I think this issue crops up more in that case.
3. Maybe we need to rethink the app's architecture when these problems crop up. If you need to set state immediately after the component *should* be introduced in the tree, maybe that state should have been propagated via `initialState`.

jonsterling:
@natefaubion I tend to agree; I still think it's important to make sure that it happens automatically, as you appear to have done, in order to enable equational reasoning. It was pretty weird that `pure unit >> m` could behave differently from `m` before.

PR #305, opened 2016-05-12
Make all event handlers use Maybe to allow filtering

garyb:
Also fix the warnings, which will makes this next release incompatible with 0.7.x.


garyb:
I'm expecting this will fail to build first time, as I've not tested all the examples yet.

garyb:
@natefaubion can you take a look please?

The real changes here are that `handler` has been removed in favour of the `handler'` variant that previously existed but was not used in the events we provided: https://github.com/slamdata/purescript-halogen/pull/305/files#diff-2c4146a14fee3fc83b5b024d1425aff7R118. The `Maybe` here allows us to return `Nothing` in response to an event, so we don't have to raise a query for every event, but can filter out events we don't need to act upon. 

There are also changes to update types that used to be `EventHandler i` to `EventHandler (Maybe i)`.

The rest of it is warning fixes for psc 0.8.5.

examples/deep-peek/bower.json
```diff
@@ -1,8 +1,7 @@
...
-    "purescript-halogen": "*"
-  }
-}
+{
+  "name": "deep-peek",
+  "private": true,
+  "dependencies": {
+    "purescript-halogen": "*"
+  }
+}
```
garyb: Some of these diffs aren't very diff-y. Not sure why... line endings? Shouldn't be changing now for any reason I can think of.

natefaubion: I think you're just trying to take credit for everything.

src/Halogen/HTML.purs
```diff
@@ -9,34 +9,36 @@ module Halogen.HTML
...
...datalist, datalist_, dd, dd_, del, del_, details, details_, dfn, dfn_, dialog, dialog_, dir, dir_, div, div_, dl, dl_, dt, dt_, em, em_, embed, embed_, fieldset, fieldset_, figcaption, figcaption_, figure, figure_, font, font_, footer, footer_, form, form_, frame, frame_, frameset, frameset_, h1, h1_, h2, h2_, h3, h3_, h4, h4_, h5, h5_, h6, h6_, head, head_, header, header_, hr, hr_, html, html_, i, i_, iframe, img, input, ins, ins_, kbd, kbd_, keygen, label, label_, legend, legend_, li, li_, link, main, main_, map, map_, mark, mark_, menu, menu_, menuitem, menuitem_, meta, meter, meter_, nav, nav_, noframes, noframes_, noscript, noscript_, object, object_, ol, ol_, optgroup, optgroup_, option, option_, output, output_, p, p_, param, pre, pre_, progress, progress_, q, q_, rp, rp_, rt, rt_, ruby, ruby_, s, s_, samp, samp_, script, script_, section, section_, select, select_, small, small_, source, span, span_, strike, strike_, strong, strong_, style, style_, sub, sub_, summary, summary_, sup, sup_, table, table_, tbody, tbody_, td, td_, textarea, tfoot, tfoot_, th, th_, thead, thead_, time, time_, title, title_, tr, tr_, track, tt, tt_, u, u_, ul, ul_, var, var_, video, video_, wbr)
```
natefaubion: 😱

examples/deep-peek/bower.json
garyb: ... busted

src/Halogen/HTML.purs
garyb: I know :see_no_evil:

natefaubion:
The weird diffs are because of windows line endings everywhere 😱 😄 

screen shot 2016-05-12 at 5 50 43 pm

garyb:
Uergh! I'll fix that..

natefaubion:
They'll always sneak up on you any chance they get. Like .DS_Store.

natefaubion:
Otherwise, everything looks 👍

garyb:
Thanks for taking a look! I've fixed the line endings, so just waiting to see if Travis will ever run this before I merge.

Everything built locally at least.

garyb:
Hah! As if by magic, I insult Travis and the build starts.

PR #306, opened 2016-05-12
Add tabIndex, add draggable to indexed props

### LINKED ISSUE #298: Tabindex property
Could you please add property `tabindex`.

garyb:
Resolves #298

Not sure why there wasn't an `Indexed` version of `draggable`, but there is now.

@natefaubion if you could take a look again? After that let's release the new version :white_check_mark:

### DIFF
--- a/src/Halogen/HTML/Properties.purs
+++ b/src/Halogen/HTML/Properties.purs
@@ -2,15 +2,16 @@
 module Halogen.HTML.Properties
   ( key
   , alt
+
   , charset
-  , class_
-  , classes
+  , class_, classes
   , cols
   , rows
   , colSpan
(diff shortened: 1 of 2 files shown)


src/Halogen/HTML/Properties.purs
```diff
@@ -2,38 +2,45 @@
...
   , placeholder
   , autocomplete
   , autofocus
   , multiple
+
   , draggable
+  , tabIndex
+
   , ref
+
```
garyb: This diff noise is because I matched up the spacing and organisation with the list in `Indexed` so it's easier to see the differences side by side.

natefaubion:
👍

PR #315, opened 2016-07-12
Halogen next (existential state, child queries, component-level incremental updates, map fusion over HTML, etc)

### LINKED ISSUE #208: Halogen and html
Is there any plans to split halogen and html similar to how react was spitted to react and react-dom?
So you could use halogen with any UI backend like canvas, mobile UI, etc and not only html.

### LINKED ISSUE #245: Alternatives to peeking
Currently the only way to signal from child to parent is through peeking, which is awkward to me for a number of reasons:
- It's yet another signalling mechanism. There's handlers for the HTML DSL, coroutines for third party components, and peeking for native components.
- It adds another layer of overloading to the query algebra. The query algebra is currently used for internal actions, external actions, external requests, parent signalling, and memo invalidation.
- It requires you to overload individual query constructors. Using something like the Ace example, you have to overload the `ChangeText` action for both signalling and performing updates, which requires a redundant check to only update if the text is different.
- It's inefficient. If you peek at your root component, it will get invoked on every single input, even though 99% of inputs will likely be ignored.
- It's unwieldy to peek at anything other than immediate children.

One idea (and you all can comment on the feasibility) is to let components opaquely reflect their current slot/address during `render` and `eval`. Essentially, this is just getting something like a `this` reference to the component, or kind of like how elm has addresses (except these are pure). It might look like this (with a lot of hand-waving):

``` purescript
type Address = ...

data Message a = ...

instance contravariantMessage :: Contravariant Message

envelope :: forall f a. (a -> f Unit) -> Address -> Message a
```

You could then pass around `Message`s to child components, and provide someway of invoking them. Since Halogen provides the opaque address, it's guaranteed to resolve to a well-formed action. And since you can only reflect your own address or take one from a parent, you can't get into a weird muddy situation where you can globally address any component in the hierarchy like you can in elm through the `Mailbox` to `Signal` mapping.

You would also want to change the HTML DSL to use this mechanism, which has some advantages:
- It unifies the signalling mechanism.
- You could take a `Message` as a parameter and give it directly to an HTML event without needing to go through some sort of proxy in your query algebra
- You can `contramap` if you need to adapt the input
- It would be more efficient in general. Right now, because we carry around `i` in the HTML DSL, in order to nest components we have to `map` and adapt the handler. This means that child components in a deep hierarchy get mapped over and over again at each level of nesting. Since all handlers would just take `Message`s, you could lose the type parameter.

I think we have all the information around to do something like this anyway (as we have to `map` over the HTML to adapt handlers), but it would be a major change. Any thoughts?

### LINKED ISSUE #277: `Par` for child queries?
Sometimes you need to issue a bunch of queries to child components (to say refresh some data in a cell). There are no `Par`-like primitives in HalogenF though, so using something like `for_` or `traverse_` will cause them to run one after the other, veeery slowly.

### LINKED ISSUE #288: Simplified component signatures
Right now, state is totally opaque to the outside world, yet we still have to thread around the composed type. Additionally, parent eval functions have to thread around the child state even though nothing can be done with it. If we want to commit to this route, with a small change we can eliminate a lot of the type boilerplate. The following is essentially the same thing we do now, but because we move the initial state into the component definition, we no longer have to deal with threading the type around, so we can focus on the only interface we actually expose: the query algebra.

``` purescript
type Render s f g =
  { state :: s
  , hooks :: Array (Hook f g)
  , tree  :: Tree f Unit
  }

-- By moving `initialState` to the component spec we can existentially hide it.
-- Components should be parameterized if they need to take a "custom" initial state.
type ComponentImpl s f g =
  { render :: s -> Render s f g
  , eval :: Natural f (ComponentDSL s f g)
  , initialState :: s
  , initializer :: Maybe (f Unit)
  , finalizers :: s -> Array (Finalized g)
  }

foreign import data Component :: (* -> *) -> (* -> *) -> *

mkComponent :: forall s f g. ComponentImpl s f g -> Component f g

runComponent :: forall f g r. (forall s. ComponentImpl s f g -> r) -> Component f g -> r

type InstalledImpl s f g =
  { component :: ComponentImpl s f g
  , state :: s
  , memo :: Maybe (Unit -> Tree f Unit)
  }

foreign import data Installed :: (* -> *) -> (* -> *) -> *

mkInstalled :: forall s f g. InstalledImpl s f g -> Installed f g

runInstalled :: forall f g r. (forall s. InstalledImpl s f g -> r) -> Installed f g -> r

newtype ParentState s f' p g = ParentState
  { state :: s
  , children :: Map p (Installed f' g)
  }

type ComponentHTML f = HTML Void (f Unit)

type ComponentDSL s f g = Free (HalogenF s f g)

type ParentQuery f f' p = Coproduct f (ChildF p f')

type ParentHTML f f' p g = HTML (Slot f' p g) (f Unit)

type ParentDSL s f f' p g = Free (HalogenFP ParentEventSource s f (QueryF s f' p g))

type QueryF s f' p g = Free (HalogenF (ParentState s f' p g) (ChildF p f') g)

data Slot f p g = Slot p (Component f g)

---

data Query

someComponent :: Component Query (Aff ...)
someComponent = component { render, eval, initialState } where

  initialState :: Int

  render :: Int -> ComponentHTML Query

  eval :: Natural Query (ComponentDSL Int Query (Aff ...))


data PQuery

anotherComponent :: Component (ParentQuery PQuery Query Int) (Aff ...)
anotherComponent = parentComponent { render, eval, initialState } where

  initialState :: String

  render :: String -> ParentHTML PQuery Query Int (Aff ...)

  eval :: Natural PQuery (ParentDSL String PQuery Query Int (Aff ...))
```

So basically, the user doesn't ever have to know or write `ParentState`. Parent eval's have the expected type: parent state, parent query, child query, slot, base functor. Is this a direction we want to move in, or do we want to figure out how we can open up state more? If we did this #221 would become impossible.

### LINKED ISSUE #290: Simpler combinators for slot/query/state unions, child paths, etc
I think everyone knows that dealing with the `Either/Coproduct` stuff with parent components is a bit of a chore. I think at one point @garyb played with the idea of using `either2of3` and `coproduct1of4` style combinators. I actually don't think that's a terrible approach, but the main problem with that is it isn't extensible, so when you go to add another data type you have to update all of your combinators.

One idea is if, instead of using the equivalent of this (which causes the extensibility issues):

``` purescript
type Either4 a b c d= Either a (Either b (Either c d))
```

We did this:

``` purescript
type Either4 a b c d = Either a (Either b (Either c (Either d Void)))
```

Where you use `Void` as a nil-type (and likewise `Const Void` for `Coproduct`). You could then define extensible injector/projectors polymorphic over the "tail".

``` purescript
inj3 :: forall a b c z. c -> Either a (Either b (Either c z))
prj3 :: forall a b c z. Either a (Either b (Either c z)) -> Maybe c
```

If we look at slots, and you are using the child path machinery, your slots are mostly unit types. Or if you have a list of some type of component, it will be indexed by an id of some sort (almost always an Int, Number, String). These means that 98% of the time, we don't actually need to define our own slot types, because the `Either` spine is really what we care about.

``` purescript
-- Unit slot (use inj3 for an indexed slot)
at3 :: forall a b z. Either a (Either b (Either Unit z))
```

I think this could eliminate a lot of boilerplate around defining `ChildPath`s. Since they are extensible you don't need to rewrite the names if you add a new data type, and you can just use them inline (the type inference should just work out).

### LINKED ISSUE #322: Rename `Halogen.HTML.Indexed` to `Halogen.HTML`
Since it is recommended to use the indexed variants

``` purescript
import Halogen.HTML.Indexed as HH
import Halogen.HTML.Events.Indexed as HE
import Halogen.HTML.Properties.Indexed as HP
```

wouldn't it make sense to rename those modules to

``` purescript
import Halogen.HTML as HH
```

and use the longer form for the non-indexed (and probably lesser-used) variants?

``` purescript
import Halogen.HTML.NonIndexed as HH
import Halogen.HTML.Events.LessSafe as HE
import Halogen.HTML.Properties.WhatEver as HP
```

garyb:
Resolves #208, resolves #245, resolves #277, resolves #288, resolves #290, resolves #322.


jdegoes:
@natefaubion @garyb Consider also fixing `g` to be `Aff`. Or better yet, `IO`, and removing free from the output functors entirely. Short of this, consider letting `g` be totally generic (not like `HalogenDSL g` etc) and using type classes to represent the set of capabilities the monad is required to possess (including `Affable` or `MonadAff`). This way you can back with Free or Aff or anything else.

garyb:
@jdegoes I think we're already doing that - or the new approach does avoid `g` being anything in particular, until you get to the driver at least. We no longer have the case where `g` is `QueryF ... g ...` for parent components and then re-interpreted into `g`.

That was part of moving the "glue" to the driver, and also is beneficial from the performance side of things as we no longer have to do any mapping/interpreting of components within the hierarchy, and therefore avoids the problems we were having with the free-related GC overhead.

jdegoes:
@garyb As far as I can see, you still have `eval :: forall m. Query ~> Free (H.ComponentDSL State Query m)`, although it looks cleaned up in some places. I think `m` should be totally polymorphic, with `MonadAff` and other constraints (e.g. `MonadHalogen`) as appropriate.

natefaubion:
> I think m should be totally polymorphic, with MonadAff and other constraints (e.g. MonadHalogen) as appropriate.

Please not until we have FDs. The type inference will be so bad it will be nearly unusable.

garyb:
I anticipate that being a problem too, but haven't tried it to be sure. Since lifting anything with rows only works through luck*, it's almost certainly going to be problematic currently.

* A bug

jdegoes:
@garyb Why would you have to use rows anywhere?

garyb:
Assuming you wanted to use constraints throughout rather than fixing `g ~ Aff` within a component you'd need to lift the `Aff` actions (as we do currently, via `Affable`). It mostly works, but there are cases where we have to explicitly annotate. I imagine there will be other cases that require annotation if we switch to the constrained-`g` approach, but FDs will fix all that. They should be coming in 0.10 though!

jdegoes:
You'd lift the `IO` actions. There would be no `Aff` actions. 😈

garyb:
You still have to lift the `Aff` actions into `IO` first, since we use `Affjax`, `AVar`s, etc :wink:. Although I guess no rows = no problems.

Ideally there will be no `Aff` actions anyway, since the changes that have already gone into `g` mean we can actually make things generic and use algebras rather than being implicitly stuck with `Aff` now. Despite having `Affable` previously there were issues in how event subscribing worked that meant you ended up requiring `Aff` to make the types work when subscribing in a parent component, but now we don't have the "reinterpret-QueryF-into-g" thing I think it works out now... but need to double check.

natefaubion:
@garyb I think subscribing will always require some sort of `Aff*` constraint since it involves `Producer`s.

jdegoes:
One of the many reasons I'm inclined to switch over to `IO` entirely.

Type classes can convey how powerful a monad is much more effectively, and you can implement them with either free or low-level effects, but the vast majority of code can be polymorphic in the monad and just require things like `MonadIO`.

natefaubion:
In any case, we still need to wait until we have FDs, because any hypothetical `MonadHalogen` will need to be parameterized by the local state and child query types.

garyb:
@natefaubion I realised the previous symmetric difference stuff with initializers and finalizers was a bit silly, as we can detect new component additions easily without having to diff the slots (by moving `addInitializer` into the point that we instantiate a driver). Still diffing for the removed component case though, and I couldn't really see how `pop` would help us out there. Key diffing is perhaps something we could PR to maps though, as `sets` would be able to use it also.

I attempted the changes to avoid `OrdBox` by packaging up the map operations in the state, but it got pretty ugly and I still ended up with a missing `Eq` instance at the end since it was needed for the slot diffing. I could have done something similar there, but I wasn't crazy about the changes that were required to get it work, since you have to wrap up the map ops as far back as `Component'`, so it kinda leaks everywhere. If you want to give it a try I'd still be open to it though, perhaps there is a better solution.

I put in a version of forking in that I'm pretty sure is unsafe, but it can definitely be fixed if we introduce a `MonadAff` constraint for the `HalogenM` instance for `MonadFork`. I forgot about the flaw in the current thing but then thought more about it and realised where the problem was.

Subscriptions now operate in the `ComponentDSL` rather than returning a query to feed back to the component's `eval`. I think that's probably better? I haven't done anything about improving cancelers for subscriptions yet though. That's probably the last missing feature.

natefaubion:
> Still diffing for the removed component case though, and I couldn't really see how pop would help us out there.

If you have a Map of old components, and an empty Map for new components, as you walk over the slots in a render cycle, you `pop` the slot off the old Map. This performs a look up and removal at the same time. If it was there you just push it on to the new Map, and at the end your "old" Map is comprised of all the components that no longer exist.

garyb:
Makes sense, I was only thinking about how to deal with the maps post-render.

natefaubion:
> I attempted the changes to avoid OrdBox by packaging up the map operations in the state, but it got pretty ugly...

I don't think this is a show stopper by any means. It's a little "weird" and _maybe_ a little inefficient, but it's not unsafe or anything. The only real concern I had was the unpacking of the entire component set on every child query, and I don't think that's an issue any more.

> Subscriptions now operate in the ComponentDSL rather than returning a query to feed back to the component's eval. I think that's probably better?

I don't know about this. It's kind of the same issue with having ComponentDSL in event handlers. It can be convenient, but it also encourages out-of-band transitions which are impossible to invoke directly. It's always possible to just invoke `eval` in this block, but I don't know, I'm a little apprehensive.

> That's probably the last missing feature.

I think the only other thing would be using extensible coproducts for slots and getting rid of the `cpR :> cpL` stuff, and just having `to4` or `at4` or something.

natefaubion:
> I put in a version of forking in that I'm pretty sure is unsafe, but it can definitely be fixed if we introduce a MonadAff constraint for the HalogenM instance for MonadFork. I forgot about the flaw in the current thing but then thought more about it and realised where the problem was.

I consider `MonadFork` a convenience (and possible footgun as it can potentially have the same out-of-band transition effect) that's unnecessary 95% of the time, so I'm actually OK with requiring MonadAff in this case, as long as people aren't _required_ to implement MonadAff in `m` _unless_ they use `MonadFork`. If at some point we can get rid of it, great.

garyb:
> The only real concern I had was the unpacking of the entire component set on every child query, and I don't think that's an issue any more.

Yeah, we only have that now for `GetSlots` - which is just used in the `queryAll` case, so the overhead is much reduced.

> I don't know about this. It's kind of the same issue with having ComponentDSL in event handlers.

Good point. I think it's something I'd started working on before and then left half done, as it didn't work either way properly when I started on it yesterday. I'll switch it back to use queries.

> I think the only other thing would be using extensible coproducts for slots

Oh right! I totally forgot about that. I'm not sure we can entirely avoid the `ChildPath` stuff since it's a combination of slot either and query coproduct, but we can make an extensible version of that too and use the `at` style, with the coproduct/either stuff underlying.

> If at some point we can get rid of it, great.

Yeah that's exactly how I feel about it too. It doesn't require the constraint until you try to use forking, so it's an inconvenience more than a problem. The fact we'll be able to fork `ComponentDSL` at all is going to pretty nice for those occasional cases where it's "needed"!

natefaubion:
> I'm not sure we can entirely avoid the ChildPath stuff since it's a combination of slot either and query coproduct

Yeah we still need it, but you don't have to define your own awkward compositions. You should be able to just use the predefined ones inline and the inference will work out.

lethom:
I do not know if it is the correct way to suggest this, but have you considered adding something like [this utility functions](https://github.com/pshendry/purescript-pshendry-halogen/blob/master/src/Halogen/Component/Utils.purs) into Halogen? Especially, the `raise` function is very useful, IMO.

garyb:
@lethom heh, those come from SlamData actually. They'll be redundant in the new version since there's a real mechanism for raising messages.

pshendry:
And here I am exposed as a criminal for lifting those methods without retaining the license and copyright notice. I'll see you in court!

Really looking forward to Halogen vNext, thanks for all the work on that front :)

arthur-xavier:
Any plans for a release date @garyb? I'm so looking forward to test it! :)

garyb:
Release date is "sometime this week" :smile: - that last merge that went into this branch means it is now as good as done, but I need to write up how it works - both revising the main guide, and providing a "changes from previous" type guide. I'm also eager to get it released and out there!

garyb:
I'm going to merge this as it is and continue to work on the readme stuff separately, using GitBook. Hopefully it'll be a lot easier to navigate and find stuff in the guide that way.

PR #317, opened 2016-07-23
Misc updates

### LINKED ISSUE #307: Use Affable constraint rather than MonadAff for event sources

### LINKED ISSUE #308: Add field to `HaltHF` to capture a message
So that if the worst occurs and a `HaltHF` arises we will know which invariant was violated.

garyb:
- Updated to latest coroutines
- Resolves #307 
- Resolves #308

### DIFF
--- a/src/Halogen/Driver.purs
+++ b/src/Halogen/Driver.purs
@@ -5,18 +5,18 @@ module Halogen.Driver
 
 import Prelude
 
-import Control.Bind ((=<<))
 import Control.Coroutine (await)
 import Control.Coroutine.Stalling (($$?))
 import Control.Coroutine.Stalling as SCR
 import Control.Monad.Aff (Aff, forkAff, forkAll)
 import Control.Monad.Aff.AVar (AVar, makeVar, makeVar', putVar, takeVar, modifyVar)
 import Control.Monad.Eff.Class (liftEff)
+import Control.Monad.Eff.Exception (error)
+import Control.Monad.Error.Class (throwError)
 import Control.Monad.Free (runFreeM)
 import Control.Monad.Rec.Class (forever, tailRecM)
 import Control.Monad.State (runState)
 import Control.Monad.Trans (lift)
-import Control.Plus (empty)
 
 import Data.Either (Either(..))
 import Data.Foldable (class Foldable, foldr)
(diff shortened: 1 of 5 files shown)


src/Halogen/Driver.purs
```diff
@@ -131,7 +131,7 @@ runUI c s element = _.driver <$> do
         when (isJust rp) $ render ref
         putVar rpRef Nothing
         q
-      HaltHF -> empty
+      HaltHF msg -> throwError $ error msg
```
natefaubion: I'm not sure we want to actually throw an error. Won't that totally kill the driver? I think we probably just want to log it to `console.error` maybe.

garyb: It's the same behaviour as before - `empty` throws a runtime exception in `Aff`. We need something like `empty` or `throwError` so the types work out.

natefaubion: I say that because I'm pretty sure this gets triggered in SlamData. It can happen even if you have a long running request, and you remove the component before the request completes. It's not necessarily a failure condition, just a timing issue.

garyb: Hmm, I can indeed see how that would result in this happening. I don't know how to fix it though - and if it's currently happening, doesn't that mean SlamData is already broken?

garyb: For reference: https://github.com/slamdata/purescript-aff/blob/f9b890130eb12e02b65d5ff5c6895acf393b1a8f/src/Control/Monad/Aff.purs#L195-L196

natefaubion: OK, then I guess I'm wrong :). I only brought it up because I know whenever we've had unhandled exeption in the FFI crop up during rendering it totally nukes the driver (app becomes unresponsive to inputs, no errors show in the console). But if it's that case that `empty` throws an error, are these errors even propagated anywhere? I've can't say I've ever seen the "Always fails" error, and I've definitely triggered Halt conditions. Maybe its something worth testing?

natefaubion: I know Aff got a rewrite of the exception machinery recently, so maybe it was just broken before?

natefaubion: Thinking about it some more, I think the only reason exceptions during rendering nuked the driver was because it results in an `AVar` not being put back, creating a dead lock.

garyb: We could probably do something to ensure we don't end up with hanging vars by doing something like this in the cases it might occur internally:

``` purescript
x <- thingThatMightFail
x <|> (putVar oldValue ref *> x) 
```

natefaubion:
I think this is good to go 👍

PR #318, opened 2016-07-25
Abstract 'HTML' out of Components and Driver

doolse:
Really this is addressing #208 . 
I want to use Halogen with React-Native, which I initially did as a proof of concept by hijacking the HTML dsl, but I wanted to do it "properly" so hence this pull request.

To some up the changes, which have remained source compatible:
* New types for "Component" et all, parameterized by the render DSL type.  All of which are prepended with "Raw" (I'm not sold on that, perhaps just prepend a H?)
* Type synonyms for all the existing types to specialize to HTML.
* A RenderDSL type class for abstracting the HTML part of rendering child components, and creating the "TreeF" structure.
* Abstracted the Virtual DOM part of the runUI into a new "runDriver" function which takes functions for initial rendering and updating of the render tree.

All of this has allowed me start work on "doolse/purescript-halogen-reactnative" for the actual react native  DSL, still very much in the initial stages but it's working.

This is my first attempt at anything useful with a haskell-ish language so if you see anything noobish please give me your constructive criticism :)


natefaubion:
Thanks for the PR @doolse! I think that decoupling a specific HTML representation from the core machinery is definitely a good idea. We have a few changes we are working on that might obviate some of what's done in this PR, though. 

Probably the biggest one that will affect your usage is that we pushing the component composition machinery to the driver. Right now, component composition is completely opaque, and a parent component handles the routing/embedding of children. As the changes in your PR highlight, this is not ideal and makes it difficult to tease apart the HTML representation. By pushing all that to the driver, it will get to be a bit more "hands on" giving us better optimization opportunities. It also means that implementing a new backend should only require implementing a new driver, and while you'll have to implement more than what you do now, it should give you a lot more flexibility.

doolse:
Thanks @natefaubion for the update, is there an issue or branch that I should monitor for progress?

natefaubion:
@doolse https://github.com/slamdata/purescript-halogen/pull/315

Our goals being:

* Move to existential (private) state. State is already effectively private, but you still have to deal with the awful types. We want to remove the state type from the public interface of a component.
* Move to level-by-level querying. If you use the `ChildPath` machinery, this is already the case because you can't make a `ChildPath` that descends into `ParentState`, but like with state, you still have to deal with the awful types. So we are only going to expose parent-to-immediate-child querying.
* Remove peeking. Peeking is bad for modularity, so we want to remove it in favor of adding an "output message" type to the component interface. Basically just a way to raise events within your component DSL. This makes component events more like DOM events, and unlike peeking, these can be bubbled (though explicitly).
* Move component glue to the driver. Right now the driver has no interesting knowledge about your component hierarchy so it can't do any interesting optimizations (it's only a dumb interpreter). With the component glue being in the driver, the driver can make decisions about how to allocate state for components, it can efficiently route inputs/queries, and it can incrementally render the component tree. Hopefully this will solve the rest of our performance woes since we won't need to nest DSL interpretations.
* Somewhat orthogonal, but we also want to move to extensible coproducts for child paths and slots, so they should be much easier to use (no need to do the weird manual composition).

I think these changes will make everything much more approachable and easier to use (we definitely feel these pains in SlamData). Obviously your input is very much appreciated, and I hope we can work out a solution for your use case as well.

doolse:
@natefaubion 

I was thinking about this quite a bit yesterday as I ran into a problem with parent components not being able to generate react callbacks which could render child components. The conclusion I came to was that I couldn't really do it with ParentComponent's and would have to push it to the driver by creating multiple React classes rather than a single application wide ReactClass. This really aligns with the changes you'll be making. I'm going to implement my own temporary work around but I expect it will be easy to convert after these changes. 

> Move to existential (private) state. State is already effectively private, but you still have to deal with the awful types. We want to remove the state type from the public interface of a component.

+1

> Remove peeking. Peeking is bad for modularity, so we want to remove it in favor of adding an "output message" type to the component interface. Basically just a way to raise events within your component DSL. This makes component events more like DOM events, and unlike peeking, these can be bubbled (though explicitly).

+1. I would go as far as to say that output events shouldn't be any different from DOM events in how they are used, so rather than output type, you have a "props" type which you pass in handlers as you would for normal DOM elements.

> Move component glue to the driver. Right now the driver has no interesting knowledge about your component hierarchy so it can't do any interesting optimizations (it's only a dumb interpreter). With the component glue being in the driver, the driver can make decisions about how to allocate state for components, it can efficiently route inputs/queries, and it can incrementally render the component tree. Hopefully this will solve the rest of our performance woes since we won't need to nest DSL interpretations.

+1. This makes sense for React as it does a good job of handling the rendering/state side.

PR #321, opened 2016-09-07
Use a phantom type to restrict LengthLiteral to correct length type

berdario:
I tried to set an img width to `Percent 100.0`, but in the DOM you'll only get back a img with width `0`

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-width

> The intrinsic width of the image in pixels. In HTML 4, either a percentage or pixels were acceptable values. In HTML5, however, only pixels are acceptable.

I was about to send a PR to remove LenghtLiteral altogether, but then I realized that object-like elements, like `object`, `embed` and `iframe` support a percent-size

This is the first time I actually used a phantom type, and I'm not sure if the names I've picked are actually good.

### DIFF
--- a/src/Halogen/HTML/Elements/Indexed.purs
+++ b/src/Halogen/HTML/Elements/Indexed.purs
@@ -125,6 +125,7 @@ module Halogen.HTML.Elements.Indexed
   ) where
 
 import Halogen.HTML.Core (HTML)
+import Halogen.HTML.Properties (Object, Element)
 import Halogen.HTML.Properties.Indexed hiding (title)
 import Halogen.HTML.Elements as E
 import Halogen.HTML.Elements
@@ -317,7 +318,7 @@ br = unsafeCoerce E.br
 button :: forall p i. Node (autofocus :: I, disabled :: I, form :: I, formaction :: I, formenctyp :: I, formmethod :: I, formnovalidate :: I, formtaget :: I, buttonType :: I, value :: I) p i
 button = unsafeCoerce E.button
 
-canvas :: forall p i. Leaf (width :: I, height :: I) p i
+canvas :: forall p i. Leaf (width :: Size Element, height :: Size Element) p i
 canvas = unsafeCoerce E.canvas
 
 caption :: forall p i. Node (align :: I, onScroll :: I) p i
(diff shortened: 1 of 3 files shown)


garyb:
This approach seems like a good idea, but how about we squash the construction functions back into two options:

``` purescript
percent :: Number -> LengthLiteral Object
pixels :: forall a. Int -> LengthLiteral a
```

I was musing some naming like `Absolute` / `Relative` rather than `Object` / `Element` too, but it's a bit weird, it would probably have to be `OnlyAbsolute` / `AllowRelative` or something to make sense, which doesn't seem much better really.

berdario:
Agree, I've done it.

re. naming:

I hoped it'd be possible to do something like
```
canvas :: forall p i. Leaf (width :: Size Absolute, height :: Size Absolute) p i
embed :: forall p i. Node (height :: Size (Absolute | Relative), src :: I, mediaType :: I, width :: Size (Absolute | Relative)) p i
```

But I'm afraid that to do something like this we'd need another ADT (called something like `AllowRelative`, which is not buying us any clarity)... maybe with rows inside rows? Or maybe Type families? (which I know haven't been implemented yet, but I was wondering what's the best we could theoretically do, given a sufficiently advanced compiler)

garyb:
You could probably do something with `DataKinds` that would better express it, as then you'd restrict the phantom type to constructors of an `AllowRelative` type (`DataKinds` allows you to promote data constructors into the type system). Something like:

``` purescript
data LengthLiteral (a :: LengthRestriction) = ...
data LengthRestriction = AllowRelative | OnlyAbsolute

canvas :: forall p i. Leaf (width :: Size 'OnlyAbsolute, height :: Size 'OnlyAbsolute) p i
embed :: forall p i s. Node (height :: Size s, src :: I, mediaType :: I, width :: Size s) p i
```

In `embed` here `s`'s _kind_ would also be `LengthRestriction`, and the only types of that kind are `'AllowRelative` and `'OnlyAbsolute`, so you wouldn't need to express the union explicitly as suggested above. (The prime prefix is a Haskell thing, we wouldn't necessarily require that, but I've not looked into the arguments and reasoning about why it's so to be able to say for sure).

garyb:
This looks good to me :+1:. I'll keep the PR hanging around until #315 is resolved though, and merge it once that is in and we're ready to release, as I might forget to credit you in the release notes otherwise.

garyb:
Thanks! Sorry about the ridiculous delay in getting it merged.

berdario:
Don't worry... I've done worse :) (issues unanswered for 7 months :P )

PR #323, opened 2016-10-13
handle null in vdiff and exception in patch

cryogenian:
This fixes weird bug when `vtree1` is `null`. I have no idea how that's possible, probably bug in _virtual-dom_

### DIFF
--- a/src/Halogen/Internal/VirtualDOM.js
+++ b/src/Halogen/Internal/VirtualDOM.js
@@ -101,7 +101,10 @@ HalogenWidget.prototype = {
       this.el = prev.el;
     } else {
       this.vdom = this.render(this.tree);
-      this.el = vpatch(node, vdiff(prev.vdom, this.vdom));
+      try {
+        this.el = vpatch(node, vdiff(prev.vdom, this.vdom));
+      } catch (e) {
+      }
     }
   }
 };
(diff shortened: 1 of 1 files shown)


garyb:
Wow, how did you trigger the case where this fails?

cryogenian:
+ During fixing flexible visualization tests I found that urlvarmaps don't propagate to decks. 
+ I've modified `evalQuery` and __BOOM!__ _SetText@purescript-ace-halogen_ stopped working. But only if there is modification of `Card.Ace.Component`'s state. 
+ Tons of debugging and `diff` in _virtual-dom_ takes `null` as first element. 

I haven't found why it's `null` though 😞

garyb:
@natefaubion do you have any idea what could be going on here?

cryogenian:
Oops, widget update's diffs don't need this `|| {}` :)

natefaubion:
I'm not sure exactly why it would be null, but I don't think this is the correct fix. This is just going to mask potential errors. I know we want to get it working, but this is a very heavy handed fix, and we are likely doing something wrong somewhere that's leaking through (I don't think this is a bug in virtual-dom).

cryogenian:
Maybe it's something like diffing thunks and vtrees?

natefaubion:
We don't use actual virtual-dom thunks to do our thunking. And if it is something as trivial as that we would have seen it earlier as our thunking kicks in all the time. We should setup some logging or breakpoints that get flagged when this stuff is null to try to figure out where exactly its getting nulled.

cryogenian:
Hm, seems that that was misuse. Closing this pr

Issue #329, opened 2016-11-04
The .gitignore for the repo doesn't exclude outputs from the examples.


philderbeast:
I made a change to fix an issue to do with the examples but find that there are thousands of pending changes. The .gitignore for the repo doesn't exclude outputs from the examples.

philderbeast:
When I build the multicomponent example, in each GUI git client I have, I see 7227 non-versioned files, mostly in the examples/multi-component/node_modules folder. I can see the same thing on the command line with ...

`>git status --untracked-files`

philderbeast:
To reproduce the problem, run `npm install` from the folder `examples/multicomponent`, ignoring the advice in `examples/readme.md` to do this install from the repo root. Then check for untracked files that aren't ignored with `git status --untracked-files`.

garyb:
Yeah, sorry for the confusion here - the individual example `.gitignore` files don't cover `node_modules`, but they should - my only objection to the PR you opened is that we should perhaps update the example-level files rather than the root one (it doesn't really matter I guess, but since we have them per example it seems reasonable to apply example-related rules there).

PR #330, opened 2016-11-04
Exclude example build outputs (#329) and vim stuff.

philderbeast:


garyb:
The examples have their own `.gitignore`s already that cover the build products (for example: https://github.com/slamdata/purescript-halogen/blob/f3da4c7885fda2dab7db3e43547d21901d662da7/examples/ace/.gitignore).

Regarding vim stuff - I think that belongs in your global `.gitignore` rather than having it at a project level, personally.

garyb:
Ah, I guess `node_modules` is missing from those though. I don't encounter that usually as if you've already `npm install`ed in the project root the individual examples don't need it doing also.

philderbeast:
Thanks for the tip on globbing on vim globally. So as it stands, do you expect that if I execute the command line to build the multicomponent example, then I should not see any build outputs as pending changes in git?

Issue #333, opened 2016-11-04
Have multi-component example use the module names


philderbeast:
I suggest a change to the multi-component example to use the module names and not have globally unique names for things. So instead of ...

```
import ComponentA (StateA, QueryA(..), SlotA(..), initStateA, componentA)
import ComponentB (StateB, QueryB(..), SlotB(..), initStateB, componentB)
import ComponentC (StateC, QueryC(..), SlotC(..), initStateC, componentC)
```

I'd like to change the example to ...

```
import ComponentA as A
import ComponentB as B
import ComponentC as C
```

Then each Component_ module would be like ...

```
module ComponentC (State, Query(..), Slot(..), initState, component) where
```

garyb:
The components are named differently purposefully actually, to make it as clear as possible the components have distinct types. It would probably be better if the example was changed to make that more apparent by having obviously different components in the UI too, as the implementation would be more obviously different then.

I'll update the example to include more obvious differences after #315 is merged.

philderbeast:
I tried modifying this example to have differently typed components but the type errors were quite hairy.

> The components are named differently purposefully actually, to make it as clear as possible the components have distinct types.

I'd like to see two multi-component examples. This one pretty much as-is but with the module name change I've suggested and a mixed-multi-component example.

I've made the change in naming already locally. It reads more clearly to my eye. I'll send a PR for your review.

This clearly shows that a module name is a scope and within that scope we can use simpler names without concern for name clashes in other modules.

We could explain in the readme for this example that the components are intentionally simple and contrived to be the same internally as this example is about combining components, ie. what is in Main.purs.

The other multi-component example would then be more advanced showing how to resolve the types when the components have different effects.

PR #334, opened 2016-11-05
Module names for multi-component example (#333).

philderbeast:


garyb:
I'm going to close this, as the examples has been reworked a bit now, as part of #389. It does now qualify the naming of the child components. Thanks though!

PR #340, opened 2016-11-14
Drop EventHandler

### LINKED ISSUE #339: Delete `EventHandler`?
Strictly speaking we don't need it, as we could capture the event and pass it back to `eval` via query, and then `preventDefault` or whatever there instead. This would also make event handling for elements exactly the same as message handling for child components.

The downside is it's less convenient when one does want to `preventDefault` or whatever, as the effect has to appear in `m` in some form - whether that's `Eff` / `Aff` or something algebraic interpreted later on.

Thoughts? /cc @natefaubion, @cryogenian, @beckyconning, @jdegoes

garyb:
Resolves #339, for discussion.


garyb:
@natefaubion using DOM events brought more simplifications! No need for existentials with the handlers now (although, they probably weren't needed before either). It does coerce `Event`s to their subtypes, but that's not really any different than what we had before. It could be done "properly" using `Foreign`, but as long as we haven't messed up anywhere with the HTML DSL this shouldn't cause any trouble.

PR #341, opened 2016-11-17
Root-level message listeners

garyb:


src/Halogen/Aff/Driver.purs
```diff
@@ -43,12 +44,17 @@ import Halogen.Query.HalogenM (HalogenM(..), HalogenF(..), HalogenAp(..))
...
--- | Type alias for driver functions generated by `runUI` - a driver takes an
--- | input of the query algebra (`f`) and returns an `Aff` that returns when
--- | query has been fulfilled.
-type Driver f eff = f ~> Aff (HalogenEffects eff)
-
-type DSL s f m eff p o = HalogenF s f m p o (Aff (HalogenEffects eff))
+-- | A record produced when the root component in a Halogen UI has been run.
+-- | `query` allows external sources to query the root component and `subscribe`
+-- | allows external consumers to receive messages raised by the root component.
+type HalogenIO f o m =
```
natefaubion: This type is not Aff specific. Should it be defined elsewhere?

src/Halogen/Aff/Driver.purs
```diff
@@ -88,28 +94,52 @@ runUI
...
+  subscribe
+    :: Ref Int
+    -> Ref (M.Map Int (AV.AVar (Either o Unit)))
+    -> ({ producer :: CR.Producer o (Aff (HalogenEffects eff)) Unit
+        , unsubscribe :: Aff (HalogenEffects eff) Boolean
+        } -> Aff (HalogenEffects eff) Unit)
+    -> Aff (HalogenEffects eff) Unit
+  subscribe fresh ref k = do
+    listenerId <- liftEff $ readRef fresh
+    liftEff $ modifyRef fresh (_ + 1)
```
natefaubion: You could reduce the liftEffs:
```purescript
listenerId <- liftEff (readRef fresh <* modifyRef fresh (plus 1))
```

This is just a matter of taste though :P

examples/counter/src/Main.purs
```diff
@@ -41,8 +41,8 @@ ui = H.component { render, eval, initialState }
 main :: Eff (H.HalogenEffects ()) Unit
 main = runHalogenAff do
   body <- awaitBody
-  driver <- runUI ui body
-  setInterval 1000 $ driver (H.action Tick)
+  app <- runUI ui body
+  setInterval 1000 $ app.query (H.action Tick)
```
natefaubion: We could probably use an example that listens to events as well.

src/Halogen/Aff/Driver.purs
garyb: I pondered that too. Maybe just in the main `Halogen` module?

examples/counter/src/Main.purs
garyb: Yeah, I was going to add that separately, will do so while working on the docs.

src/Halogen/Aff/Driver.purs
garyb: How about
``` purescript
listenerId <- liftEff do
  lid <- readRef fresh
  modifyRef fresh (_ + 1)
  pure lid
```
We get "magic do" optimisations that way? Or I could just leave it 😛

src/Halogen/Aff/Driver.purs
natefaubion: That won't cause circular dependency issues?

src/Halogen/Aff/Driver.purs
natefaubion: Not a sticking point :)

examples/counter/src/Main.purs
natefaubion: I just meant to show that it does indeed work, before merging :D

natefaubion: Like, I don't care if its even to this one. This could be the "IO" example.

garyb: Ah ok, I did check that! https://gist.github.com/garyb/e6aacc4019a1781d846a2a82e203c41b

src/Halogen/Aff/Driver.purs
garyb: No, the driver stuff isn't re-exported from `Halogen`, and `HalogenIO`'s types are not driver spectific either.

PR #342, opened 2016-11-18
Better subscription handling

garyb:
This works and everything, but because a process doesn't end immediately when a consumer closes (it waits for one more emit from the producer or the producer to end) this means the detach handlers might not run at all.

I have an idea that might work for a bit of a hack, if I can come up with a way of interleaving (rather than fusing) producers, but not having much luck with that so far.

### DIFF
--- a/src/Halogen/Aff/Driver.purs
+++ b/src/Halogen/Aff/Driver.purs
@@ -20,7 +20,6 @@ import Control.Monad.Eff.Ref (Ref, modifyRef, writeRef, readRef, newRef)
 import Control.Monad.Error.Class (throwError)
 import Control.Monad.Free (foldFree)
 import Control.Monad.Fork (fork)
-import Control.Monad.Rec.Class (Step(..), tailRecM)
 import Control.Monad.Trans.Class (lift)
 import Control.Parallel (parSequence_, sequential, parallel)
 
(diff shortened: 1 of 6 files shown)


garyb:
Alright, this works, although it's not very pretty. 

- The consumer forks query evaluation so there should be no issues with queuing evals.
- With this new thing, the consumer does still hang around waiting for one more input before actually closing, but I'm using an `AVar` to block rather than the process now, so when a query eval returns false we write to the var to unblock, which then allows the detach handler to run.

I also made some changes to the `eventSource` stuff itself to better align it with how we do everything else: no more `Eff` in the callback - as with normal event handlers and the component messages it's now just an `input -> Maybe (Query Boolean)` now. It means we can _almost_ use `HE.input` again here, but not quite, as that's for actions and these are requests now. 

The above also means the `eventSource_` handlers ("no input" handlers) just have a static query they raise every time now, and there's no point in having the `Maybe` in there. Making some `eventSource_ attach Nothing` would be an rather inefficient no-op.

garyb:
This is the last one @natefaubion, honest!

src/Halogen/Aff/Driver.purs
```diff
@@ -154,8 +154,23 @@ runUI renderSpec component = do
...
     Subscribe es next -> do
-      let consumer = forever (lift <<< evalF ref =<< CR.await)
-      forkAff $ CR.runProcess (unwrap es $$ consumer)
+      forkAff do
+        { producer, done } <- ES.unEventSource es
+        isDone <- liftEff $ newRef false
+        blocker <- AV.makeVar
+        let
+          consumer :: CR.Consumer (f' Boolean) (Aff (HalogenEffects eff)) Unit
+          consumer = flip tailRecM unit \_ -> do
```
natefaubion: `tailRecM` is unnecessary in Co/FreeT since it already requires a `MonadRec` constraint to interpret it. This is just adding a multiplier of binds.

natefaubion:
> With this new thing, the consumer does still hang around waiting for one more input before actually closing, but I'm using an AVar to block rather than the process now, so when a query eval returns false we write to the var to unblock, which then allows the detach handler to run.

Would this be fixed by an alternative fuse?

natefaubion:
The issue now is that the normal coroutine fusion is unbiased. It runs the producer and consumer in parallel. I think if you had a consumer biased fusion operator, you could alleviate this.

garyb:
> Would this be fixed by an alternative fuse?

Possibly, but I burnt about an hour and a half on it and had more than enough, so went with this instead. I think that method would have been equally unpleasant to what we have here anyway!

garyb:
> I think that method would have been equally unpleasant to what we have here anyway!

Actually I was talking about something else there. Biased consumer would be the most pleasant fix.

PR #343, opened 2016-11-18
Revise AJAX example

### LINKED ISSUE #139: Provide HTTP example which works
The provided HTTP example doesn't work without disabling CORS. This can be very disconcerting for new users as immediately it looks as if Halogen's HTTP capacity doesn't work.

garyb:
Resolves #139

### DIFF
--- a/examples/ajax/src/Main.purs
+++ b/examples/ajax/src/Main.purs
@@ -2,14 +2,9 @@ module Main where
 
 import Prelude
 
-import Control.Alt ((<|>))
 import Control.Monad.Aff (Aff)
 import Control.Monad.Eff (Eff)
-import Control.Monad.Except (runExcept)
 
-import Data.Either (Either(..))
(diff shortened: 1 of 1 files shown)


examples/ajax/src/Main.purs
```diff
@@ -19,91 +14,76 @@ import Halogen.HTML.Properties as HP
...
+            [ HP.value st.username
+            , HE.onValueInput (HE.input SetUsername)
+            ]
+        ]
+    , HH.button
+        [ HP.disabled st.busy
+        , HE.onClick (HE.input_ MakeRequest)
+        ]
+        [ HH.text "Fetch info" ]
+    , HH.p_
```
garyb: Yeah, I'd probably do it a bit differently for a real thing, but I want to keep everything super simple that is not related to what the example is trying to demonstrate - I'd probably `foldMap` or `guard` the things in a real situation.

PR #360, opened 2017-01-05
Update missing link URLs in GUIDE.md

noraesae:
This fixes missing Pursuit links in `GUIDE.md`.

Just before uploading this PR, I found #355 is ongoing. This patch seems to be obsolete after the new guide is published, but it may help users to read the guide before the new one is coming. It is completely okay not to accept this if it looks redundant.

By the way, thanks for the great work.


dackerman:
I just want to throw in my support for this PR. I started learning purescript & halogen a couple days ago and have been pretty frustrated with the Guide's links being broken. I made this exact same PR myself until I realized this one existed.

I strongly feel that it's worth merging this before the new guide (although I very much look forward to seeing the new one!) even if it was < a week out from shipping.

garyb:
Ok, I'll merge this then - but it's worth noting that the guide is out of date in more ways than the links, it is no longer suitable as a reference for the actual code that is in master.

PR #370, opened 2017-01-24
Bake InputF into IndexedProp

natefaubion:
This takes the weird `InputF` rekinding, but bakes it into `IndexedProp`. This completely hides `InputF` from the user. This lets us put things of kind `*` into input handlers, and lets us keep simple `map` over `HTML` to change the input type.


src/Halogen/HTML/Properties.purs
```diff
@@ -3,7 +3,7 @@
 -- | standard unrefined versions.
 module Halogen.HTML.Properties
   ( IndexedProp(..)
-  , IProp
+  , IndexedProp
```
garyb: :eyes:(..)
:eyes:

examples/ace/src/AceComponent.purs
```diff
@@ -52,17 +52,17 @@ aceComponent =
     { editor: Nothing }
 
   -- As we're embedding a 3rd party component we only need to create a
-  -- placeholder div here and attach the ref property which will raise a query
-  -- when the element is created.
+  -- placeholder div here and attach the ref property which will let us reference
+  -- the element in eval.
```
garyb: Good catch!

src/Halogen/HTML/Properties.purs
garyb: Maybe we should keep using the `IProp` name too, and remove `IndexedProp` entirely again - it will break just a few less things when we update to this version of Halogen :smile:

It'll also fix the tests.

natefaubion:
* Changed back to `IProp`
* Fixed test

PR #373, opened 2017-01-30
Get rid of crazy nonsense in ChildQuery

garyb:
:tada:


garyb:
Will use this instead if it's approved: https://github.com/purescript/purescript-free/pull/69

garyb:
Now using the normal `Coyoneda` rather than a custom `ChildQuery` type.

PR #380, opened 2017-02-04
Make children initializers run in parallel

garyb:
(Also don't blow the stack).

### DIFF
--- a/src/Halogen/Aff/Driver/Eval.purs
+++ b/src/Halogen/Aff/Driver/Eval.purs
@@ -14,7 +14,7 @@ import Control.Monad.Error.Class (throwError)
 import Control.Monad.Fork (fork)
 import Control.Monad.Free (foldFree)
 import Control.Monad.Trans.Class (lift)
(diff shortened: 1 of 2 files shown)


src/Halogen/Aff/Driver/Eval.purs
```diff
@@ -48,9 +47,7 @@ handleLifecycle lchs f = do
   result <- liftEff f
   { initializers, finalizers } <- liftEff $ readRef lchs
   forkAll finalizers
-  -- No need to par/fork initializers here as there's only ever zero or one at
-  -- this point, due to the squashing at each level of the component hierarchy.
-  sequence_ initializers
+  parSequence_ initializers
```
natefaubion: Why is the comment no longer valid (and you want the initializers to be parallel)?

garyb: The comment was only true due to the sequence-squashing that occurred before. Here is where our 900 initializers run in the example. So yeah, we want them to run in parallel here.

PR #381, opened 2017-02-04
Performance improvements for child component init

garyb:
- Avoids blowing the stack even when initializing 9990 child components at once by not relying on `ParAff`
- Removed now-unnecessary handling of `pendingRefs`, since they don’t play into lifecycle anymore
- Made `handlePending` work in `Eff` since it didn’t really need to be `Aff`

### DIFF
--- a/src/Halogen/Aff/Driver/Eval.purs
+++ b/src/Halogen/Aff/Driver/Eval.purs
@@ -4,7 +4,7 @@ import Prelude
 
 import Control.Applicative.Free (hoistFreeAp, retractFreeAp)
 import Control.Coroutine as CR
-import Control.Monad.Aff (Aff, forkAff, forkAll)
+import Control.Monad.Aff (Aff, attempt, forkAff, forkAll)
 import Control.Monad.Aff.Unsafe (unsafeCoerceAff)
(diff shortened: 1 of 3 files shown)


garyb:
Maybe todo: reverse+count list in `parSequenceAff_`, see what needs doing about error handling.

src/Halogen/Aff/Driver/Eval.purs
```diff
@@ -37,6 +39,25 @@ type LifecycleHandlers eff =
...
 
+parSequenceAff_
+  :: forall eff a
+   . L.List (Aff (avar :: AV.AVAR, ref :: REF | eff) a)
+  -> Aff (avar :: AV.AVAR, ref :: REF | eff) Unit
+parSequenceAff_ as = do
+  let n = L.length as
+  when (n > 0) do
+    var <- AV.makeVar
+    ref <- liftEff $ newRef n
```
natefaubion: Since we are specialized to Lists, we can get away without the `length`, and just put the actual List in the Ref, and pattern matching on the spine.

garyb:
Made the `List`-related changes now, and added error propagation.

PR #389, opened 2017-02-09
Components guide chapter & examples

garyb:


garyb:
The last part of chapter 5 is all that is outstanding here, which will cover multi-child-type parent components. There's a revised example for it included though.

examples/basic/package.json
```diff
@@ -3,11 +3,8 @@
   "scripts": {
     "build": "pulp browserify --optimise --to dist/example.js"
   },
-  "dependencies": {
-    "virtual-dom": "^2.1.1"
-  },
   "devDependencies": {
     "pulp": "^10.0.0",
-    "purescript": "^0.10.2"
+    "purescript": "^0.10.5"
```
natefaubion: Do these need to be bumped to 0.10.7?

examples/components-inputs/dist/index.html
```diff
@@ -0,0 +1,16 @@
+<!doctype html>
+<html xmlns="http://www.w3.org/1999/xhtml" xmlns:foo="http://www.w3.org/2000/svg">
+  <head>
```
natefaubion: Why the xmlns:foo?

docs/5 - Parent and child components.md
```diff
@@ -1,3 +1,473 @@
...
+    HandleButton (Button.Toggled _) next -> do
+      H.modify (\st -> st { toggleCount = st.toggleCount + 1 })
+      pure next
+    CheckButtonState next -> do
+      buttonState <- H.query ButtonSlot $ H.request Button.IsOn
+      H.modify (_ { buttonState = buttonState })
+      pure next
+```
+
+A runnable version of this is available in the [`components` example](../examples/components/).
```
natefaubion: I think it's better to start with the example, but it should probably mention that the main difference is that it's using `parentComponent`, but otherwise the spec as the same arguments.

examples/components-inputs/dist/index.html
garyb: o_O

examples/basic/package.json
garyb: Yeah I guess that makes sense. Also all of the examples aside from `ace` don't need to use `pulp browserify` either, `build` with `--to` suffices, so I'll update those too.

garyb:
I'm done messing around with this now, think everything should be good to go!

PR #394, opened 2017-02-14
Prevent lifecycle handlers being clobbered by receiver-raised rendering

garyb:


garyb:
@natefaubion the `Ref` stuff in the `VDom` driver is a little unfortunate, but unless we can do something to allow passing of an extra argument to the `Step`s (like, you can do that of course, but not when going through `buildVDomSpec` as it fixes the `Step`'s `a` type to be `VDom`) then I think it'll have to be this.

Took me a while to figure out that the VDom machinery was caching the function for child rendering :cry:

PR #396, opened 2017-02-16
[Speculative] Use FreeT for HalogenM

garyb:


garyb:
Closing this for now, will continue experimenting with this as it does seem like a more correct way to do things, but it'll be for a future version now.

PR #400, opened 2017-02-20
use `GetState` for H.get

themoritz:
If I don't want to trigger a re-render of my component when querying its state, `MonadState`'s `get` shouldn't be used, because its implemented in terms of `ModifyState`.

This defines `get` and `gets` in terms of `GetState` and re-exports them instead.

I think users should also be warned somewhere not to use `get` of the `MonadState` instance if performance is important to them.

Ideally, we could define the `MonadState` instances using a separate `get` and `modify` like in Haskell, but I guess implementing `get`+`modify` or `state` is not possible in PureScript?

### DIFF
--- a/src/Halogen/Query.purs
+++ b/src/Halogen/Query.purs
@@ -31,12 +31,12 @@ import DOM.HTML.Types (HTMLElement, readHTMLElement)
 
 import Halogen.Component.ChildPath (ChildPath, injSlot, prjSlot, injQuery, cpI)
 import Halogen.Query.EventSource (EventSource, SubscribeStatus(..), eventSource, eventSource_)
-import Halogen.Query.HalogenM (HalogenM(..), HalogenF(..), getRef, getSlots, checkSlot, mkQuery)
+import Halogen.Query.HalogenM (HalogenM(..), HalogenF(..), getRef, getSlots, checkSlot, get, gets, mkQuery)
 
 import Control.Parallel (parTraverse)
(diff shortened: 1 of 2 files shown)


natefaubion:
Another option might be to do a referential check before re rendering. I prefer to try and make something like that unnecessary, but it's a decent workaround that lets us keep the instance.

garyb:
@themoritz thanks for opening this, I meant to check this out after a problem we recently had with initializers where doing a `get` seemed to be triggering a render.

> @themoritz: Ideally, we could define the MonadState instances using a separate get and modify like in Haskell, but I guess implementing `get`+`modify` or `state` is not possible in PureScript?

It's not (at the moment), as we can't specify default member implementations in classes - there are quite a few things that would potentially benefit from being able to do this, but for now it's still on the todo list.

> @natefaubion: Another option might be to do a referential check before re rendering. I prefer to try and make something like that unnecessary, but it's a decent workaround that lets us keep the instance.

I like that. Well, "like", but I think it's a decent solution for now.

garyb:
I'll look into this and the other issues you opened a bit later on today.

themoritz:
I agree, a referential check would also be nice, and I guess in JS doing this is safe compared to doing similar things in Haskell? Then the `GetState` could also be dropped, since it's not used anywhere.

natefaubion:
It's safe regardless. Because `get` is polymorphic in the state type, it can't do anything but yield the same reference.

PR #401, opened 2017-02-20
Fix render on state get with unsafeRefEq

natefaubion:


themoritz:
Awesome! I'll close the other PR.

Issue #402, opened 2017-02-21
halogen-vdom adds additional element when diffing


themoritz:
This is likely a vdom issue but I'm still posting it here. It's really hard to reproduce this in a small example since the setting is so complex, but here is basically what happens:

There is an element somewhere deep in the render function of one of my components which gets copied on every render. However, it's part of a popup menu, so a descendant of some other `div` which sometimes gets removed from the DOM (i.e., when the menu is closed). Whenever I close and open the popup again, everything renders correctly, so I assume during the diffing process of the element something goes wrong.

There is nothing special about the element, it's just a `div` with a `class` attribute, has one `div` child, and has another `div` as its parent. There's also another popup menu similar like this, but which is not affected by the bug (it doesn't have the `.cell-list__element-delete` element below).

Here is the markup, if that helps debugging:

```html
<div class="cell-expanded">    <----- this is the root of the popup
  <div class="cell-expanded__body p1">
    <div class="cell-list">
      <div class="cell-list__element flex mb1">
        <div class="cell-list__element-delete p1">
          <button class="cell-button"><i class="fa fa-minus-circle"></i></button>
        </div>
        <div class="flex-auto p1">
          <div class="right-align full-height">    <----- this div gets duplicated
            <div class="editbox ">0</div>
          </div>
        </div>
      </div>
      <div class="cell-list__new p1">
        <button class="cell-button"><i class="fa fa-plus-circle"></i></button>
      </div>
    </div>
  </div>
</div>
```

garyb:
Nathan is having trouble with GitHub at the moment, but he's pretty baffled by what is going on here.

We suspect it may be similar to the problem that the `onBlur` issue raised though, where `render`s are overlapping - once there's a fix in for that (I'll work on it now), perhaps you could try again then and see if it still manifests?

themoritz:
Yeah, it seems crazy and there shouldn't be much room for these kind of bugs since everything is so generic.

> perhaps you could try again then and see if it still manifests?

Sure thing!

themoritz:
The problem still persists on your `render-queuing` branch @garyb.

garyb:
Ok thanks, darn.

natefaubion:
[...]

PR #403, opened 2017-02-21
Render queuing

### LINKED ISSUE #398: DOMException when using input onBlur events
The following example throws `Uncaught DOMException: Failed to execute 'removeChild' on 'Node': The node to be removed is no longer a child of this node. Perhaps it was moved in a 'blur' event handler?` when pressing `Enter` on the `input` element (only in Chrome, not in Firefox, and not when pressing "Stop Edit" or deselecting the `input` element):

```purescript
module Main where

import Prelude
import Halogen as H
import Halogen.HTML as HH
import Control.Monad.Aff (Aff)
import Control.Monad.Eff (Eff)
import DOM.Event.KeyboardEvent (code)
import Data.Maybe (Maybe(..))
import Halogen.Aff (awaitBody, runHalogenAff)
import Halogen.Aff.Effects (HalogenEffects)
import Halogen.VDom.Driver (runUI)

data Query a
  = StartEdit a
  | SetValue String a
  | StopEdit a

type State =
  { editing :: Boolean
  , value :: String
  }

app :: H.Component HH.HTML Query Unit Void App
app = H.component
  { initialState: const
      { editing: false
      , value: ""
      }
  , render
  , eval
  , receiver: const Nothing
  }

render :: State -> H.ComponentHTML Query
render st = HH.div_
  [ case st.editing of
      true ->
        HH.input
        [ HP.value st.value
        , HE.onValueInput (HE.input SetValue)
        , HE.onKeyDown \e -> case code e of
            "Enter" -> Just (H.action StopEdit)
            _       -> Nothing
        , HE.onBlur (HE.input_ StopEdit)
        ]
      false ->
        HH.text st.value
  , HH.button
    [ HE.onClick $ HE.input_ (if st.editing then StopEdit else StartEdit) ]
    [ HH.text (if st.editing then "Stop Edit" else "Start Edit") ]
  ]

eval :: Query ~> H.ComponentDSL State Query Void App
eval = case _ of
  StartEdit next -> do
    H.modify _{ editing = true }
    pure next
  SetValue val next -> do
    H.modify _{ value = val }
    pure next
  StopEdit next -> do
    H.modify _{ editing = false }
    pure next

type AppEffects = HalogenEffects ()
type App = Aff AppEffects
    
main :: Eff AppEffects Unit
main = runHalogenAff $ do
  body <- awaitBody
  runUI app unit body
```

I did some research and it seems this is somehow because removing an element from the DOM fires an `onblur` event on it (see [this SO answer](http://stackoverflow.com/a/22934552/3176849)). Also the [MDN node.removeChild() doc](https://developer.mozilla.org/en-US/docs/Web/API/Node/removeChild) says:

> If child is actually not a child of the element node, the method throws an exception. This will also happen if child was in fact a child of element at the time of the call, but was removed by an event handler invoked in the course of trying to remove the element (eg, blur.)

Is this something that can be taken care of in `halogen-vdom`?

garyb:
Should fix #398

### DIFF
--- a/src/Halogen/Aff/Driver.purs
+++ b/src/Halogen/Aff/Driver.purs
@@ -136,7 +136,7 @@ runUI renderSpec component i = do
     -> Eff (HalogenEffects eff) Unit
   render lchs var = readRef var >>= \(DriverState ds) -> do
     shouldProcessHandlers <- isNothing <$> readRef ds.pendingHandlers
-    unless shouldProcessHandlers $ writeRef ds.pendingHandlers (Just L.Nil)
+    when shouldProcessHandlers $ writeRef ds.pendingHandlers (Just L.Nil)
     writeRef ds.childrenOut M.empty
     writeRef ds.childrenIn ds.children
     let
--- a/src/Halogen/Aff/Driver.purs
+++ b/src/Halogen/Aff/Driver.purs
@@ -14,6 +14,7 @@ import Control.Monad.Eff (Eff)
 import Control.Monad.Eff.Class (liftEff)
 import Control.Monad.Eff.Exception (error, throw, throwException)
 import Control.Monad.Eff.Ref (Ref, modifyRef, writeRef, readRef, newRef)
+import Control.Monad.Rec.Class (Step(..), tailRecM)
 
 import Data.List ((:))
 import Data.List as L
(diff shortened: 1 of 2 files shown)


themoritz:
Yep, this fixes #398 for me :tada:

garyb:
How about #402? :smile:

themoritz:
Negative :cry:

garyb:
@natefaubion how does this version look?

themoritz:
@garyb commit 7a92a7d reintroduces the onblur problem.

garyb:
@themoritz can you try this again on your #402 case? I think we might have it this time.

garyb:
:smile:

themoritz:
Ok, 1a19db3 fixes it again :joy:

garyb:
not

PR #411, opened 2017-02-26
Point people to the v0.12 tag in the README

dackerman:
For people just coming to Purescript, the examples alone aren't very helpful to understanding how things work. I added a link to the latest release tag so that people can see the guide and examples from that period. I don't know when 1.0 is planning to ship, but I think this should help newcomers until then.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,8 @@ A declarative, type-safe UI library for PureScript.
 
 :warning: The guide and template project are currently being updated for the new release. There are plenty of [examples](examples/) to dig into though.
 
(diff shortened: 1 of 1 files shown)


garyb:
I intend to release the version as non-RC on Monday, so I don't think there's much advantage at this point :smile:

We unlinked the v0.12 docs for the opposite reason to this too, it was confusing people who were trying to use master. Next time there's a major breaking change it will live on a separate branch.

dackerman:
Alright, I'll close this PR then. Early congrats on the upcoming release, and I look forward to reading the latest docs! :champagne:

PR #419, opened 2017-03-04
Add some stuff for file selection

jacereda:


jacereda:
So, @natefaubion enlightened me in https://github.com/slamdata/purescript-dom-indexed/pull/3 but it seems I would still need an `accept` property for a file requester, right?

natefaubion:
@jacereda Yes, we still need an `accept` prop.

jacereda:
Ok, I'll close this mess and send a separate PR for `accept`

PR #424, opened 2017-03-06
Add accept property

jacereda:


garyb:
I'm not sure this is quite the right type for the property, from the small amount of googling I just did it suggests there might be something about filenames as well as media types that go/can go in here?

garyb:
Ignore that, seems like `MediaType` probably is sufficient - there are other ways of specifying the acceptable types, but they seem less good than `MediaType`, and this is what we have in `dom-indexed` right now, so let's go with it :smile:

Thanks!

PR #429, opened 2017-03-18
Add Profunctor component instance

natefaubion:


src/Halogen/Component/Profunctor.purs
```diff
@@ -0,0 +1,53 @@
...
+import Prelude
+import Control.Applicative.Free (hoistFreeAp)
+import Control.Monad.Free (hoistFree)
+import Data.Newtype (class Newtype, over)
+import Data.Profunctor (class Profunctor, lmap)
+import Halogen.Component as HC
+import Halogen.Query.HalogenM as HM
+import Halogen.Query.ForkF as FF
+
+newtype ProComponent h f m i o = ProComponent (HC.Component h f i o m)
```
cryogenian: Can this be defined on `Component` not `newtype`?

natefaubion: Only if we change the order of parameters. That's why this newtype exists.

natefaubion:
It's not possible to implement `Choice` and `Strong` without closing the `f` parameter.

PR #430, opened 2017-03-19
Fix link rendering issue

safareli:


safareli:
There is some issue when github is rendering links like this:
```
[text](link to file with space.md)
```

If we use that snippet in comment it is not rendered as desired: [text](link to file with space.md). But if you escape spaces it renders properly: [text](link%20to%20file%20with%20space.md). Curently in [2 - Defining a component.md](https://github.com/slamdata/purescript-halogen/blob/master/docs/2%20-%20Defining%20a%20component.md) links like that render properly for example this once in last paragraph:
```
[run to produce a UI](4 - Running a component.md "Running a component")
[do something effectful](3 - Handling effects.md "Handling effects")
```

But if you take a look at the same file on my branch [safareli-link](https://github.com/slamdata/purescript-halogen/blob/safareli-link/docs/2%20-%20Defining%20a%20component.md) (which has no changes) you will see that this links are not rendered as desired.

Also if you try to [edit `2 - Defining a component.md`](https://github.com/slamdata/purescript-halogen/edit/master/docs/2%20-%20Defining%20a%20component.md) on master, and take a look at preview tab, all links like that, are not rendered as desired any more.

(First I thought that adding title will fix this but in preview it was not fixed, then I thought maybe it should be committed and preview had some bug, but as of this commit, it still is not rendered properly so adding/removing title will change nothing).


So what I think is happening is that the links rendered as desired before (when first committed) and the result was cached by github. Then github changed markdown renderer and contents of `2 - Defining a component.md` from master is still from cache, but cache for example for `3 - ...md` is invalidated. after cache for `2 - ...md` gets invalidated, we will obser links not rendering properly here too. 

So looks like we need to escape (change ` ` to `%20`) such links to fix this issue.

What you think on the issue/solution?

safareli:
update in [2 - Defining a component.md](https://github.com/slamdata/purescript-halogen/blob/master/docs/2%20-%20Defining%20a%20component.md)  links are not renderd properly any more looks like the cache was cleared.


so now to solve this issue links containing spaces need to be escaped

garyb:
Switching to %20 makes sense to me - hopefully we can make the links reference-style like the rest of them too that way, as the only reason they were inline before is they didn't parse correctly as references.

Thanks for looking into this! I hadn't even noticed there was a problem.

PR #435, opened 2017-03-26
Update effects-aff-ajax example

stephencelis:
Using a form element makes interacting with the example a bit easier (hitting "enter" on the keyboard, for example).


garyb:
Thanks!

PR #438, opened 2017-03-30
adding Namespaced counterpart functions to element, keyed, and attr t…

toastal:
…o make it easier to create things like SVGs


src/Halogen/HTML/Core.purs
```diff
@@ -70,17 +70,18 @@ text :: forall p i. String -> HTML p i
 text = HTML <<< VDom.Text
 
 -- | A smart constructor for HTML elements.
-element :: forall p i. VDom.ElemName -> Array (Prop i) -> Array (HTML p i) -> HTML p i
-element = coe (\name props children -> VDom.Elem (VDom.ElemSpec Nothing name props) children)
+element :: forall p i. Maybe VDom.Namespace -> VDom.ElemName -> Array (Prop i) -> Array (HTML p i) -> HTML p i
+element ns =
+  coe (\name props children -> VDom.Elem (VDom.ElemSpec ns name props) children)
   where
```
natefaubion: @garyb Do we consider these things part of the public API? This adds an argument to the function.

toastal: For perspective, JavaScript has `createElement` and `createElementNS`... where `createElement` is `createElementNS(null, ...)`

natefaubion: Yeah, I consider this the correct signature. It's just a matter of whether we consider this a breaking change. If it is, we might just have to copy it to a new definition and define the old one in terms of this.

garyb: Yeah, I would say this is a breaking change. We could hold off releasing it properly until we do the 0.11 update and do them together then?

Also, minor nitpick, but I'd prefer we used `NS` in the names rather than `Ns` :smile:

toastal: No other comments about style?

garyb: Nope, the rest looks 👌 to me

natefaubion: @toastal needs this for his stuff, so we can either work from a branch, or we can just make it backwards compatible.

garyb: I would suggest using a branch for now. Unless you think it's a bad idea?

natefaubion: Nope, don't care.

natefaubion:
I've changed the base to point to `next`.

garyb:
You can do that 😮

natefaubion:
Yeah, that's a somewhat recent addition to GH.

PR #440, opened 2017-04-07
Updates for PureScript 0.11

NasalMusician:
I spent some time to update this for the new release, pretty excited to be able to use it with PS 0.11.

There are a few dependencies that we are waiting on, which I've listed in 934d7ce. I've adjusted my things locally to verify that my changes build and set pull requests for the non-trivial updates I had to make:
* [purescript-affjax](https://github.com/slamdata/purescript-affjax/pull/98)
* [purescript-ace](https://github.com/slamdata/purescript-ace/pull/25) – non-essential, dev dependency of this project

The only breaking change is selectElement now accepts an argument of type QuerySelector, to fit in with the other API changes. Otherwise the updates are internal, in response to API changes.

The few compile issues I had to work around are documented in 0ab5a06 and 30621af. I am not sure if these are compiler bugs or if there's better ways to either fix them or work around them, but what I have seems to work. Let me know if I can fix this better!

One last comment is that I worked from the next branch, which has changes I believe you were waiting to merge basically ... now? So we can merge it all here :)

Let me know if there is anything else I can do. Thank you!


src/Halogen/Aff/Driver/State.purs
```diff
@@ -160,7 +160,7 @@ initDriverState component input handler prjQuery lchs = do
       , pendingOuts
       , pendingHandlers
       , rendering: Nothing
-      , prjQuery
+      , prjQuery: prjQuery :: (forall a. f a -> Maybe (z a))
```
natefaubion: Alternatively, I think you could annotate `ds`, which would probably a better option.

NasalMusician:
Hm ... I think it is just choking on the test/[websocket dev dependency](https://github.com/zudov/purescript-websocket-simple), otherwise I believe the library itself works – thanks for updating the dependencies!

nwolverson:
On the second noted compiler issue, while the error is bogus, `dom :: DOM, dom :: DOM` correctly does not unify with `dom :: DOM` (duplicate labels are allowed but distinct). Your unsafeCoerceEff should IMO be typed explicitly to restrict what it is permitted to do e.g.
```
coerceDomDom :: forall eff a. Eff (dom :: DOM, dom :: DOM | eff) a -> Eff (dom :: DOM | eff) a 
```
And I guess this could be added somewhere common (there is no meaning of distinct `DOM` labels, unlike some other cases). The change which causes the issue is this change in purescript-dom @garyb 

https://github.com/purescript-contrib/purescript-dom/commit/e4412c59b03d96a67466a403bf2ca8ed98ab6d6f#diff-990ec12feb5a00188dd90d8032997a24

garyb:
@nwolverson yeah :cry: I thought making that change was safe - it seems like it should be with the duplicate row restriction having been lifted.

src/Halogen/HTML/Events.purs
```diff
@@ -211,25 +206,27 @@ clipboardHandler = unsafeCoerce
...
 -- | Attaches event handler to event `key` with getting `prop` field as an
 -- | argument of `handler`.
-addForeignPropHandler :: forall r i value. IsForeign value => EventType -> String -> (value -> Maybe i) -> IProp r i
-addForeignPropHandler key prop f =
-  handler key (either (const Nothing) f <<< runExcept <<< readProp prop <<< toForeign <<< EE.currentTarget)
+addForeignPropHandler :: forall r i value. EventType -> String -> (Foreign -> F value) -> (value -> Maybe i) -> IProp r i
+addForeignPropHandler key prop reader f =
+  handler key (either (const Nothing) (maybeRead >=> f) <<< runExcept <<< readProp prop <<< toForeign <<< EE.currentTarget)
+  where
+    maybeRead = reader >>> runExcept >>> either (const Nothing) Just
```
garyb: This change can be much simpler:

``` purescript
addForeignPropHandler :: forall r i value. (Foreign -> F value) -> EventType -> String -> (value -> Maybe i) -> IProp r i
addForeignPropHandler read key prop f =
  handler key (either (const Nothing) f <<< runExcept <<< (read <=< readProp prop) <<< toForeign <<< EE.currentTarget)
```

garyb: Oh, that version has the `read` in a different place than yours though :)

src/Halogen/Aff/Util.purs
```diff
@@ -6,27 +6,22 @@ module Halogen.Aff.Util
...
+import DOM.HTML.Event.EventTypes (load)
 import DOM.HTML.Types (HTMLElement, windowToEventTarget, htmlDocumentToParentNode, readHTMLElement)
 import DOM.HTML.Window (document)
-import DOM.Node.ParentNode (querySelector)
-
+import DOM.Node.ParentNode (QuerySelector(..), querySelector)
+import Data.Either (either)
+import Data.Foreign (toForeign)
+import Data.Maybe (Maybe(..), maybe)
 import Halogen.Aff.Effects (HalogenEffects)
```
garyb: Can you sort the imports alphabetically again please?

src/Halogen/HTML/Events.purs
```diff
@@ -48,24 +48,19 @@ module Halogen.HTML.Events
...
+import Data.Either (either)
+import Data.Foreign (Foreign, F, toForeign, readString, readInt, readBoolean)
+import Data.Foreign.Index (readProp)
+import Data.Maybe (Maybe(..))
 import Halogen.HTML.Core (Prop)
 import Halogen.HTML.Properties (IProp)
 import Halogen.Query (Action, action)
 import Halogen.Query.InputF (InputF(..))
-
 import Unsafe.Coerce (unsafeCoerce)
```
garyb: Again the import sorting - `psc-ide` sorts them in a different way to our preferred style.

garyb:
I'll do something for the websockets example to make it so we don't need the dependency on websockets-simple anymore... just need to add it to `-dom` first!

src/Halogen/Aff/Util.purs
NasalMusician: Okay, thanks for the sharp eyes. Out of curiosity, are the differences between your style and psc-ide's documented somewhere? Is it mostly just string comparison (upper-case first) versus alphabetical ordering?

Personally I would sort imports from like most fundamental to most high-level, but I bet that's too subjective and complex to be practical. It would still suggest that I need to fix this though. ;)

src/Halogen/HTML/Events.purs
NasalMusician: Perfect, that makes sense. Would you like me to also make an addStringPropHandler etc. now? (= addForeignPropHandler readString, in your version).

garyb: I think it's fine to just apply `readString`, etc. where they're used - we only have these 4 cases of it anyway. I don't mind whether the `read` part of it is positioned at the start or the end either, I just happened to write it that way when checking the types work out for my version!

src/Halogen/Aff/Util.purs
garyb: Our style is purely "sort the non-Prelude lines alphabetically" with no semantic understanding of the imports at all - it's just easier to maintain consistency that way, as it means you don't have to be using `psc-ide` to figure out the correct ordering... I got Christoph to explain the psc-ide ordering to me this week and I've probably forgotten it already 😜, but I think it does:

1. Open imports first
2. Qualified `as` imports next
3. Explicit imports last

It's something like that anyway. I don't know how that works when you have qualified-and-explicit imports either.

NasalMusician:
Okay, a few more updates, thanks for the comments! I also submitted a merge request for websocket-simple but not sure when that may be updated...

PR #443, opened 2017-04-16
Updates for PureScript 0.11

rahulmutt:
Mostly trivial changes:
- Bumping dependencies
- Replacing `*` kind with `Type`
- Replacing `!` kind with `Effect`
- Explicity adding `_ <- ...` in monadic expressions that discard their result.
- In `HalogenEffects`, renaming `err` to `exception` (due to breaking change in `prescript-exceptions`)
- Replacing tupled constraints with curried `=>`
- Accounting for `Nullable` -> `Maybe` change in APIs in `purescript-dom`.
- Fixing a name-shadowing error.

The non-trivial change:
It seems `Data.Foreign.Class` was transferred from `purescript-foreign` to `purescript-foreign-generic` and the function `readProp` was remove so I inlined it here and replaced:
- `IsForeign` -> `Decode`
- `read` -> `decode`
It should most be harmless and carry the same meaning as before, but I'd like someone more experienced to verify that the change is valid.


garyb:
Thanks, but there's already a PR underway for this - #440

rahulmutt:
Sorry - missed that!

PR #447, opened 2017-04-20
Add  as an exposed HTML property

AidanDelaney:
purescript-dom allows `list` as a property of <input>.  I've added it to the properties that halogen exposes.


garyb:
Thanks!

PR #448, opened 2017-04-21
Prevent `done` from running multiple times for subscriptions

### LINKED ISSUE #418: The `done` part of an `EventSource` is run more often than expected
I was trying to figure out Halogen's `EventSource` mechanism and ran into an issue which might be a bug (I'm not sure).

TL;DR: The `done` part of an `EventSource` is run more often than I would expect.

To illustrate the issue I've created a repo [here](http://www.github.com/menelaos/halogen-issue). Just run `pulp browserify > app.js` and e.g. `python -m http.server`.

Basically, I have a `Parent` component that can switch between two child views. To keep it simple, `ChildB` only displays text. `ChildA` is a `lifecycleComponent` which subscribes to a `busEventSource` upon initialization. The bus is part of the reader environment (`Wiring`) and emits `Unit` values. Whenever the bus fires, a `Toggle` query is raised in `ChildA` which toggles its `On/Off` state. After 10 seconds, the bus receives a `Bus.write` in the `main` function. The `busEventSource` function is taken from SlamData's `Halogen.Component.Utils` module. I've modified the `done` function in `affEventSource` in order to log a random number to the console whenever it is run.

Here's how to reproduce the issue:
* Open the console
* Load the page
* Within the first 10 seconds, switch between `ChildA` and `ChildB` a few times.

You will notice that each time `ChildA` is un-mounted, the `done` function is run as expected. After 10 seconds, however, all `done` functions are run again. I.e. if `ChildA` has been mounted/unmounted `n` times, `n` `done` functions will be run after 10 seconds in addition to those that have been run immediately upon unmounting.

Note that this does not affect the `Toggle` query in `ChildA` which is correctly dispatched only once after 10 seconds (if the component is mounted, of course).

Is this behaviour intended?
Thanks for your time!

garyb:
Resolves #418

### DIFF
--- a/src/Halogen/Aff/Driver/Eval.purs
+++ b/src/Halogen/Aff/Driver/Eval.purs
@@ -22,7 +22,7 @@ import Data.Either (Either(..))
 import Data.List (List, (:))
 import Data.List as L
 import Data.Map as M
-import Data.Maybe (Maybe(..), isJust)
+import Data.Maybe (Maybe(..), isJust, maybe)
 import Data.StrMap as SM
 import Data.Tuple (Tuple(..))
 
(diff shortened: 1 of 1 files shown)


src/Halogen/Aff/Driver/Eval.purs
```diff
@@ -114,10 +114,15 @@ eval render r =
...
-        i <- liftEff do
-          i <- modifyRef' fresh (\i -> { state: i + 1, value: i })
-          modifyRef subscriptions (map (M.insert i done))
-          pure i
+        i <- liftEff $ modifyRef' fresh (\i -> { state: i + 1, value: i })
+        let
+          done' = do
+            subs <- liftEff $ readRef subscriptions
+            when (maybe false (M.member i) subs) do
+              pure unit
```
kRITZCREEK: this seems odd?

garyb: Quick! Look over there! *waves hands*

What seems odd? :wink:

garyb:
@natefaubion unfortunately this does not fix whatever is going on with the "leaking" subscription in SlamData. I'm still trying to figure out what's happening in that case, as I can't see how Halogen would let it happen.

PR #464, opened 2017-05-31
Fix 'effects-aff-ajax' example

jfraudeau:
Bump dependencies for this example as was done on other examples
Solves a problem with the behavior of form elements on Firefox : The submission of the form refreshes the page and thus resets everything on the page.

The solution works but may be too complex considering how early it appears in the guide, another solution would be to change the form element back to a div.

### DIFF
--- a/examples/effects-aff-ajax/bower.json
+++ b/examples/effects-aff-ajax/bower.json
@@ -2,7 +2,7 @@
   "name": "effects-aff-ajax",
   "private": true,
   "dependencies": {
-    "purescript-affjax": "^3.0.0",
+    "purescript-affjax": "^4.0.0",
     "purescript-halogen": "*"
   }
 }
--- a/examples/effects-aff-ajax/package.json
+++ b/examples/effects-aff-ajax/package.json
@@ -4,7 +4,7 @@
     "build": "pulp build --to dist/example.js"
   },
   "devDependencies": {
-    "pulp": "^10.0.0",
-    "purescript": "^0.10.7"
+    "pulp": "^11.0.0",
+    "purescript": "^0.11.4"
   }
 }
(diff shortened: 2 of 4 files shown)


garyb:
I'm still thinking about how best to do this one... I'd be tempted to remove the `onSubmit` and go back to only having the button only, as that avoids a lot of the "irrelevant" stuff here (in regards to what the example is supposed to be illustrating). 

On the other hand, it'd be good to show off the pattern for  `preventDefault` handling where a query like `PreventDefault Event (Query a)` is used (unlike this fixed version :wink:).

jfraudeau:
I'm unsure what you mean with `PreventDefault Event (Query a)`, is it about making the query algebra recursive and still have a special case in the handle function ? Or something totally different ?
I just saw that preventDefault is also used in the keyboard example

garyb:
Yeah that's the idea - by separating it out into another query algebra constructor it means it can be reused where appropriate, and by having the constructor carry `Query a` again it forces it to be used in `eval`, since the only possible way of satisfying that is by calling `eval` again on the inner query.

It's one of the things I want to add a bit about in the guide as it occurs reasonably often.

jfraudeau:
I updated the code following your suggestions, still don't know if it's a good place to introduce this pattern.

jfraudeau:
@garyb Is this still up to date ? I'd like to see this PR closed if it is not to be merged.

cryogenian:
This is fixed by https://github.com/slamdata/purescript-halogen/pull/532/files#diff-fc7253792c155768107a26c7a5bdb9e0R52, closing. 

Basically if `onClick` behaviour on form is incorrect due to default submit it's easy to add `HP.type HP.ButtonButton`. 

@jfraudeau Thank you

PR #466, opened 2017-06-06
remove of the obsolete unsubscribe field

grumpi:
If I understand things correctly, unsubscription is now triggered by returning `H.Done` from the keyboard handler's query, which is a much nicer solution than having to keep track of the subscription manually.

Removing the `unsubscribe` field from the example's state doesn't seem to change the example's functionality. That's why I removed it.

### DIFF
--- a/examples/keyboard-input/src/Main.purs
+++ b/examples/keyboard-input/src/Main.purs
@@ -23,17 +23,17 @@ import Halogen.VDom.Driver (runUI)
 
 import Keyboard as K
 
-type State eff = { chars :: String, unsubscribe :: Maybe (Aff (Effects eff) Unit) }
+type State = { chars :: String }
 
-initialState :: forall eff. State eff
-initialState = { chars : "", unsubscribe: Nothing }
+initialState :: State
+initialState = { chars : "" }
 
 data Query a
   = Init a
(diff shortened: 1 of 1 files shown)


grumpi:
Question: Should there be an example of how to manage a subscription like this manually, i.e. in order to be able to unsubscribe in arbitrary queries, not just from the `HandleKey` query?

garyb:
Thanks! And yes, we should have an example for that too. Someone is asking that very question on StackOverflow, but I've not had time to put something together yet. :)

grumpi:
tbh, the StackOverflow question is what made me look into this because it sounded like an interesting thing for a beginner to figure out and, if successful, explain. Knowing how, in general, to explicitly subscribe and unsubscribe to EventSources from inside components sure seems useful in a bunch of cases.

From looking at the Halogen sources and making wild guesses, I get the impression that maybe all that's needed to manually unsubscribe is to solve the puzzle of how to get at the `done` function of the `EventSource`.

natefaubion:
> From looking at the Halogen sources and making wild guesses, I get the impression that maybe all that's needed to manually unsubscribe is to solve the puzzle of how to get at the done function of the EventSource.

You would have to change the signature of `HalogenF.Subscribe` to yield the `done` effect instead of `Unit`.

PR #474, opened 2017-09-12
Add onWheel event

kwohlfahrt:
Adds `onWheel` event (like `onClick`, but for mouse wheels). Seems to be in `DOM.Event.Types` already, but not anywhere else.

Seems simple and works with my use-case, but not really tested. slamdata/purescript-dom-indexed#10 is required for use of the tag to type-check.

Reference: https://developer.mozilla.org/en-US/docs/Web/Events/wheel

### DIFF
--- a/src/Halogen/HTML/Events.purs
+++ b/src/Halogen/HTML/Events.purs
@@ -26,6 +26,7 @@ module Halogen.HTML.Events
   , onMouseOver
   , onMouseOut
   , onMouseUp
+  , onWheel
   , onKeyDown
   , onKeyPress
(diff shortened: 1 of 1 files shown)


kwohlfahrt:
I've signed the CLA, but is there a reason it requires access to *private* email addresses? The public one really should be sufficient.

garyb:
Thanks, I'll merge this when we make the release that updates for new `Aff`, as both are breaking changes and the `Aff` one shouldn't be long in coming.

I have no idea why the CLA thing wants a private email 😕

PR #482, opened 2017-09-30
Update for Aff v4

### LINKED ISSUE #481: Needs Aff 4.0 upgrade
It seems like at least the change of Control.Monad.Fork -> Control.Monad.Fork.Class needs to be done first, and then there are some Aff-related changes that need to be done that I don't know how to do.

garyb:
Resolves #481


src/Halogen/Aff/Util.purs
```diff
@@ -7,33 +7,31 @@ module Halogen.Aff.Util
...
-    >>= windowToEventTarget
-    >>> addEventListener load (eventListener (\_ -> callback unit)) false
+awaitLoad = makeAff \callback -> liftEff do
+  et <- windowToEventTarget <$> window
+  let listener = eventListener (\_ -> callback (Right unit))
+  addEventListener load listener false et
+  pure $ Canceler \_ -> liftEff (removeEventListener load listener false et)
 
 -- | Waits for the document to load and then finds the `body` element.
 awaitBody :: forall eff. Aff (dom :: DOM | eff) HTMLElement
```
natefaubion: We should fix `awaitBody` so that it doesn't block forever if the document has already loaded.

garyb: Ah yeah 👍

spicydonuts: I took a stab at this here: https://github.com/garyb/purescript-halogen/pull/5

i-am-the-slime: What's holding this back?

PR #489, opened 2017-10-27
upgrade purescript-aff-coroutines to v6.0.0

jimmyhuco:
As purescript-halogen has upgraded to v3.0.0 which requires
purescript-aff v4.0.0, but purescript-aff-coroutines v5.0.0 requires
purescript-aff v3.0.0.


garyb:
Thanks!

PR #491, opened 2017-10-29
Load on DOMContentLoaded instead of full window load

spicydonuts:
Mounts the app as soon as the DOM exists rather than when all initially requested resources have finished loading (css, js, images, etc).


garyb:
Thanks!

PR #500, opened 2018-01-12
IProp is a Functor

MonoidMusician:
Just a simple instance that I've found useful

### DIFF
--- a/src/Halogen/HTML/Properties.purs
+++ b/src/Halogen/HTML/Properties.purs
@@ -95,6 +95,7 @@ import Unsafe.Coerce (unsafeCoerce)
 newtype IProp (r :: # Type) i = IProp (Prop (InputF Unit i))
 
 derive instance newtypeIProp :: Newtype (IProp r i) _
+derive instance functorIProp :: Functor (IProp r)
 
 -- | Creates an indexed HTML property.
 prop


src/Halogen/HTML/Properties.purs
```diff
@@ -95,6 +95,7 @@ import Unsafe.Coerce (unsafeCoerce)
 newtype IProp (r :: # Type) i = IProp (Prop (InputF Unit i))
 
 derive instance newtypeIProp :: Newtype (IProp r i) _
+derive instance functorIProp :: Functor (IProp r)
```
garyb: You could `derive newtype instance` here even I think?

MonoidMusician: I don’t think so, `IProp r` is more like `Compose Prop (InputF Unit)`, but that’s not the newtype we have.

MonoidMusician: In particular the implementation of it would use `map` twice: https://github.com/MonoidMusician/purescript-halogen-behaviors/blob/master/src/Halogen/Behavior/Internal.purs#L21 (from my project)

MonoidMusician:
Thanks Gary!!

garyb:
Thanks for the PR :smile:

I'll release this later today, once some other non-breaking bits and pieces are also in.

PR #501, opened 2018-01-12
Provide functions for mapping the input and output of a Component

MonoidMusician:
I think these are useful, maybe not very widely – perhaps a component's input has extra data that doesn't change often (think like a render function), this would allow exposing that as a function argument producing a component with a simpler input, instead of just a component with a complex input. Unfortunately the type arguments of `Component` are in the wrong order for a `Profunctor`...

### DIFF
--- a/src/Halogen/Component.purs
+++ b/src/Halogen/Component.purs
@@ -264,6 +264,42 @@ hoist nat =
       , mkOrdBox: c.mkOrdBox
       }
 
+-- | Changes the component's `i` type, contravariantly.
+cmapInput
+  :: forall h f i i' o m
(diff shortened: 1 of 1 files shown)


cryogenian:
https://pursuit.purescript.org/packages/purescript-halogen/3.0.1/docs/Halogen.Component.Profunctor#t:ProComponent

cryogenian:
And this https://github.com/slamdata/purescript-halogen/blob/master/src/Halogen/Component/Profunctor.purs (:

MonoidMusician:
Ah!! Thanks, I missed that. 😁

PR #507, opened 2018-01-18
remove `bimap id id`

safareli:
`bimap id id` is same as `id` and can be removed, i guess it's leftover or something


safareli:
I guess Bifunctor instance was checked :d

garyb:
Haha true! I think you're right that it was leftover from something that used to require it.

PR #508, opened 2018-01-20
Fix list property/attribute

fehrenbach:
This makes using an input with a `list` referring to a datalist element work instead of throwing a type error due to setting a read-only property.

Unlike most properties, the list property (on an HTMLInputElement) is read-only. To set the underlying attirbute programmatically we need to use `setAttribute` instead of just setting the property. This is done for `attr`s but not `prop`s.

Not sure this is the correct fix, but it seems to work.
Thanks for the great work on Halogen!


garyb:
Thanks! I eventually tracked down the IDL and you're absolutely right that it's defined as `readonly` (I did believe you :wink: just wondered if it was specc'd that way, or if this was a browser implementation thing).

Issue #483, opened 2017-10-04
Monoid instance for ClassName


safareli:
When you have dynamic classes we have couple options to encode them:

### `<>` and `guard`

``` purs
[ ...
, HP.classes $
  [ HH.ClassName "baz"
  , HH.ClassName "foo"
  ] 
  <> (guard isBar $> HH.ClassName "bar")
  <> (guard isFiz $> HH.ClassName "fiz")
]
```

### `fold` and `guard`
``` purs
[ ...
, HP.classes $ fold
    [ [ HH.ClassName "baz"
      , HH.ClassName "foo"
      ]
    , guard isBar $> HH.ClassName "bar"
    , guard isFiz $> HH.ClassName "fiz"
    ]
]
```

### `STArray`

``` purs 
unArr do
  push $ HH.ClassName "baz"
  push $ HH.ClassName "foo"
  guard isChecked $> HH.ClassName "bar"
where
unArr = unsafeFreeze >>> runST >>> runPure
push = pushSTArray
```

### `mguard` - needs Monoid instance

``` purs
[ HH.ClassName "baz"
, HH.ClassName "foo"
, mguard isChecked $ HH.ClassName "bar"
]
where
mguard b m = if b then m else mempty
```

I think last one looks better, also it should be better in terms of performance.

So as we already have Semigroup (#451), what if we added Monoid instance to ClassName?

only drawback might be having extra spaces in dom, here is definition of classes:
`classes = prop (PropName "className") <<< joinWith " " <<< map unwrap`

safareli:
If spaces become an issue we can update `classes` function to just fold over array and use spaces only when element is not empty string ([it might even be faster then `join`](https://stackoverflow.com/a/4717855/1798418))

spicydonuts:
Btw, that mguard exists as Data.Monoid.guard now, but a new release hasn’t been tagged yet

safareli:
@spicydonuts didn't knew about that, Thanks! updated description

cryogenian:
How about this? 
```purescript
className :: String -> Maybe ClassName 
className = Just <<< ClassName 

classes = prop (PropName "className") <<< joinWith " " <<< catMaybes <<< map unwrap 

[ HH.className "baz"
, HH.className"foo"
, guard isChecked *> HH.className"bar"
]
```

I'm not totally sure about semantics of monoid's `mempty` but `pure mempty === empty === mempty` in this case seems a bit dodgy, doesn't it?

[...]

PR #512, opened 2018-01-31
Semigroup instance on ClassName now joins with a space

machinedgod:
This produces a valid class string for the HTML 'class' attribute.

### DIFF
--- a/src/Halogen/HTML/Core.purs
+++ b/src/Halogen/HTML/Core.purs
@@ -190,5 +190,8 @@ newtype ClassName = ClassName String
 derive instance newtypeClassName :: Newtype ClassName _
 derive newtype instance eqClassName :: Eq ClassName
 derive newtype instance ordClassName :: Ord ClassName
-derive newtype instance semigroupClassName :: Semigroup ClassName
+--derive newtype instance semigroupClassName :: Semigroup ClassName
 derive instance genericClassName :: Generic ClassName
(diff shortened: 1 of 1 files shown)


safareli:
Why do you need `<>` to use space as glue? For the behavior you are proposing you can use Array.

Point of Semigroup instance was to "namespace" classnames like:

```
app = ClassName "sd-"
app <> ClassName "footer"
```

justinwoo:
I think what @machinedgod wants is to append together a bunch of separate classnames, which should be done with [intercalate](https://pursuit.purescript.org/packages/purescript-foldable-traversable/3.4.0/docs/Data.Foldable#v:intercalate) instead e.g. `intercalate " " [a,b,c,d]`

safareli:
you would be doing:
`HH.div  [HP.classes [H.ClassName "foo", H.ClassName "bar"]]`

ClassName is just a newtype over String so `classes` take Array of ClassName instead of Array of String.

btw you can take a look at https://github.com/slamdata/purescript-halogen/issues/483 it's also related

machinedgod:
Excellent, I wasn't aware of 'classes' property. That does much better.
I guess it was just intuitive for me to use 'append' to append classes together. Maybe the namespacing could be expanded upon, and have a newtype for a namespace with a separate function that applies namespace to every ClassName?

PR #516, opened 2018-02-12
Fix shadowed import warnings

garyb:


safareli:
Shouldn't build have failed in case of warnings?

garyb:
These warnings lack position information in 0.11, so they don't get picked up by `psa --strict`. They're fixed in 0.12.

PR #517, opened 2018-02-13
Rework lifecycle hooks

### LINKED ISSUE #515: Use a separate library algebra for lifecycle methods
Rather than having multiple configuration options to map to a query, maybe we should just have something like:

```purescript
component = H.component { render, eval, lifecycle }
  where
  lifecycle :: H.Lifecycle Input -> HalogenM ... Unit
  lifecycle = case _ of
    H.Initialize -> ...
    H.Receive input-> ...
    H.Finalize -> ...
```

This would make it easier to write component proxies since you don't have to parameterize the query algebra with child input types.

garyb:
Resolves #515


natefaubion:
It wasn't readily apparent, but in #515 I suggested that `lifecycle` be in `HalogenM` rather than returning a query. What do others think of this? The reason I suggested it is because lifecycles are the only form that potentially allow an outsider to get direct access to a constructed `Query`. Otherwise it would possible to make a Query completely opaque, save for the type.

natefaubion:
It also means that HOCs can decorate a components lifecycles.

garyb:
Ah I see. I'll have to take another look then, that's a perhaps a bigger change.

I'm not totally sure if it fixes the problem either, as `HalogenM` still references the query type for a component?

natefaubion:
Hmm that's true. I guess one can still get a hold of queries in a round-about way using EventSource. Maybe EventSource should be in HalogenM 😆. I think it's possible, but then it goes down the route of why aren't HTML events in HalogenM, etc. We wanted a single entry point for eval rather than baking it into callbacks.

Slightly off topic, but if we are going to make breaking changes, are there other things we'd like to do. Slots with rows, etc? Maybe we should make an issue for planning that.

safareli:
Slots with rows 😍

garyb:
https://github.com/slamdata/purescript-halogen/issues?q=is%3Aopen+is%3Aissue+milestone%3Anext

garyb:
Closing this for now since this particular approach isn't actually going to help with the linked issue :wink:

PR #521, opened 2018-03-11
Update package versions for ajax examples.

jeremyjh:
I couldn't build the example without these changes. It looks like only the Ajax example had not yet been updated to purescript 0.11.4.

### DIFF
--- a/examples/ace/package.json
+++ b/examples/ace/package.json
@@ -5,6 +5,6 @@
   },
   "devDependencies": {
     "pulp": "^11.0.0",
-    "purescript": "^0.11.4"
+    "purescript": "^0.11.7"
   }
 }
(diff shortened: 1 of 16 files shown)


justinwoo:
Just a minor thing, probably could be upgraded to 0.11.7?

jeremyjh:
@justinwoo Every other example is at 11.4. Should I just bump them all to 11.?

safareli:
> @justinwoo Every other example is at 11.4. Should I just bump them all to 11.7?

@jeremyjh it would be nice if you can do that!

jeremyjh:
I've updated the rest of examples too.

garyb:
Thanks!

PR #522, opened 2018-03-31
New subscribe/unsubscribe/EventSource API

### LINKED ISSUE #479: Is there a way to cancel an event source subscription from "outside"?
I am trying to implement throttling (send HTTP request after few keystrokes, instead of immediately) with `EventSource`. I succesfully implmeneted some delay with `setTimeout` but I realized that there doesn't seem to be a way to cancel ticking `setTimeout` clock from "outside".

You can do that only form inside query cases in `eval` which will be called by subscribed event source, but this would be equivalent to being able to cancel `setTimeout` only from within passed callback:

```purescript
 -- parts of eval function:
    InputChange newInput next -> do
     -- when InputChange is called again I would like to cancel previous subscription, but I can't?
      H.subscribe $ HES.eventSource_' (\eff -> do
        id <- setTimeout 1000 eff
        pure $ void $ clearTimeout id) (H.request FetchOptions)
      H.modify \state -> state { input = newInput, options = Nothing }
      pure next
    FetchOptions status -> do
      input <- H.gets \state -> state.input
      when (input /= "") do
        getOptions <- H.gets \state -> state.getOptions
        options <- H.lift $ getOptions input
        H.modify \state -> state { options = Just options }
      pure $ status H.Done
```
It seems to me that `subscribe` should return some kind of mechanism for cancelling subscription.

garyb:
Resolves #479 

This will:

- [x] allow arbitrary unsubscriptions (currently can only be done in response to an event)
- [x] simplify the API for event sources
- [x] ensure that queries that evaluate asynchronously will not block the subscription loop


src/Halogen/Query/EventSource.purs
```diff
@@ -1,187 +1,150 @@
...
+  -> EventSource m f
+affEventSource recv = EventSource $ liftAff do
+  inputVar <- AV.makeEmptyVar
+  finalizeVar <- AV.makeEmptyVar
+  let
+    producer = do
+      lift $ liftAff $ flip AV.putVar finalizeVar =<< recv (Emitter (flip AV.putVar inputVar))
+      CR.producer $ liftAff $ AV.takeVar inputVar
+    finalizer = Finalizer do
+      liftAff (attempt (AV.takeVar finalizeVar)) >>= case _ of
```
safareli: Why attempt is needed?

garyb: In case the finalizer is run a second time - it kills the var, so attempting to `take` from it would result in an `Aff` error.

garyb:
Hmm, maybe... I was looking forward to getting rid of that `fresh` state :smile:

What use case are you envisioning for that? Fire-and-forget subscriptions? As otherwise they'll need saving in the component state anyway, in which case I'd be tempted to leave the onus on the component to generate names rather than it being a capability of Halogen in general.

natefaubion:
I suppose since subscriptions implies MonadAff anyway, there's no reason you can't generate unique names yourself with `random`.

src/Halogen/Query/EventSource.purs
```diff
@@ -1,186 +1,172 @@
...
+eventListenerEventSource
+  :: forall m f
+   . MonadAff m
+  => E.EventType
+  -> ET.EventTarget
+  -> (E.Event -> Maybe (f Unit))
+  -> EventSource m f
+eventListenerEventSource eventType target f = effEventSource \emitter -> do
+  listener <- ET.eventListener (maybe (pure unit) (emit emitter <<< pure) <<< f)
+  ET.addEventListener eventType listener false target
```
safareli: Have you considered to also take `capture :: Boolean` as argument?

src/Halogen/Query/EventSource.purs
```diff
@@ -1,186 +1,172 @@
...
+-- |
+-- | There is a `Monoid` instance provided for finalizers, so `mempty` can be
+-- | used in cases where there are no relevant clean-up actions to take.
+newtype Finalizer m = Finalizer (m Unit)
+
+instance semigroupFinalizer :: Apply m => Semigroup (Finalizer m) where
+  append (Finalizer a) (Finalizer b) = Finalizer (a *> b)
+
+instance monoidFinalizer :: Applicative m => Monoid (Finalizer m) where
+  mempty = Finalizer (pure unit)
```
safareli: I think `mempty = Finalizer mempty` should work too 😜

src/Halogen/Query/HalogenM.purs
```diff
@@ -123,10 +130,20 @@ checkSlot p = HalogenM $ liftF $ CheckSlot p identity
...
+subscribe :: forall s f g p o m. ES.EventSource m f -> HalogenM s f g p o m SubscriptionId
+subscribe es = HalogenM $ liftF $ Subscribe (\_ -> es) identity
+
+-- | An alternative to `subscribe`, intended for subscriptions that unsubscribe
+-- | themselves. Instead of returning the `SubscriptionId` from `subscribe'`, it
+-- | is passed into an `EventSource` constructor. This allows emitted queries
+-- | to include the `SubscriptionId`, rather than storing it in the state of the
+-- | component.
+subscribe' :: forall s f g p o m. (SubscriptionId -> ES.EventSource m f) -> HalogenM s f g p o m Unit
+subscribe' esc = HalogenM $ liftF $ Subscribe esc (const unit)
```
safareli: As I understand `subscribe'` is same as `void $ subscribe` and if it's correct, that why do we need this?
it's easy to add `void $ ` or `_ <-` in front of `subscribe`

cryogenian:
@garyb Would you mind pushing it into #532?

garyb:
@cryogenian I'll merge this, and then merge master into that :smile: - I wanted to make reviewing the other part clearer.

src/Halogen/Query/EventSource.purs
garyb: Not really - I just included this as a convenience for the common case. It's easy to implement with `effEventSource` anyway, so I figured there's that if people require it.

src/Halogen/Query/HalogenM.purs
garyb: It's not the same as `subscribe`:
``` purescript
subscribe :: forall s f g p o m. ES.EventSource m f -> HalogenM s f g p o m SubscriptionId
subscribe' :: forall s f g p o m. (SubscriptionId -> ES.EventSource m f) -> HalogenM s f g p o m Unit
```

safareli: Just looked at the result 🤦‍♂️

src/Halogen/Query/EventSource.purs
garyb: Can you think of a case where using `capture` is common? I'd be tempted to add it as a separate named function if we could think of a situation in which listening in the capture phase was the norm.

safareli: It's useful when you are listening to some event on body for example and might want to cancel propagation of event, or just want to change/set something before handler of some element deep in the dom is fired. but usually it's not that common.

More real example would be if you are embedding some lib in the dom and the lib on clickdoes something but also cancels event propagation. if you want to know if click happened, you could set capture to true on wrapper node and your listener will be fired before the lib get's it

PR #523, opened 2018-04-08
awaitLoad canceler removes correct eventListener

acple:
I found a little mistake in halogen aff, `awaitLoad` Aff actions cancellation wouldn't remove eventListener correctly.


garyb:
Thanks!

PR #525, opened 2018-04-24
Use rows for child slots

### LINKED ISSUE #439: Use `RowCons` for child components
With `RowCons`, it's possible to define a mapping for child components without all the coproduct/either cruft. This should be more efficient and ergonomic because child queries and state can be dispatched directly without any wrapping.

```purescript
data At ix (g ∷ Type → Type)

type ChildSlots =
  ( navBar ∷ At Unit Nav.Query
  , item   ∷ At Item.Id Item.Query
  )

data HalogenM s (f ∷ Type → Type) (c ∷ # Type) (m ∷ Type → Type) a = ...

query
  ∷ ∀ sym s f g ix r1 r2 m a
  . RowCons sym (At ix g) r1 r2
  ⇒ IsSymbol sym
  ⇒ Ord ix
  ⇒ SProxy sym
  → ix
  → g a
  → HalogenM s f r1 m (Maybe a)
```
Internally, we could build a record based on this set of rows, with individual `Map`s for each component type.

This is great because we can define the Slot -> Query mapping together in a single definition. We had originally wanted to do this with TC magic and type-level assoc lists, but it was quite hideous.

natefaubion:
Fixes #439 

This is still a WIP but is usable and ready for testing/review of the approach. This also fixes the issue where you can't rebind the output handler once the component is mounted. It should just work like normal handlers do.

I think we should remove the `Component/Parent` distinction for HTML and DSL now. The only difference now will be that leaf components/html just have `()` as their slot type. We can also provide `absurdSlot :: forall a b. ComponentSlot () a -> b`

- [x] Add back queryAll support
- [x] Remove Component/Parent distinction
- [x] Fix all examples
- [x] Fix documentation comments
- [x] Remove ChildPath/Prism


natefaubion:
I wonder if it's worth baking `ComponentSlot` into `HTML`.

thomashoneyman:
👍 on removing the parent / component distinction and type synonyms. This is much clearer IMO.

natefaubion:
Also, `LifecycleComponentSpec` is just `Component'` now, and `lifecycleComponent` is just `mkComponent`. Not sure how we want to handle that now.

natefaubion:
If we can find a good solution to the lifecycle situation, we can just have the one constructor. That would be ideal.

garyb:
If nothing else, we could change the API so it's like #517 - it wasn't what you meant with the proposal for it, but doing would reduce the lifecycle handling aspects to one function at least.

natefaubion:
I think for now I'm going to remove the distinction and just have `component`, and we can address lifecycle stuff in another PR.

natefaubion:
I think most of the example diffs are due to windows line endings :rage4:

natefaubion:
One thing that's neat about the rows approach is you can export a default slot constructor from your component module.

From the `components-inputs` example, `Display` could export:
```purescript
display :: forall p f m r. Ord p => p -> Input -> H.ComponentHTML f (display :: Slot p | r) m
display slot i = HH.slot (SProxy :: SProxy "display") slot component i absurd
```
Which can be used in the `Container` just as:
```purescript
  render :: State -> H.ComponentHTML Query ChildSlots m
  render state =
    HH.div_
      [ HH.ul_
          [ Display.display 1 state
          , Display.display 2 (state * 2)
          , Display.display 3 (state * 3)
          , Display.display 4 (state * 10)
          , Display.display 5 (state * state)
          ]
  ...
```

natefaubion:
Bikeshed opportunity: order of type arguments.

paulyoung:
This is amazing 👏

acple:
(re-)rendering child components always output "Duplicate slot address" warning. cond inverted?

natefaubion:
> (re-)rendering child components always output "Duplicate slot address" warning. cond inverted?

Yes, you're right. Thanks!

paulyoung:
Is this likely to land any time soon?

PR #532, opened 2018-05-28
Consolidated prs

cryogenian:
#522 and #525 on top of v4.00-rc1 

Both #522 and #525 were approved I think. But there is a small fix I made in `onValueInput` I'll mark the change with comment.


src/Halogen/HTML/Events.purs
```diff
@@ -260,7 +260,11 @@ touchHandler = unsafeCoerce
 -- | argument of `handler`.
 addForeignPropHandler :: forall r i value. EventType -> String -> (Foreign -> F value) -> (value -> Maybe i) -> IProp r i
 addForeignPropHandler key prop reader f =
```
cryogenian: Here is it. Frankly it looks like 4.00 is silently broken due to `unsafeToForeign` :confused:

thomashoneyman:
Does this have any chance of landing in the `4.0` release or a soon-after `4.1` release?

PR #537, opened 2018-06-08
Lifecycle algebra in `eval`, existential internal/"public" external queries

### LINKED ISSUE #515: Use a separate library algebra for lifecycle methods

### LINKED ISSUE #526: Proposal: Existentially hidden query algebra
Currently the component query algebra is part of the public interface of a component. This makes it difficult to implement transparent HOCs because the types become infectious. I propose that we existentially hide a component's query algebra.

In #515 I proposed adding a separate library algebra for lifecycle events:

```purescript
component = H.component { render, eval, lifecycle }
  where
  lifecycle :: H.Lifecycle Input -> HalogenM ... Unit
  lifecycle = case _ of
    H.Initialize -> ...
    H.Receive input-> ...
    H.Finalize -> ...
```

I'm going to propose we actually just make this `eval` and add two more constructor so that it looks like this.

```purescript
data HalogenQ' f g i a
  = Initialize a
  | Finalize a
  | Receive i a
  | Internal (f a)
  | External (g (Maybe a))

data HalogenQ (g :: Type -> Type)  i a

eval :: forall f g i. HalogenQ' f g i ~> HalogenM s f ps o m
```

This takes the `Halogen.Component.Proxy` functionality and makes it first class. We separate a public query algebra from an internal query algebra which means type parameters are no longer infectious or invariant. We can even recover the existing component spec definitions in terms of this by making `f ~ g`.

Using this we can write an encapsulated HOC where no extra types leak.

```purescript
type QueryI o a
  = Toggle Boolean a
  | Raise o a

type State i =
  { visible :: Boolean
  , input :: i
  }

type ChildSlots g o =
  ( child :: H.Slot g o Unit
  )

toggleHoc :: forall g i o m. Component HH.HTML g i o m -> Component HH.HTML g i o m
toggleHoc inner = component { initialState, render, eval }
  where
  initialState :: i -> State i
  initialState i =
    { visible: true
    , input: i
    }

  render :: State i -> HH.ComponentHTML QueryI (ChildSlots g o) m
  render state =
    HH.div_
      [ HH.button
          [ HP.onClick (HE.input_ (Toggle (not state.visible))) ]
          [ HH.text "Toggle" ]
      , if state.visible
          then HH.slot (SProxy :: SProxy "child") unit inner state.input (HE.input Raise)
          else HH.text ""
      ]

  eval :: HalogenQ' (QueryI o) g i ~> HalogenM (State i) f (ChildSlots g o) m
  eval = case _ of
    Initialize next -> pure next
    Finalize next -> pure next
    Receive input next -> H.modify _ { input = input } $> next
    Internal (Toggle visible next) -> H.modify _ { visible = visible } $> next
    Input (Raise o next) -> H.raise o $> next
    External q -> H.query (SProxy :: SProxy "child") unit q
```

Note I used `External (g (Maybe a))` which means external queries can be partial. This is always the case anyway since a component may not be mounted, but this allows the inner component to control that partiality as well.

garyb:
Resolves #526, resolves #515


src/Halogen/Query/HalogenM.purs
```diff
@@ -34,37 +34,37 @@ newtype SubscriptionId = SubscriptionId Int
 derive newtype instance eqSubscriptionId :: Eq SubscriptionId
 derive newtype instance ordSubscriptionId :: Ord SubscriptionId
 
-newtype UnpackQuery ps g o f b =
-  UnpackQuery (forall slot m . Applicative m => (slot g o -> m b) -> SlotStorage ps slot -> m (f b))
+newtype UnpackQuery ps g i o f b =
```
kRITZCREEK: redundant

src/Halogen/Aff/Driver/Eval.purs
```diff
@@ -42,31 +44,39 @@ handleLifecycle lchs f = do
...
 
-eval
-  :: forall h r s f ps i o a
+evalF
+  :: forall h r s f act ps i o a
    . Renderer h r
-  -> Ref (DriverState h r s f ps i o)
-  -> InputF a (f a)
+  -> Ref (DriverState h r s f act ps i o)
+  -> InputF a (Coproduct f (Tuple act) a)
```
kRITZCREEK: `Either (f a) act`

src/Halogen/Query/HalogenM.purs
```diff
@@ -208,35 +210,35 @@ imapState f f' (HalogenM h) = HalogenM (hoistFree go h)
...
+  :: forall s act act' ps o m
    . Functor m
-  => (f ~> f')
-  -> HalogenM s f  ps o m
-  ~> HalogenM s f' ps o m
-mapQuery nat (HalogenM h) = HalogenM (hoistFree go h)
+  => (act -> act')
+  -> HalogenM' s act ps o m
+  ~> HalogenM' s act' ps o m
+mapMessage f (HalogenM h) = HalogenM (hoistFree go h)
```
natefaubion: Should this be `mapAction`?

src/Halogen/Query/HalogenQ.purs
```diff
@@ -0,0 +1,22 @@
...
+import Prelude
+
+import Data.Bifunctor (class Bifunctor)
+
+data HalogenQ f act i a
+  = Initialize a
+  | Finalize a
+  | Receive i a
+  | Handle act a
+  | Request (f a)
```
natefaubion: Are we not doing anything wrt partiality of requests?

src/Halogen/Query/HalogenM.purs
garyb: Yes, thanks!

src/Halogen/Query/HalogenQ.purs
garyb: I wasn't intending to - I'm not sure there's much of an advantage to it other than in the specific example case you came up with. It still needs to be dealt with somewhere, so I guess we'd have to halt with a generic message instead. At least the halt can be given a relevant message when we are forced to deal with it in the component.

Maybe I'm wrong though, and there is a better way to deal with the failed queries internally?

natefaubion: How does a `halt` propagate if you are doing a child query? If I'm in a parent, and I query a child which halts, will the parent query resume with `Nothing`, or will the entire logical threat blow up?

garyb: I see what you're saying now, I was thinking of this still causing a halt, just elsewhere - but actually it could behave much like a "failed" query does right now, in that any attempted child query might return `Nothing`.

garyb: This does sorta seem to push me into making all `eval` partial, like `eval :: forall a. HalogenQ f act i a -> HalogenM' s act ps o m (Maybe a)` which is a bit 🤔

natefaubion: That's why I added `RequestBox` in https://github.com/slamdata/purescript-halogen/issues/526#issuecomment-385271087 so that only `Request`s were potentially partial. An alternative would be to use `halt`, but I think that might be more strange, as you'd potentially be swallowing exceptions just because it happened during a request.

natefaubion: Maybe you could do `Request (f a) (Unit -> a)` instead which is equivalent to the Maybe construction.

garyb: I still haven't managed to figure out how to make this work, so I'm going to merge it as it is for now, to make a pre-release.

I see how it's supposed to work, I just haven't been able to rework eval to accommodate it. The closest I got was making `QueryBox` carry a `HalogenQ` query rather than `g`, but that falls apart due to existential `act` and `i` not working out between the `DriverState` ref and the `QueryBox`.

At least using this technique inside a HOC still works, at the expense of needing a wrapper query algebra, even if it's not baked into Halogen directly:

``` purescript
data PartialF f a = PartialF (f a) (Unit -> a)

eval = case _ of
    H.Request (PartialF q k) ->
      H.query _child unit q >>= case _ of
        Nothing -> pure (k unit)
        Just qa -> pure qa
  ...


join <$> (H.query _something unit $ HOC.PartialF (SomeQuery Just) (const Nothing))
```

natefaubion: I think that's probably OK, considering partial HOCs aren't likely to be the common case.

garyb:
Going to do another PR after this that cleans up some stuff, fixes comments, etc.

PR #545, opened 2018-06-26
Use ForkIds for forking

garyb:


src/Halogen/Query/HalogenM.purs
```diff
@@ -163,8 +163,16 @@ getRef p = HalogenM $ liftF $ GetRef p identity
...
 raise o = HalogenM $ liftF $ Raise o unit
 
-fork :: forall s act ps o m a. MonadAff m => HalogenM' s act ps o m a -> HalogenM' s act ps o m (Error -> m Unit)
-fork a = map liftAff <$> HalogenM (liftF $ Fork $ FF.fork a)
+newtype ForkId = ForkId Int
+
+derive newtype instance eqForkId :: Eq ForkId
+derive newtype instance ordForkId :: Ord ForkId
+
+fork :: forall s act ps o m. MonadAff m => HalogenM' s act ps o m Unit -> HalogenM' s act ps o m ForkId
```
safareli: Would be nice if you add a comment to this function #546

src/Halogen/Query/HalogenM.purs
```diff
@@ -163,8 +163,16 @@ getRef p = HalogenM $ liftF $ GetRef p identity
 raise :: forall s act ps o m. o -> HalogenM' s act ps o m Unit
 raise o = HalogenM $ liftF $ Raise o unit
 
-fork :: forall s act ps o m a. MonadAff m => HalogenM' s act ps o m a -> HalogenM' s act ps o m (Error -> m Unit)
-fork a = map liftAff <$> HalogenM (liftF $ Fork $ FF.fork a)
+newtype ForkId = ForkId Int
```
safareli: Should constructor be exported?

garyb: It's unavoidable, unfortunately - the driver creates/manages the IDs, so we can't make it private. It's the same thing with subscription IDs too.

safareli: if only operation we care about the ID is to create initial id and increment it to calculate next one, then we might be able to hide constructor and export `initial = ForkId 0`  `next :: ForkId -> ForkId`

src/Halogen/Query/HalogenM.purs
natefaubion: Can't you remove the `MonadAff` constraint now?

src/Halogen/Query/HalogenM.purs
```diff
@@ -163,8 +163,16 @@ getRef p = HalogenM $ liftF $ GetRef p identity
...
+
+derive newtype instance eqForkId :: Eq ForkId
+derive newtype instance ordForkId :: Ord ForkId
+
+fork :: forall s act ps o m. MonadAff m => HalogenM' s act ps o m Unit -> HalogenM' s act ps o m ForkId
+fork hmu = HalogenM $ liftF $ Fork hmu identity
+
+kill :: forall s act ps o m. MonadAff m => ForkId -> HalogenM' s act ps o m Unit
+kill fid = HalogenM $ liftF $ Kill fid unit
```
natefaubion: Same here. I don't see why `MonadAff` is necessary now.

src/Halogen/Query/HalogenM.purs
garyb: Actually, I'm not sure that it's much better - a user can mess things up with access to a `succ` / `next` function just as much as they can be constructing an id manually 😕.

At least the re-exports for it from `Halogen` etc. only export that types and omit the constructors.

safareli: Yes, that's also true  :(

test/Test/Component/ForkTest.purs
```diff
@@ -0,0 +1,90 @@
...
+    H.liftEffect $ Ref.modify_ (msg : _) logRef
+    pure Nothing
+
+  io.query (H.action StartFork)
+  Aff.delay (Aff.Milliseconds 350.0)
+  io.query (H.action KillFork)
+
+  -- TODO: revisit this: why do we need to wait to receive `raise`d messages
+  -- from the component, if the `raise` occurs after any bind?
+  Aff.delay (Aff.Milliseconds 15.0)
```
natefaubion: The Aff coroutines are forked and run in parallel separate from the main thread, so it may be just how the Aff runtime ends up interleaving things. If you want things to coordinate properly you should have a separate AVar that you take from when it gets the killed message.

garyb: I did consider that. Tried to :duck: it with Christoph, there is definitely something a bit weird happening though - if the `raise` is the first thing in the `Kill` eval it works without this delay, if you even put `pure unit` in front of it then the delay is required. Basically it's the difference between the `raise` happening before the `io.query` completes, or after.

There's a chance it's specific to the driver IO behaviour too, since that has to jump through some hoops to produce the `subscribe` function for `HalogenIO`. I'm not too worried about it right now though, really I just wanted to check that forking and killing was working properly here.

PR #550, opened 2018-06-28
`hoist` fusion

garyb:
@natefaubion I did all the obvious parts, but it becomes non-obvious where the hole is. Depending on how things are implemented we need `Aff` to be `m'` or vice versa to make things work out, but since `m'` has no constraints we can't actually do anything in `m'`, and if we want it to be `Aff` it appears we need a contravariant mapping too.

I'll let you scratch your head over it for a while 😜 I'm going to press on with fork-auto-clean-up and component disposal.

### DIFF
--- a/src/Halogen/Aff/Driver.purs
+++ b/src/Halogen/Aff/Driver.purs
@@ -31,6 +31,7 @@ import Halogen.Aff.Driver.State (DriverState(..), DriverStateRef(..), DriverStat
 import Halogen.Component (Component, ComponentSlot, unComponent, unComponentSlot)
 import Halogen.Data.Slot as Slot
 import Halogen.Query.EventSource as ES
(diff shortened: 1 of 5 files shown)


src/Halogen/Aff/Driver.purs
```diff
@@ -92,10 +93,10 @@ import Halogen.Query.Input (Input(..))
 -- | `const (pure unit)`.
 type RenderSpec h r =
   { render
-      :: forall s act ps o
+      :: forall s act ps o m
```
natefaubion: I had a go at this, and I tried adding the m here like this as well, but I think that's the problem. Rendering does not require it to be fixed to `Aff`, however everything in `VDom.Driver` uses `Aff` in type signatures. The `m` parameter needs to be carried around in the types of everything, rather than foralled, I think.

PR #554, opened 2018-07-04
Add Halogen.VDom.Thunk support

natefaubion:
This lets us skip rendering for arbitrary HTML trees, not necessarily for just components.


src/Halogen/HTML.purs
```diff
@@ -67,4 +70,53 @@ slot
...
+  -> ComponentHTML' act ps m
+memoized eqFn f a = Core.slot (ThunkSlot (thunked eqFn f a))
+
+-- | Skips rendering for referentially equal arguments. You should not use this
+-- | function fully saturated, but instead partially apply it for use within a
+-- | Component's scope.
+lazy
+  :: forall a act ps m
+   . (a -> ComponentHTML' act ps m)
+  -> a
```
safareli: don't you need Eq instance for it?

safareli: I thought the "referentially" was related to "referential transparency", what about `Skips rendering for arguments which are equal by reference`?

Issue #528, opened 2018-05-25
Bring guide to completion


linearray:
I'd really like to see the halogen guide completed, because it's by far the easiest way to learn about halogen.

I can put in some time next week to write about lifecycle methods, provided that somebody who really knows what's going on plays editor. I cannot be much help with event sources, but I think this chapter is badly needed as well.

Anyone willing to team up and commit to this?

garyb:
Sorry, I meant to reply to this days ago!

Once we get the next release out I'll definitely try and fill in the gaps - there's going to be a few changes (including event sources and subscriptions specifically, to hopefully make them more pleasant to use), so it's probably worth holding off till then.

JordanMartinez:
I read through the guide (which seems to exclude the recent additions of `initializer` and `finalizer`) and came up with this flowchart:

![halogen](https://user-images.githubusercontent.com/8413037/41687940-b7ee0700-749e-11e8-8939-2cebbc149c6e.png)

garyb:
Nice work! `initializer` and `finalizer` have been present for some time, but only recently become required (there was `component` vs `lifecycleComponent`). It's one of the places the current guide falls short.

JordanMartinez:
Thanks! Ah, ok. Since you mentioned in above, what's the ETA for the next release?

I also realized that the above flowchart does not include a dotted line from the `initial state rendering` and `re-render state` nodes to the `Parent passes in value of INPUT type (i) to child)` node. Hmm... actually, the flowchart doesn't also clarify what happens when the Parent passes in the same input value (child gets potentially re-rendered) as opposed to a different one (e.g. child is destroyed and a new one is created)

I'm not sure whether you'd want to use the updated flowchart or not in your documentation. If so, I can submit a PR that adds that to the docs part of this project and license it using CC0 so neither of us have to worry about license problems now or in the future.

JordanMartinez:
More complicated version, but better distinguishes how the whole thing works:
![halogen - grouped](https://user-images.githubusercontent.com/8413037/41812734-03ea51a0-76dd-11e8-8215-4ae448aa9558.png)

Note that this flowchart does not include what happens when a query evaluation done by a parent causes it to re-render its state, which passes in a new input value that destroys the old child and creates a new one. It still doesn't include the `initializer` and `finalizer` stuff as well, though I think I should add that after looking through the corresponding examples more

safareli:
[...]

PR #559, opened 2018-07-17
Fix broken example called "components" according to sync with docs files

sdolidze:
docs and examples were out of sync, "components" example was not even buildable. Minor change to sync example with docs.

### DIFF
--- a/examples/components/package.json
+++ b/examples/components/package.json
@@ -1,10 +1,10 @@
 {
   "private": true,
   "scripts": {
-    "build": "pulp build --main Example.Components.Main--to dist/example.js"
+    "build": "pulp build --main Example.Components.Main --to dist/example.js"
   },
   "devDependencies": {
-    "pulp": "^11.0.0",
-    "purescript": "^0.11.7"
+    "pulp": "^12.2.0",
+    "purescript": "^0.12.0"
   }
 }
--- a/examples/components/src/Button.purs
+++ b/examples/components/src/Button.purs
@@ -16,7 +16,6 @@ data Query a
 
 data Message = Toggled Boolean
 
-type Slot = H.Slot Query Message
 
 myButton :: forall m. H.Component HH.HTML Query Unit Message m
 myButton =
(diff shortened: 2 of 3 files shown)


garyb:
Hey, thanks - it looks like you've reverted some changes in the example here though. Master is the "next" version, so the `ChildSlots` business was correct, it's just the `package.json` that needed updating.

sdolidze:
Thanks for the quick reply :) I'm a PureScript newbie trying to make sense of how Halogen works. I tried to build the project with ChildSlots and it did not work (including changing the package.json). Is that the intended behaviour?

garyb:
If you've checked out `master`, you should be able to build the example from the root of the project with `npm run example-components` instead of building it per example.

Perhaps we should remove the per-example `package.json`s, as in times like this where `master` is ahead of the released version they won't work properly, unless you do a bunch of non-obvious stuff to make it symlink bower dependency for Halogen instead. 🤔 

Another option is to check out the v4.0.0 tag instead of `master`, and then you should be able to run the individual examples as standalone projects too. I'd perhaps suggest doing that for now, since the docs are only updated for v4 still! 🙂

sdolidze:
Now it makes total sense :) When I saw `master` and `next` branches, I assumed next was ahead. I assumed that `ChildSlots` was the old approach, when it was the new one :) (It's my third day using PureScript, so I guess it's understandable).

I totally agree, removing per-example package.json would create less room for confusion for newbies like me. I guess this PR makes no sense.

Would you accept PR deleting per-example package.json-s? Or is it something that you need to think about first?

JordanMartinez:
In your PR, you should probably include a note in the examples' ReadMe about checking out the latest release tag before trying out the examples because I don't think this is the first time this issue has appeared.

JordanMartinez:
@sdolidze If you haven't already, you might want to check out [the flowchart I made in 528](https://github.com/slamdata/purescript-halogen/issues/528#issuecomment-400071113)

sdolidze:
@JordanMartinez thanks for the flowchart, it's quite handy.

Adding note in README to check out latest release tag will not help for the time being. I just checked out `V4.0.0` and it will not build, because package.json has older version of purescript (0.11.7 instead of 12.2.0).

I totally agree with @garyb's suggestion that we should remove `package.json`-s altogether as it creates more problems than provides use for beginners.

Should I submit a PR deleting `package.json`-s in all `examples` folders and updating `REAMDE.md`-s to build only from root of the project?

sdolidze:
@JordanMartinez I'm consistently getting same error messages, even when running in clean docker container. It could be the case that I'm doing something wrong. here is my `Dockerfile` and output that it wrote, please have a look. https://gist.github.com/sdolidze/7d11d72f6ebbbaa9ab7fc246500fb053

sdolidze:
This PR does not make much sense, therefor I'm closing it. In my opinion, the best solution to this problem is to remove per-example `package.json`-s. Here is a related PR: #560

Issue #426, opened 2017-03-07
What happened to preventDefault and friends?


boothead:
I can't find the preventDefault function in halogen's API anymore, or see it mentioned in the docs. How do we do that now?

I'd also just like to say a massive thanks for halogen. The latest changes for version 1 make for a much nicer to use API and refactoring a couple of thousand line app was pretty painless. You guys rock!

cryogenian:
It works this way now 
```purescript 
import DOM.Classy.Event (toEvent, preventDefault)
data Query a 
  = DoSomething a 
  | PreventDefault Event (Query a) 

-- somewhere in render function 
onClick \e -> Just $ PreventDefault (toEvent e) $ action DoSomething

eval :: Query ~> DSL 
eval (PreventDefault ev q) = do
  liftEff $ preventDefault ev 
  eval q
eval (DoSomething next) = pure next
```

boothead:
Aaah! Perfect. thanks. Is this in the docs anywhere? Shall I leave this open as a reminder or close it?

cryogenian:
I'd leave and change the header to something like `Add preventDefault/stopPropagation etc to docs" or something :)

boothead:
Well then, I will see what I can do :-)

saylu:
This issue was very helpful for me. I often have to prevent link clicks from causing the browser to load the page, but using `preventDefault` inside every query handler is tedious.

In my project, I created variants of the events from `Halogen.HTML.Events` that have prevention built in. For example, `onClick` has its parallel in `preventClick`, which makes using `preventDefault` as simple as using one of the `prevent-` functions in your `render`.

Thought this may be useful for others. Full excerpt:

```purescript
import DOM.Classy.Event (preventDefault, toEvent)
import DOM.Event.Event (Event)
...

data Query a
  = NoOp a
  | PreventDefault Event (Query a)

eval = case _ of
  NoOp next -> pure next
  PreventDefault ev q -> do
    liftEff $ preventDefault ev
    eval q

preventClick q =
  HE.onClick \e -> Just $ PreventDefault (toEvent e) $ H.action q

render :: H.ComponentHTML Query
render =
  HH.a [ HP.href "", preventClick NoOp ] [ HH.text "Do something" ]
```

JordanMartinez:
Building off of @cryogenian and @thomashoneyman's work, I abstract things even more in my work on #563, so that I could write
[...]

PR #561, opened 2018-07-24
Add Halogen flowchart; add page on event handling and css

JordanMartinez:
This adds last version of the flowchart I created in #528.


JordanMartinez:
Documents the solutions/answers in #426

garyb:
Thanks for working on this stuff, I should have raised this sooner - but essentially this is all out of date though, you're documenting the last released version rather than the "current" version - the thing that is holding up the release of the current version is up to date docs too :wink:

A specific comment regarding the flowchart: the distinction between child and parent components doesn't really exist. In the version you are using they're just using type synonyms, and in the current/master version the synonyms are gone even.

JordanMartinez:
That's ok. I'm currently using `v4.0.0` since we're all waiting for the next release, so this helped me understand it more.

What could I do to help in terms of documentation for the upcoming release? I'm assuming that the only things that have changed are the parent-child relationship. In other words, the stuff I've covered so far in #563 is still correct, right?

> A specific comment regarding the flowchart: the distinction between child and parent components doesn't really exist. In the version you are using they're just using type synonyms, and in the current/master version the synonyms are gone even.

True. Still, I distinguished between the two to show that a child won't handle raised messages from its children because it doesn't have any whereas a parent (which could be a child to another parent) does need to handle these things.
Still, a single component that links everything together would be more accurate to the spec and simpler to chart.

PR #563, opened 2018-07-25
Bottom-Up approach to explaining Halogen

JordanMartinez:
As described [in my comment](https://github.com/slamdata/purescript-halogen/issues/528#issuecomment-407833179)

I'm still working on the event handling part. I ran out of time today.

### DIFF
--- a/docs/v2/1 - Introduction.md
+++ b/docs/v2/1 - Introduction.md
@@ -1,0 +1,13 @@
+# Introduction
+
+A Halogen application consists of a tree of components. Each component is a self contained unit that has its own state, and re-renders when the state changes.
+
+A component modifies its state by evaluating inputs known as queries. Each component defines a type that describes all the possible queries it will expect. We use the term "query" rather than something like "input" as they can return details about the component state as well as being able to update it.
+
+A component can also emit messages during query evaluation. As with queries, these messages have a type specified by the component. Messages are used to notify listeners of activity within the component. The motivating example for this is to allow a parent component to observe activity within a child.
+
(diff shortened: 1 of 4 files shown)


JordanMartinez:
Note: the `recursiveQuery` function described in the `Evaluating Recursive Queries` section results in a compiler error for `HH.button`.

JordanMartinez:
From slack:
> kadblas [3:12 PM]
> Halogen question here. Is there a way to query the view for information during runtime? For example, if I create a `div` in `render`, is there a way to get the width of that div during runtime? If so, how?
> 
> justinw [3:13 PM]
> easy way is through refs
> - https://github.com/justinwoo/purescript-web-audio-player-demo/blob/6f6b692ea83e840d39b32f2bd7db4a7965f44ca7/src/Main.purs#L82
> - https://github.com/justinwoo/purescript-web-audio-player-demo/blob/6f6b692ea83e840d39b32f2bd7db4a7965f44ca7/src/Main.purs#L133
> 
> garyb [3:15 PM]
> example in here too https://github.com/slamdata/purescript-halogen/blob/v4.0.0/examples/ace/src/AceComponent.purs#L57
>
> kadblas [3:17 PM]
> Excellent. Thanks @justinw & @garyb

This PR should extract this code into a pattern to follow.

AlexaDeWit:
Just wanted to say, as a Halogen user... thanks for working on this!

JordanMartinez:
@AlexaDeWit Thanks! I've been wanting to finish documenting it, but I'm waiting for the `5.0.0` release.

Also, I've since learned how `VariantF` works. If one used `VariantF` as their query type, one could mix in `preventDefault` and `stopPropagation` into their `eval` function easily.

JordanMartinez:
Closing this since I've created the `learn-halogen` repo that is even better than this.

PR #564, opened 2018-07-27
Added inputR for easier recursive query evaluation

JordanMartinez:
This adds the function I came up with in #426.

### DIFF
--- a/src/Halogen/HTML/Events.purs
+++ b/src/Halogen/HTML/Events.purs
@@ -1,6 +1,8 @@
 module Halogen.HTML.Events
   ( input
   , input_
+  , inputR
+  , inputR_
   , handler
   , onAbort
   , onError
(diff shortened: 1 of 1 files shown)


JordanMartinez:
Closing this one as well as I'm not sure this really is necessary/helpful anymore.

cdupont:
Hello, I'm also trying to stop event propagation. Why is this not necessary anymore?

garyb:
`Action`s are no longer kinded `Type -> Type` so this helper isn't required in the example situation any more.

This wasn't directly related to stopping propagation, it was just an example of the pattern where this function would have been useful.

garyb:
If you're just looking for help on how to stop propagation, then this is the function you're looking for: https://pursuit.purescript.org/packages/purescript-web-events/docs/Web.Event.Event#v:stopPropagation 🙂. Use it with `liftEffect` in the action handler that receives the `Event`. 

There's a similar example for `preventDefault` here: https://github.com/purescript-halogen/purescript-halogen/blob/master/examples/keyboard-input/src/Main.purs#L59, although this also required coercing the `KeyboardEvent` to `Event` (`KE.toEvent`) which won't be necessary if the handler you're using raises `Event`.

PR #565, opened 2018-08-10
Mention bower and purescript deps in readme

dv00d00:
On a machine without installed bower build fails, had to take look into travis build log to figure out what is going on...


garyb:
Extremely belated thanks!

PR #576, opened 2018-09-23
Partial query evaluation

garyb:
Check it out @natefaubion!

Two-fold rationale:
1. we looked into it before of course, but I couldn't figure out the plumbing - so the same reasons as then, allows proxy components to fail properly, etc.
2. I wanted to make a less boilerplatey construction method for the components again, and by allowing query eval to return `Nothing` that leaves us free to have the query type polymorphic until a real eval is provided for it.

I've only updated two of the examples, for proof that it's usable, as I wanted to see if we can think of reasons why we shouldn't do this.


src/Halogen/Component.purs
```diff
@@ -147,7 +72,41 @@ component = mkComponent <<< go
...
-  , eval :: HalogenQ f act i ~> HalogenM' s act ps o m
+  , eval :: HalogenQ f act i ~> HalogenM s act ps o m
+  }
+
+type EvalSpec s f act ps i o m =
+  { handleAction :: act -> HalogenM s act ps o m Unit
+  , handleQuery :: forall a. f a -> HalogenM s act ps o m (Maybe a)
+  , receive :: i -> Maybe act
+  , initialize :: Maybe act
+  , finalize :: Maybe act
```
garyb: I guess these could be `Maybe (HalogenM s act ps o m Unit)` rather than going via action handling, but I kinda like handling these in one place (which is already true if `HalogenQ` is handled rather than going the `mkEval` route).

natefaubion: I don't think they would need to be `Maybe`, since they can just be `pure unit`.

src/Halogen/Aff/Driver/Eval.purs
```diff
@@ -52,23 +53,23 @@ evalF render ref = case _ of
       st { refs = M.alter (const el) p st.refs }
   Action act -> do
     DriverState st <- liftEffect (Ref.read ref)
-    evalM render ref (st.component.eval (Handle act unit))
+    void $ evalM render ref (st.component.eval (Handle act unit))
```
safareli: Last part of `evalM` is `HalogenM s act ps o Aff ~> Aff` so I guess you don't need `void` here.

src/Halogen/Component.purs
```diff
@@ -147,7 +72,41 @@ component = mkComponent <<< go
...
+  }
+
+mkEval
```
safareli: What you think about calling `mkEval` from `mkComponent` (i.e. `ComponentSpec` taking `EvalSpec` ?

src/Halogen/Aff/Driver/Eval.purs
garyb: `evalM` is `-> Aff (Maybe a)` now, so we need this void to avoid a discard error.

src/Halogen/Component.purs
garyb: We can't do that without exposing another different version of `mkComponent`, since Halogen uses it internally. Plus I wanted to provide the choice of writing an `eval` in terms of `HalogenQ` directly. 

Having `mkEval` in there seemed better given both of those considerations, as it should avoid some confusion about things like when we had 4 different component constructors before - they were all the same aside from some synonyms and field-defaulting, but that wasn't obvious to the casual user. Since the same actual `component` constructor is used here, and we just use another function to build up part of its internals, it shows that there's nothing special going on, no matter how the component is constructed.

src/Halogen/Aff/Driver/Eval.purs
safareli: evalQ is `-> Aff (Maybe a)` but [evalM is `~> Aff`](https://github.com/garyb/purescript-halogen/blob/partial-query/src/Halogen/Aff/Driver/Eval.purs#L68-L73) and as `x ~> y = forall a. x a -> y a` I thought `void` is not needed.

garyb: Ah yeah, you're right. I must have changed some stuff around while working towards this PR, as I wouldn't have added a `void` without the compiler requiring me to do so :smile:

safareli: Yeah I thought so 😃

acple:
more I like `mkEval` has this type:
```purescript
mkEval
  :: forall s f act ps i o m
   . (EvalSpec s f act ps i o m -> EvalSpec s f act ps i o m)
  -> HalogenQ f act i ~> HalogenM s act ps o m
```
does no longer require expose `defaultEval`.
`eval: H.mkEval _{ handleAction = handleAction }`

garyb:
I did consider that too, or exporting a function that does that as well as having a `defaultEval` record.

PR #589, opened 2019-01-02
Remove purescript-profunctor-lenses dependency

davezuch:
Looks like the `ChildPath` component was the only place this was used, with that gone it seems this isn't a necessary dependency anymore


garyb:
Nice, thanks!

PR #590, opened 2019-01-14
Add Real World Halogen to the getting started section

thomashoneyman:
Small commit: this helps new users be aware of Real World Halogen so they can see how the framework looks when it's used to implement a full application.


garyb:
Oops, sorry! This should have been insta-merge since we'd already talked about it :smile:

PR #593, opened 2019-02-06
Example demonstrating how to use the ProComponent API

nicholassaunders:
I stumbled across the [```Halogen.Component.Profunctor```](https://github.com/slamdata/purescript-halogen/blob/master/src/Halogen/Component/Profunctor.purs) module by chance a few days ago but couldn't find any relevant documentation or examples, hence this pull request.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -50,6 +50,7 @@
     "purescript-ace": "^7.0.0",
     "purescript-aff-coroutines": "^7.0.0",
     "purescript-affjax": "^9.0.0",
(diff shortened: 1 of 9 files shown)


nicholassaunders:
I'm going to close this for now, for two reasons:
* The Halogen API has changed, so this is broken.
* This example probably doesn't line up with a real-world use case. I think this is more likely to be used for something like wrapping/unwrapping values rather than conversions.

PR #595, opened 2019-02-24
Prerelease info updates

garyb:


safareli:
I would suggest to mention that forked computations are killed if component is unmounted, where in v4 it wasn't happening.

PR #599, opened 2019-03-21
Add `renderWidget` for `HTML`

garyb:


src/Halogen/HTML/Core.purs
```diff
@@ -57,6 +58,10 @@ instance bifunctorHTML :: Bifunctor HTML where
 instance functorHTML :: Functor (HTML p) where
   map = rmap
 
+renderWidget ∷ ∀ p q i j. (i → j) → (p → HTML q j) → HTML p i → HTML q j
+renderWidget f g (HTML vdom) =
```
natefaubion: The function we use for constructing this below is `slot`, but here it's called `Widget`.

garyb: Yeah, I wondered what to call it - since we still call the type var `p` for "placeholder" too, which is super duper old terminology. What do you think... `renderSlot`? `replaceSlot`?

natefaubion: Is `Core.slot` used as primitive anywhere other than in `HTML.slot`? Maybe `Core.slot` should be `Core.widget` and then `renderWidget` is A-OK! 😄

garyb: I like that, yeah.

PR #607, opened 2019-07-01
additional workaround for memory leak issue

acple:
Hi!
# 606 is less enough to resolve memoryleak issue.
So I added some workaround fix.
Here is repro project: https://github.com/acple/purescript-halogen-memoryleak-repro (updated to rc.5), I checked heap consumption does not increased after apply this patch.
See also # 586 comments!


garyb:
Thank you! Sorry I missed this part in the original fix, and for the delay in getting this merged.

PR #608, opened 2019-07-16
Use event rather than coroutines

garyb:
I haven't decided for sure whether to go with this yet, comments welcome!


thomashoneyman:
With a brief read through this is _much_ more understandable, which is a thumbs up from me.

garyb:
Yeah, I think it's quite a bit simpler, and avoids the complication of `EventSource` having to carry the `m` too.

There's no equivalent for the old `affEventSource` now, but I'm not sure how valuable that was. I'm pretty sure I only use it as `effectEventSource` (which is now kinda redundant, since that's basically what `Event` is itself) and `eventListenerEventSource`.

I guess maybe when using `Bus` I used `affEventSource`, but lately I've found I can replace those with `Event` too for my use cases.

thomashoneyman:
Ah, that's true, I use `affEventSource` in Real World Halogen with `Bus` to push global state changes to subscribed components. I'm on vacation now so I haven't explored this, but it seems like a straightforward use case that should still work -- SlamData does the same thing, right?

garyb:
My current thinking for this is to add an `affEventSource` equivalent (it'll just use `launchAff` to make it work, pretty much), but then release it as Halogen v6. So there will never actually be a v5 release, it'll just go from the RCs to v6. Technically I could make it a new RC, but I think that'll break stuff as it seems v5 is getting a decent amount of use already.

I've been working on the docs again in earnest too. As soon as I get something that is up to date and covers all the major necessary aspects I'll make a release, it can be improved after the fact if there are things that are unclear or missing. 🙂

thomashoneyman:
In my opinion this could be another RC leading to Halogen 5, or the first thing going into Halogen 6, but it would feel a little odd for there to never be an official Halogen 5. But I’m not sure how ‘stable’ release candidates are meant to be.

Sounds good wrt the docs!

garyb:
My worry about making an RC with this change in, is I think bower will auto upgrade people if they haven't exactly pinned the `-rc.x` part. I guess I could just make the v5 release and then immediately do the v6 with this change - that might be better, just in case a fix needs backporting or something.

thomashoneyman:
That seems sensible. I'm still a little thrown by an added dependency requiring a major version bump when I still have the idea of major version bumps as significant changes to a library (like Halogen 4 -> 5). But there's no real harm in moving to Halogen 6 directly.

PR #613, opened 2019-08-17
Add associated-type like thing for surface/element

garyb:
This means `HalogenF`, etc. no longer has HTML specific things lurking in it at all.

Plus, @thomashoneyman and I have been discussing adding a "portal" thing to `ComponentSlot`, and that would involve needing to provide an element appropriate to the surface type too, and I didn't really want to hard code `Element` in there as well, given it already had `surface` as part of the type.

### DIFF
--- a/examples/ace/src/AceComponent.purs
+++ b/examples/ace/src/AceComponent.purs
@@ -53,7 +53,7 @@ initialState _ = { editor: Nothing }
 render :: forall m. State -> H.ComponentHTML Action () m
 render = const $ HH.div [ HP.ref (H.RefLabel "ace") ] []
 
-handleAction :: forall m. MonadAff m => Action -> H.HalogenM State Action () Output m Unit
+handleAction :: forall m. MonadAff m => Action -> H.HalogenM HH.HTML State Action () Output m Unit
 handleAction = case _ of
   Initialize -> do
     H.getHTMLElementRef (H.RefLabel "ace") >>= traverse_ \element -> do
@@ -72,7 +72,7 @@ handleAction = case _ of
       text <- H.liftEffect (Editor.getValue editor)
       H.raise $ TextChanged text
 
-handleQuery :: forall m a. MonadAff m => Query a -> H.HalogenM State Action () Output m (Maybe a)
+handleQuery :: forall m a. MonadAff m => Query a -> H.HalogenM HH.HTML State Action () Output m (Maybe a)
 handleQuery = case _ of
   ChangeText text next -> do
     maybeEditor <- H.gets _.editor
--- a/examples/ace/src/Container.purs
+++ b/examples/ace/src/Container.purs
@@ -55,14 +55,14 @@ render { text: text } =
         [ HH.text ("Current text: " <> text) ]
     ]
 
-handleAction :: forall o m. MonadAff m => Action -> H.HalogenM State Action ChildSlots o m Unit
+handleAction :: forall o m. MonadAff m => Action -> H.HalogenM HH.HTML State Action ChildSlots o m Unit
 handleAction = case _ of
   ClearText ->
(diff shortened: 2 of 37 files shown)


garyb:
Mainly looking for feedback on the `Surface` class / `SurfaceElement` "associated type".

https://github.com/slamdata/purescript-halogen/pull/613/files#diff-46f0dcd8aed6e5d3a7872613a8ae103d

Is it unsafe in some way that I'm not seeing, etc.

garyb:
The other option is maybe we just delete `surface` entirely, since I don't think anyone has even tried to do anything other than HTML UIs. Well, there's SVG too, but that uses the same basic types.

MonoidMusician:
I'm curious, could you expand on what you mean by "portal to `ComponentSlot`"?

I've been thinking recently that it would be nice to have an actual functor for `ComponentHTML`, to map over output messages easily, but maybe that's not related ... Here's how I did it in my project: https://github.com/MonoidMusician/dhall-purescript/blob/ee8c46ec4c130f60d16fd8e341a302d8c2e71d9e/src/Dhall/Interactive/Halogen/AST.purs#L41

Anyways, I think the associated type idea is sound, Phil's used it: https://pursuit.purescript.org/packages/purescript-incremental-functions/2.0.0/docs/Data.Incremental#t:Change 😄

Could we have a type synonym for the common case, with `HTML`?

I do agree that the generality may or may not be worth it. I know there has been talk about making a console renderer for Halogen, but you know what they say about all talk and no action.

thomashoneyman:
Portals are in reference to this (and the feature of the same name in React):
https://github.com/thomashoneyman/purescript-halogen-portal

but for this to work well it can’t be done in user land and so would make a better addition to Halogen directly

garyb:
Oops, I did make a branch that dropped `HTML`, I guess I didn't push it / make a PR for it. But it fully stripped the ability to abstract over surface out, which would indeed make it quite a chore to restore.

I think I'll do it again, but just strip it out of the component layer, and preserve it in the driver stuff, as then it wouldn't be too bad to bring back. It might even be better - could just define a new `Component` type for an alternative surface, as it's not like the `HH` / `HE` / `HP` stuff will work with it anyway, Mixing surface types isn't something that's allowed for now either, so having two entirely separate "worlds" of `Component` is no worse in that regard.

garyb:
Hmm, nevermind, that doesn't work because the driver stuff makes references to `ComponentSpec`, `ComponentSlot` and so on, so the surface var would just be a weird type floating around that wouldn't make much sense anymore.

garyb:
It only really affects the `Component` type from a library user's perspective. It does simplify thing in the driver parts too, but nobody's reading that 🤣. We could actually simplify it further and throw out the option of any driver aside from the `halogen-vdom` one at this point too, using the same reasoning.

But! A major plus of getting rid of `surface`, even if it only really appears in one place, is it's just one less thing to ever have to explain. And one less type to remember to apply when defining components.

garyb:
> That one I'm not as keen on, unless throwing out the option happens to simplify something from a library user's perspective also. Would it?

No, just the driver implementation.

> Just a thought... If someone did want to implement a different `surface` type, would it be better for that library to fork this repo and continue its development separate from this one, even if they share a lot of similarities? I imagine they would experience a similar problem of "but the type signatures!" if the `surface` type as kept. (Again, I'm speaking from ignorance, not depth of understanding as to what the full implications of this decision is)

A completely separate fork is what would have to happen if we make this change, yeah. Realistically, if someone is able to dig into the driver implementation enough to be able to change the `surface` then one extra type variable is barely going to be a dent cognitively. :wink:

I'll chew on this a bit more, but probably will go for it. Halogen-with-surface has been around for nearly 3 years at this point.

PR #617, opened 2019-09-01
Update README link for Real World Halogen

thomashoneyman:
Real World Halogen's master branch is now on Halogen 5. I've updated the link here so that it no longer points to the pull request.


garyb:
Thanks 👍

PR #618, opened 2019-09-09
CHG: Minimal migration to Aff with generalised errors.

eric-corumdigital:


garyb:
I assume this is something we'd need to do if https://github.com/purescript-contrib/purescript-aff/pull/177 ever gets merged?

eric-corumdigital:
> 
> 
> I assume this is something we'd need to do if [purescript-contrib/purescript-aff#177](https://github.com/purescript-contrib/purescript-aff/pull/177) ever gets merged?

Yes, though I have given up on modifying Aff to that extent. By the time I achieve what I want it is too far incompatible with the original Aff. I am instead working on a new library that builds from the lessons of Aff. I will retract both these PRs.

garyb:
Ok, thanks! I'm interested to see what you end up with.

PR #621, opened 2019-09-11
Use `tell` or `request` directly to raise queries

### LINKED ISSUE #611: Merge `request` and `tell` with `query`?
So instead of doing:

``` purescript
H.query _child "id" (H.request GetInfo)
void $ H.query _child "id" (H.action DoWhatever)
```

There'd be:

``` purescript
H.request _child "id" GetInfo
H.tell _child "id" DoWhatever
```

Aside from it being a bit simpler visually, and not having to explain that `request` is just `_ $ identity` / `tell` is just `_ $ unit`, it also removes the need to `void` all the tell-style requests, as they return `Maybe Unit` so can't be `discard`ed.

I guess the same would go for `queryAll` - have a `requestAll` with a result, `tellAll` with `Unit`.

Thoughts?

garyb:
Resolves #611 

Unfortunately had to keep the old functions as `mkTell` and `mkRequest` though, for use with the top level driver `query`. I suppose another option would be to replace that with separate `tell` and `request`
properties on `HalogenIO` too.

### DIFF
--- a/examples/components-multitype/src/Container.purs
+++ b/examples/components-multitype/src/Container.purs
@@ -73,7 +73,7 @@ render state = HH.div_
 handleAction :: forall o m. Action -> H.HalogenM State Action ChildSlots o m Unit
 handleAction = case _ of
   ReadStates -> do
-    a <- H.query _a unit (H.request CA.IsOn)
-    b <- H.query _b unit (H.request CB.GetCount)
-    c <- H.query _c unit (H.request CC.GetValue)
+    a <- H.request _a unit CA.IsOn
+    b <- H.request _b unit CB.GetCount
+    c <- H.request _c unit CC.GetValue
     H.put { a, b, c }
--- a/src/Halogen/Query.purs
+++ b/src/Halogen/Query.purs
@@ -2,8 +2,10 @@
 -- | component's `eval` function.
 module Halogen.Query
   ( Tell
+  , mkTell
   , tell
   , Request
+  , mkRequest
   , request
   , getHTMLElementRef
   , module Exports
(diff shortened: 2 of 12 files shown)


examples/components-multitype/src/Container.purs
```diff
@@ -73,7 +73,7 @@ render state = HH.div_
 handleAction :: forall o m. Action -> H.HalogenM State Action ChildSlots o m Unit
 handleAction = case _ of
   ReadStates -> do
-    a <- H.query _a unit (H.request CA.IsOn)
-    b <- H.query _b unit (H.request CB.GetCount)
-    c <- H.query _c unit (H.request CC.GetValue)
+    a <- H.request _a unit CA.IsOn
```
thomashoneyman: 👍 so much nicer

src/Halogen/Query.purs
```diff
@@ -51,10 +56,21 @@ type Tell f = Unit -> f Unit
...
 -- | sendTick :: forall o. H.HalogenIO Query o Aff -> Aff (Maybe Unit)
--- | sendTick app = app.query (H.tell Tick)
+-- | sendTick app = app.query (H.mkTell Tick)
 -- | ```
-tell :: forall f. Tell f -> f Unit
-tell act = act unit
+mkTell :: forall f. Tell f -> f Unit
+mkTell act = act unit
+
+tell
```
thomashoneyman: I think this could use a documentation comment (maybe that's for a later PR as this becomes more stable) along the lines of the text in the original issue

garyb: Yeah, definitely 🙂

JordanMartinez:
> Unfortunately had to keep the old functions as mkTell and mkRequest though, for use with the top level driver query. I suppose another option would be to replace that with separate tell and request
properties on HalogenIO too.

Not sure if I'm understanding the full issue here, but why not name them someting like `ioTell` and `ioRequest`? Or `tellIO` and `requestIO`?

JordanMartinez:
Oh, never mind. I see what the code change is now.

Issue #588, opened 2019-01-01
Documentation is not up to date for master branch


beckyconning:
https://github.com/slamdata/purescript-halogen/blob/master/docs/5%20-%20Parent%20and%20child%20components.md

Is misleading as the references to `ParentHTML` and `ParentDSL` give the impression that there is a difference between parent and child components.

garyb:
Yeah, the documentation is all pretty wrong for the code in master - it will essentially need a rewrite before the next release is made. The plan is to pre-release after #576, and then make a proper release once the docs are updated too.

JordanMartinez:
(I'm posting this in other doc-related issues. Sorry for the multiple notifications.)

Just FYI. I've put together a bottom-up tutorial on Halogen for the `v5.0.0-rc.4` release: https://github.com/jordanmartinez/learn-halogen There are still a few things it doesn't cover yet, but it should get most people up to speed.

Benjmhart:
I'm getting set up to do a bunch of documentation updates in the coming weeks, hopefully will be able to bring the docs fully up to date for v5

garyb:
I do have a whole new set of documentation in progress, I had been making good progress with it, but have a deadline this week so it's been on hold again. If you get there before I finish, I'll definitely take it though! It would make some things easier, as there are a few changes stacking up that I might want to make as a v6 rather than having a confusing distinction between v5s, so I could drop my new document then instead.

Benjmhart:
I'm about 3 chapters in,  so hoping to have the existing chapters replaced with v5 docs within another two weeks depending on my availability.

Benjmhart:
if we get this merged in,  is there anything else holding back v5 from official release?   part of my motivation is getting the Pursuit docs updated to v5 eventually. @garyb

garyb:
Yeah, the docs are the only thing holding back a v5 release.

Benjmhart:
cool : )

sigilion:
Have you made much progress? I might have time to jump in if there's still work to do.

Benjmhart:
Hey,  sorry for the delay I've had some travel for work,   I'm revising the
last chapter to bring us to documentation parity with v4 right now, I'm
hoping to PR that by Sunday or Monday.

I have some additions, but I think those are best left to a follow-up PR.

how does that sound?

Benjamin James Mikel Hart

sigilion:
[...]

PR #627, opened 2019-09-30
v5 documentation - first pass

Benjmhart:
Hi there,  I've gone over the first 5 chapters and updated to use v5 types etc,  and shuffled things around a bit.

I'm also adding a few chapters that I feel would be beneficial,  but while I'm writing those and making example code I figured we should get the current chapters PR'ed so I can get any feedback on factual/stylistic correctness.

Please let me know if you have any feedback and I'll correct.

Thanks!


Benjmhart:
I should link this to the original issue: https://github.com/slamdata/purescript-halogen/issues/588

garyb:
I'm really sorry this has just sat since you opened it, I do very much appreciate the effort. I'm aiming to get it reviewed/merged/released this week for sure.

Benjmhart:
@garyb - it's no big deal to me! I know the life of a maintainer can be tough and I appreciate you getting to it whenever you can.

Though i would really appreciate the review so I can move ahead and finish the docs with the planned chapters (and possibly add to the working examples).

vlatkoB:
First of all, big thanks for these docs where one can see the current state. :-)

I'm a beginner and started reading the docs v5 (only two pages for now), and found some issues and have some suggestions. Not sure if this is the right place for such comments, so let me know if not.

1) https://github.com/Benjmhart/purescript-halogen/blob/master/docs/1%20-%20Introduction.md#introduction

> Message - Defines the outbound events that the component can broadcast.  ...

I found it very confusing for a beginner (me) to grasp all the directions the info goes. There are `Messages`, `Queries`, `input`. It seems clear when you read it, but I get lost when implementing, using wrong ones in the wrong places.

I see the term  `Message` as a direct communication between two known parties (unless broadcast or _message in a bottle_ :-) ), and in this context, the info goes only one way (_upward_) and to the unknown someone (_parent_) and there can't be any response, so maybe some other term could be used. `Output` would be fine, but maybe `Report` is even better, as it is usually used in the context of notifying the superior about the current state or some progress, and the response is not expected (as in this case). A nitpick...


2) https://github.com/Benjmhart/purescript-halogen/blob/master/docs/2%20-%20Defining%20a%20component.md#halogen-component

> o - the type of output generated by this component, this would be a value sent to **a child** component. These are also called messages.


Should **a child** be **the parent**?

3) https://github.com/Benjmhart/purescript-halogen/blob/master/docs/2%20-%20Defining%20a%20component.md#the-html-dsl
> elementName [ ... properties ... ] [ ... children ... ]
> 
> There are exceptions for elements that expect no children, such as text, in which case the **second array** is omitted.

Aren't both arrays omitted and replaced by a single value (at least for `text`)?

4) https://github.com/Benjmhart/purescript-halogen/blob/master/docs/2%20-%20Defining%20a%20component.md#evaluating-actions

Just one example, but would be nice for all examples in the docs, not necessarily in the code:
> handleAction ∷ forall o m. Action → H.HalogenM State Action **()** o m **Unit**

As simple stuff is used, it is not easy to get the meaning of some things. I would personally prefer something like this:

```
type ChildSlots = ()
type ReturnValue = Unit
handleAction ∷ forall o m. Action → H.HalogenM State Action ChildSlots o m ReturnValue
````
This way we can gradually get accustomed to the types and their meanings, even though they may not be relevant at that point (i.e. they are _unused_/_empty_) . Not sure what others would think...

**Comment by Benjmhart** [comment]
@vlatkoB - Thanks for your great feedback! I'll try and incorporate it within the next day or two

**Comment by garyb** [comment]
A couple of minor comments in response to the response 😄:

Regarding "message", in the completely new reference docs thing I started working on, I changed to consistently using the term "output message". Not sure if that helps, but I prefer that to "report", since to me "report" implies a request was made first, which isn't necessarily the case.

`handleAction` is required to return `Unit`, so I don't think naming that `ReturnValue` will help 🙂. Instead of making `()` a synonym using polymorphic variable names might be an option too? So like: 

``` purescript
handleAction ∷ forall slots output m. Action → H.HalogenM State Action slots output m Unit
```

I recommend using polymorphic variables where possible for things that "don't matter" on function/value definitions anyway, rather than fixing them to `Void`, `()`, and the like.

vlatkoB:
@garyb What do you mean by "handleAction is required to return Unit"? 
I have a function
```handleQuery :: forall a. CQuery a -> H.HalogenM CState CAction () CMessage m (Maybe a)```

Other comments look fine to me. It is an improvement. Especially polymorphic vars. Shorter, but with the same amount of information.  Nice.

**Comment by garyb** [comment]
I'm not sure what you mean also now :smile:. `handleQuery` and `handleAction` have different types, so they'd also need different type aliases. Would `ReturnValue a` any better than `Maybe a`? That seems to make it more opaque rather than less, since you'd have to go through another step to figure out what the type is.

**Comment by vlatkoB** [comment]
What I meant is to somehow use more descriptive names to get to know the types. That's all.

Maybe there could be a convention that when hardcoded types are used in the signature, it would mean that they are fixed.
For example, if `handleAction` must return `Unit`, that would be fine, if `slots` must be empty, `()` is fine.
But if a function can return something else, let's use `a`, so one knows it depends on something.
Of course, I'm talking about the code in docs, not the code in examples.

I'm writing this from memory, as I have become a bit more knowledgable about Halogen that is not an issue for me anymore, but I remember that was one of the largest obstacles for me. I never knew why is there some hardcoded type, is it fixed or is it from the example above, the previous one, etc.
Just my two cents. :-)

**Comment by vlatkoB** [comment]
5)
https://github.com/Benjmhart/purescript-halogen/blob/master/docs/4%20-%20Running%20a%20component.md#output-and-subscribers

Maybe explain why in`RunUI` paragraph example `hoist` is used, and in `Output and Subscribers` paragraph example the component is used directly.
```
let
    rootComponent :: H.Component HH.HTML Router.Query {} Void Aff
    rootComponent = H.hoist (runAppM environment) Router.component
  io <- runUI rootComponent {} body
```
vs
```
io <- runUI B.component unit body
```
We do not see the type of `Router.component`, so it is not clear.

6) Same page. Mixing of Unicode and Ascii chars

```
body <- HA.awaitBody
state0 ← io.query $ H.request B.IsOn
```

7) 
https://github.com/Benjmhart/purescript-halogen/blob/master/docs/5%20-%20Parent%20and%20child%20components.md#parent-and-child-components

Do you think this
```
HH.slot _button unit Button.component unit (Just <<< HandleButton)
```
would be better as 
```
HH.slot _button unit Button.component unit handleButtonOutputMessage
  ...
handleButtonOutputMessage :: ButtonMessage -> Action
handleButtonOutputMessage (Just <<< HandleButton)
```
`(Just <<< HandleButton)` occurs on 2 places

8)
https://github.com/Benjmhart/purescript-halogen/blob/master/docs/5%20-%20Parent%20and%20child%20components.md#rendering

The code slipped in the beginning of the paragraph.

> HH.slot ButtonSlot Button.myButton unit (HE.input HandleButton)u The render function for a parent component must 

Same paragraph
```render :: s -> H.ComponentHTML Action () m```
Confused by
> `s` is the surface or way that we'll be rendering the component (typically HTML)
Should be 
```
render :: forall m. State -> H.ComponentHTML Action () m 
```

9)
> The change in component and handler would have no effect. The child component that was initialized in that slot will be preserved, and the original handler setup will be used when it emits messages.

I'd put that in **bold** or put a "Warning:" in front. I had an issue with exactly that yesterday. It is somewhat unexpected behavior, and yet easy to miss on the first read.

**Comment by Benjmhart** [comment]
these documents are now updated to include your feedback @garyb and @vlatkoB.

some notes:

- terminology around 'output' and 'message' has been unified to 'Output messages' and language about the data flow from child to parent was clarified
- several typos were corrected
- language was added to clarify that different `Leaf` types within the HTML DSL expect different parameters depending on their semantics within HTML
- I didn't make type synonyms around every occurrence where `()` or `unit` took the place of a named type variable. if there is strong enthusiasm I will, however it does require changes to be reflected in the examples as well.
- additional language was added to clarify the usage of `hoist`
- the no-op behavior between parent and child was bolded.

in general, this addresses all the concerns I've seen as of this writing.

**Comment by Benjmhart** [comment]
@garyb  - any other feedback?   I'd really like to start adding the additional chapters and do some work around documenting how to do semantic UI's in Halogen

**Comment by milesfrain** [comment]
@Benjmhart Excellent work on the v5 update.

I read through all chapters with a beginner's eye, and tried to add some clarity to the parts I found confusing. Otherwise, all my other changes are minor typos and formatting.

I made a PR on your fork in https://github.com/Benjmhart/purescript-halogen/pull/1

If these changes look good, then you can just merge my PR into your `master` branch, and then it should get tacked-onto this PR.

Here are some notes on the topics that I'm still confused by:

1. **What is `a` in `HalogenM`?**
```purescript
newtype HalogenM state action slots output m a
```
Chapter 2 describes this as "an underlying value of unit". Don't know what that means.

2. **Why do queries need `a`?** (seen in chapter 4).
```purescript
data Query a
  = IsOn (Boolean -> a)
  | SetState Boolean a

handleQuery :: forall m a. Query a -> H.HalogenM State Action () Message m (Maybe a)
handleQuery = case _ of
  IsOn k -> do
    enabled <- H.gets _.enabled
    pure (Just (k enabled))
  SetState enabled a -> do
    H.modify_ (_ { enabled = enabled })
    pure (Just a)
```
Chapter 5 explains that `a` can be `unit` in the guide's examples. In what situations would `a` need to be something other than `unit`? 

> `handleQuery` accepts some `Query a`, which refers to the component's Query type. It returns a `HalogenM` with the new `State` value (which may or may not have been modified), along with the other `HalogenM` parameters, the last parameter is `(Maybe a)`, `a` here should always be `unit`, however since we cannot prove that, we leave it as a type variable.

3. **Where should we manually diff state?**

Chapter 5 suggests doing this in `handleAction`:

> We can see in this case we are using `when (oldN /=n)` to determine whether or not to update the state, to avoid unnecessary rendering.

Doesn't halogen have some built-in mechanism of diffing state between renders and skipping the re-render if the state is unchanged? Does `put` force a re-render? Or is this diffing operation also a step that we're trying to avoid?

[Real World Halogen](https://thomashoneyman.com/guides/real-world-halogen/using-halogen-components/) has more detailed performance tips, but reading those tips leads me to believe that there are two diffing phases, one to check for state changes before each render call, and one to diff the entire virtual DOM once all renders are complete:

> Every time the state of your component changes, Halogen will re-run your render function and use its virtual DOM to diff changes against the current state of your UI in the browser. This loop is run whether you actually changed how your page should render or not. Be careful about excessively updating your component state and try to minimize the number of times you call `modify` or `put`.

Perhaps @thomashoneyman can clarify. Also, do you have write access to this org and free cycles to help with merging this v5 documentation?

Benjmhart:
@milesfrain  - Thanks for your feedback - unfortunately I'm not as free to do open source these days as I was when I put this PR up,  however these are friendly amendments to the PR.   

I'll incorporate these when I get time again, which may be in a couple of weeks.

garyb:
I'm finally going to merge this - further improvement PRs can be opened on this repo. 🙂

Again, sorry this has taken me so long to get to, there were some more unforeseen things that have kept me occupied, but also I'm going to try to be less precious about changes to the documentation and such. 

Halogen is now in its own org so I'll be able to add new contributors to hopefully ensure this situation never happens again. 🙂

flip111:
Congratulations on the v5 documentation!

I would like lo know how to use `Halogen.HTML (HTML)` with events

example: https://github.com/thomashoneyman/purescript-halogen-realworld/blob/master/src/Component/Part/FavoriteButton.purs#L42

The documentation covers `Halogen.ComponentHTML`

Not sure about the pros and cons of each and when to use them.

garyb:
`ComponentHTML` is a synonym built on top of `HTML`:

``` purescript
type ComponentHTML action slots m = HTML (ComponentSlot HTML slots m action) action
```

You can use `HTML` whenever you have a piece of HTML that doesn't have child components, but I'd say generally it's unusual to use `HTML` over `ComponentHTML`, which is why it's not discussed in the current docs. It'll probably be mentioned in the reference docs that I'll be adding for v6.

flip111:
Thanks a lot @garyb

PR #636, opened 2020-02-09
Simplify handlers by removing `Maybe` return value

garyb:
In the ongoing idea to simplify things in Halogen 6, I've been thinking
a lot about doing this lately. I don't think I've ever used a `Nothing`
return in an event handler based on the value of the `Event`.

I have very occasionally had a listener that returned `Nothing` based on
the current component state, but there are two ways that can be handled
even after removing this: either omitting the listener in the prop
construction of the HTML (with `guard` or something), or just checking
the component state when handling the raised action instead.

Similarly the handler for `slot`ted child components - I do occasionally
use `const Nothing` as the handler when setting things up, but I've
sometimes been bitten by that by forgetting to go back and handle the
messages. If the component emits no message then `absurd` still works as
a handler in the absence of the `Maybe`.

So basically: I think this is more ergonomic for the vast majority of
use cases for handlers. For the rarer occurrences where filtering was
use, it's still possible to do that, just there's an extra step to deal
with.

I did also consider adding primed variants of everything, that still had
`Event -> Maybe i`, but despite my love of primed things, I figured
leaving them out entirely would probably be nicer from a new user
perspective :wink:

What do you think @natefaubion, @thomashoneyman, @JordanMartinez, @Benjmhart, or anyone else out there with an interest?

### DIFF
--- a/examples/ace/src/Container.purs
+++ b/examples/ace/src/Container.purs
@@ -2,7 +2,6 @@ module Example.Ace.Container where
 
 import Prelude
 
-import Data.Maybe (Maybe(..))
 import Data.Symbol (SProxy(..))
 import Effect.Aff.Class (class MonadAff)
 import Example.Ace.AceComponent as AceComponent
@@ -45,12 +44,12 @@ render { text: text } =
     , HH.div_
         [ HH.p_
             [ HH.button
-                [ HE.onClick \_ -> Just ClearText ]
+                [ HE.onClick \_ -> ClearText ]
                 [ HH.text "Clear" ]
             ]
         ]
     , HH.div_
-        [ HH.slot _ace unit AceComponent.component unit (Just <<< HandleAceUpdate) ]
+        [ HH.slot _ace unit AceComponent.component unit HandleAceUpdate ]
     , HH.p_
         [ HH.text ("Current text: " <> text) ]
     ]
--- a/examples/basic/src/Button.purs
+++ b/examples/basic/src/Button.purs
@@ -2,7 +2,6 @@ module Example.Basic.Button (component) where
 
 import Prelude
 
-import Data.Maybe (Maybe(..))
 import Halogen as H
 import Halogen.HTML as HH
 import Halogen.HTML.Events as HE
@@ -30,7 +29,7 @@ render state =
   in
     HH.button
       [ HP.title label
-      , HE.onClick \_ -> Just Toggle
+      , HE.onClick \_ -> Toggle
       ]
       [ HH.text label ]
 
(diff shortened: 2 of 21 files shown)


thomashoneyman:
I'm speaking from a narrow experience, but I have never used `Nothing` in an event handler in component HTML and I'm in favor of this change.

I _have_ used `const Nothing` to ignore messages from a component in a slot, however, and not just during prototyping. Sometimes a component has messages as output (so `absurd` isn't in play) but I don't care about the messages in this particular instance.

The workaround here, I suppose, would be to provide a handler anyway which simply doesn't do anything? Or is there another way to throw away those messages without having to introduce a new action to handle the child component in a slot?

garyb:
Yeah, the latter case I was a little less sure about, as you have no choice about asking to subscribe to messages - you get them whether you want them or not, unlike events.

I still have this dream vision of component embedding being much more like normal elements, where inputs can be provided like properties, and messages subscribed to like events, but that's not something I'm thinking about for v6. I'd like to get it turned around relatively quickly :wink:. It's something that probably could possibly be built on top of what we have already as a proof of concept too.

Having an action that does nothing when evaluated would be the only way to handle it after this, yeah.

thomashoneyman:
Despite the minor papercut of having to add an action that does nothing if you want to ignore the messages from a component I'm still in favor of this change.

natefaubion:
I don't have a strong opinion. You could add a helper to lift actions to `Maybe Action` and ignores the Nothing case.

Benjmhart:
I don't have a strong opinion on this - if the goal is to make the API simpler, I don't think `Maybe` is a source of complexity for most users who are already committing to purescript/halogen,  but I'm willing to accept the possibility of being wrong on this.

thomashoneyman:
I agree that this isn't a source of real complexity for users. But it does feel like a place where there's (admittedly a small amount of) unnecessary extra work on the part of the user in order to support what feels to me like an edge case that can be achieved through other methods.

I'm in favor because it seems like a nice, small way to improve the ergonomics of something Halogen users do all the time with little to no downside. While small, things like this and dropping the `surface` parameter all contribute to a feeling of overall polish.

garyb:
Yeah, ☝️ that was my feeling on it too, not necessarily optimising for new users but trying to remove a little friction from the most common use cases.

I'll go ahead and merge this then. We can always revert it or reconsider the case for component messages if it becomes annoying before v6 is release. I'm definitely going to try using it a bit before releasing it. 🙂

JordanMartinez:
I know it's already too late to comment on this, but I did want to echo what Thomas originally said:

> I have never used Nothing in an event handler in component HTML and I'm in favor of this change.

I've only used it in the `learn-halogen` repo because I was explaining how it works.

garyb:
Ah great, thanks for the data point!

PR #641, opened 2020-02-11
More comments describing parameters

milesfrain:
Filled in some missing documentation on parameters to make IDE tooltips more helpful for beginners.
Here's a screenshot of what things looked like before.

halogenm_tooltip

I'd like some feedback about what to put for the `a` parameter of `HalogenM`. I copied from [this part](https://github.com/purescript-halogen/purescript-halogen/blob/master/docs/2%20-%20Defining%20a%20component.md#evaluating-actions) of the guide.
```
`a` is "an underlying value of unit". Just populate with `Unit` ?
```

### DIFF
--- a/src/Halogen/Query/HalogenM.purs
+++ b/src/Halogen/Query/HalogenM.purs
@@ -73,8 +73,8 @@ instance functorHalogenF :: Functor m => Functor (HalogenF state action slots ou
 -- | - `output` is the type of messages the component can raise
 -- | - `m` is the effect monad used during evaluation
 -- | - `a` is the result of the HalogenM expression. Use the following pattern:
--- |     `handleAction :: forall st ac sl o m. ac -> H.halogenM st ac sl o m Unit
--- |     `handleQuery  :: forall st ac sl o m a. YourQuery a -> H.halogenM st ac sl o m (Maybe a)
+-- |     `handleAction :: forall st ac sl o m.            ac -> H.HalogenM st ac sl o m Unit
+-- |     `handleQuery  :: forall st ac sl o m a. YourQuery a -> H.HalogenM st ac sl o m (Maybe a)
 newtype HalogenM state action slots output m a = HalogenM (Free (HalogenF state action slots output m) a)
 
 derive newtype instance functorHalogenM :: Functor (HalogenM state action slots output m)
--- a/src/Halogen/Query/HalogenM.purs
+++ b/src/Halogen/Query/HalogenM.purs
@@ -31,6 +31,14 @@ import Prim.Row as Row
 import Web.DOM (Element)
 
 -- | The Halogen component eval algebra.
+-- |
+-- | - `state` is the component's state
+-- | - `action` is the type of actions; messages internal to the component that
+-- |   can be evaluated
+-- | - `slots` is the set of slots for addressing child components
+-- | - `output` is the type of messages the component can raise
+-- | - `m` is the effect monad used during evaluation
+-- | - `a` is "an underlying value of unit". Just populate with `Unit` ?
 data HalogenF state action slots output m a
   = State (state -> Tuple a state)
   | Subscribe (SubscriptionId -> ES.EventSource m action) (SubscriptionId -> a)
@@ -57,6 +65,14 @@ instance functorHalogenF :: Functor m => Functor (HalogenF state action slots ou
     GetRef p k -> GetRef p (f <<< k)
 
 -- | The Halogen component eval effect monad.
+
+-- | - `state` is the component's state
+-- | - `action` is the type of actions; messages internal to the component that
+-- |   can be evaluated
+-- | - `slots` is the set of slots for addressing child components
+-- | - `output` is the type of messages the component can raise
+-- | - `m` is the effect monad used during evaluation
+-- | - `a` is "an underlying value of unit". Just populate with `Unit` ?
 newtype HalogenM state action slots output m a = HalogenM (Free (HalogenF state action slots output m) a)
 
 derive newtype instance functorHalogenM :: Functor (HalogenM state action slots output m)
--- a/src/Halogen/Query/HalogenM.purs
+++ b/src/Halogen/Query/HalogenM.purs
@@ -38,7 +38,7 @@ import Web.DOM (Element)
 -- | - `slots` is the set of slots for addressing child components
 -- | - `output` is the type of messages the component can raise
 -- | - `m` is the effect monad used during evaluation
--- | - `a` is "an underlying value of unit". Just populate with `Unit` ?
+-- | - `a` is the result of the HalogenF expression. See HalogenM for usage example.
 data HalogenF state action slots output m a
   = State (state -> Tuple a state)
   | Subscribe (SubscriptionId -> ES.EventSource m action) (SubscriptionId -> a)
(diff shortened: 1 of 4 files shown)


src/Halogen/Query/HalogenM.purs
```diff
@@ -57,6 +65,14 @@ instance functorHalogenF :: Functor m => Functor (HalogenF state action slots ou
...
 
 -- | The Halogen component eval effect monad.
+
+-- | - `state` is the component's state
+-- | - `action` is the type of actions; messages internal to the component that
+-- |   can be evaluated
+-- | - `slots` is the set of slots for addressing child components
+-- | - `output` is the type of messages the component can raise
+-- | - `m` is the effect monad used during evaluation
+-- | - `a` is "an underlying value of unit". Just populate with `Unit` ?
```
garyb: I think the docs are giving the wrong impression somewhere. `a ~ Unit` only in some situations, and it can't be fixed to `Unit` in general.

For evaluation to work, a natural transformation (`f ~> g`, or `forall a. f a -> f g`) is needed, as the `a` in question changes with each step of the eval. The code written for `eval` looks like a sequence of instructions thanks to `do`, etc. but it's stored as a data structure (`Free (HalogenF ...)`) which later is folded to actually run the actions and produce a result. During this folding process, the `a` value will change as each part of the data structure is consumed. 

A perhaps simpler way of looking at it would be to consider this piece of code from an `handleAction`:

``` purescript
  CheckButtonState -> do
    buttonState <- H.query _button unit $ H.request Button.IsOn
    H.modify_ (_ { buttonState = buttonState })
```

For the `buttonState` binding, `a` is `Maybe Boolean`, for the `H.modify_` line it's `Unit`.

You can actually see that if you examine the type of the expressions too:

``` purescript
    buttonState <- (H.query _button unit $ H.request Button.IsOn) :: ?expr1
    (H.modify_ (_ { buttonState = buttonState })) :: ?expr2
```

```
[1/2 HoleInferredType] examples\components\src\Container.purs:64:70

  64      buttonState <- (H.query _button unit $ H.request Button.IsOn) :: ?expr1
                                                                           ^^^^^^

  Hole 'expr1' has the inferred type

    HalogenM
      { buttonState :: Maybe Boolean
      , toggleCount :: Int
      }
      Action
      ( button :: Slot Query Message Unit
      )
      o0
      m1
      (Maybe Boolean)

  in value declaration handleAction

  where m1 is a rigid type variable
          bound at (line 60, column 16 - line 65, column 60)
        o0 is a rigid type variable
          bound at (line 60, column 16 - line 65, column 60)

[2/2 HoleInferredType] examples\components\src\Container.purs:65:54

  65      (H.modify_ (_ { buttonState = buttonState })) :: ?expr2
                                                           ^^^^^^

  Hole 'expr2' has the inferred type

    HalogenM
      { buttonState :: Maybe Boolean
      , toggleCount :: Int
      }
      Action
      ( button :: Slot Query Message Unit
      )
      o0
      m1
      Unit

  in the following context:

    buttonState :: Maybe Boolean

  in value declaration handleAction

  where m1 is a rigid type variable
          bound at (line 60, column 16 - line 65, column 60)
        o0 is a rigid type variable
          bound at (line 60, column 16 - line 65, column 60)
```

milesfrain: Thanks for this explanation and example. Is there a situation where the type of `a` of the `HalogenM` type returned by `handleAction` could ever be something other than `Unit`?

I'm still struggling to find the phrasing that I would use to describe what `a` is in the docstring. Do you have a suggested short description, and would it be accurate for both `HalogenM` and `HalogenF`?

Here's the [line of the documentation](https://github.com/purescript-halogen/purescript-halogen/blame/master/docs/2%20-%20Defining%20a%20component.md#L192) where I pulled my original description from. Unfortunately you can only link to readme lines in the blame viewer.

garyb: For `handleAction` it will always be `Unit`, because there's no value in it returning anything other than that - the result isn't accessible anywhere.

`handleQuery` is `Query a -> HalogenM ... (Maybe a)`, the `a` is in the result here because queries can request information from the component, and the `Maybe` is in the result because there are situations where a component is unable to answer a query. 

I also don't have a particularly good way of explaining `a` in a way that won't just confuse matters further :wink:. I guess you could say it's the result/return of a HalogenM expression, but I don't know if that's a helpful thing to say really?

I think I'll change that line in the docs, probably just to remove the mention of `unit` there since I'm not sure it's helpful information to talk about at that point anyway - we can bring it up in the explanation of `handleAction` instead or something.

milesfrain: This seems similar to the confusion around the `a` in `Query a`. I still don't have a good handle on that either. My takeaway for now is to just use `Maybe a` for `query` and `Unit` for `tell` and not get too stuck trying to reason through it during this early stage of learning.
I think for the sake of making things easiest for new users, the `HalogenM` docstring for `a` should just say to make this `Maybe a` for `handleQuery` and `Unit` for `handleAction`.
Interesting parallel between these two. I assume that's not a coincidence:
* `tell` and `handleAction` : use `Unit`
* `query` and `handleQuery` : use `Maybe a`

garyb: > This seems similar to the confusion around the `a` in `Query a`

Yeah, it's exactly that again. 🙂 I'll try and write something outside of this PR/issue to explain that and ping you on it to see if I can successfully clear it up.

I think something like that you're proposing is a decent compromise - I'd perhaps still say something like `a` is the "result", but giving the concrete examples of how it is used for `handleAction` and `handleQuery`.

The reason I like mentioning the result is there are other functions that halogen provides that operate in `HalogenM` and have other values for the `a`, such as:

- `subscribe`: https://github.com/purescript-halogen/purescript-halogen/blob/337d450fdd5e18de6f75920f808c34a173ff8d88/src/Halogen/Query/HalogenM.purs#L153
- `fork`: https://github.com/purescript-halogen/purescript-halogen/blob/337d450fdd5e18de6f75920f808c34a173ff8d88/src/Halogen/Query/HalogenM.purs#L196
- `getHTMLElementRef`: https://github.com/purescript-halogen/purescript-halogen/blob/337d450fdd5e18de6f75920f808c34a173ff8d88/src/Halogen/Query.purs#L87-L91
- etc.

And even things like `get` and `modify`... although I'm probably just confusing the matter here, as they're not listed explicitly in this form since they're `MonadState`-provided as like `get :: forall m s. MonadState s m => m s`

milesfrain:
The latest PR uses this docstring.
```
- `a` is the result of the HalogenM expression. Use the following pattern:
     `handleAction :: forall st ac sl o m. ac -> H.halogenM st ac sl o m Unit
     `handleQuery  :: forall st ac sl o m a. YourQuery a -> H.halogenM st ac sl o m (Maybe a)
```
Open to other suggestions for single-character type-variable names. The collisions of `s` and `a` are unfortunate.

src/Halogen/Query/HalogenM.purs
JordanMartinez: @milesfrain I think to understand what `a` is requires understanding a few concepts: 
1. what the `Free` monad is and how it works
2. how the `Free` monad can be used to model an Abstract Syntax Tree (AST)
3. how a Natural Transformation from the pure side-effect-free `Free` monad to your side-effectful monad such as `Aff` (e.g. `forall a. Free AlgebraF a -> Aff a`) models how to interpret that pure side-effect-free `Free`-based AST into an impure side-effectful program that does useful things. 

Each of these things is covered in my learning repo here: https://github.com/JordanMartinez/purescript-jordans-reference/tree/latestRelease/21-Hello-World/05-Application-Structure/src/03-Free. 
However, the core idea is described here: https://github.com/JordanMartinez/purescript-jordans-reference/blob/latestRelease/21-Hello-World/05-Application-Structure/src/03-Free/02-Why-Use-the-Free-Monad/07-Defining-Modular-Monads.md

milesfrain:
@garyb Can this be merged? I'm often missing these tooltips.

PR #642, opened 2020-02-12
Add a variant of slot to ignore child output

paul-rouse:
The discussion of #636 mentioned the slight inconvenience of replacing `const Nothing` when child output messages need to be ignored.  I had several of these in one of the projects I'm testing against *next-6*.

How about adding a variant of `slot` which ignores outputs?  It ccould also be used in the `Void` case, slightly simplifying things there, too.  Here's an implementation, but I'm not sure about the name `slot_`; I wanted to avoid `slot'` since that meant something different up to version 4.

I realise this makes use of the fact that the `Maybe` is still there in the underlying implementation, and you may have been intending to remove that sometime in the future, so feel free to ignore this PR if there is a better direction to go!

### DIFF
--- a/src/Halogen/HTML.purs
+++ b/src/Halogen/HTML.purs
@@ -5,6 +5,7 @@ module Halogen.HTML
   , PlainHTML
   , fromPlainHTML
   , slot
+  , slot_
   , memoized
   , module Halogen.HTML.Core
   , module Halogen.HTML.Elements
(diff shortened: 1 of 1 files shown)


garyb:
Seems reasonable to me - and I think the `slot_` name makes sense, as underscore-suffixed things often imply some kind of result is being ignored. Thanks!

PR #643, opened 2020-02-14
Update library to compile on 0.14.0 release

JordanMartinez:
Note: I'm not sure whether I converted every empty data declaration (e.g. `data Foo (a :: Type -> Type)` -> `foreign import data Foo :: (Type -> Type) -> Type`. `purs` didn't complain even when I hadn't gotten to some.^

^ At least I don't think it did.


natefaubion:
Empty data decls can be an issue when there are no kind signatures, since they are all inferred polymorphic. They won't cause problem just on their own necessarily except for pedantic reasons.

PR #644, opened 2020-02-26
Updated todo list example for v5

AndrewBrownK:


examples/todo/src/Component/Task.purs
```diff
@@ -1,71 +1,82 @@
...
 import Prelude
 
-import Control.Monad.State as CMS
 import Data.Maybe (Maybe(..))
+import Example.Todo.Model (Task, initialTask)
 import Halogen as H
 import Halogen.HTML as HH
 import Halogen.HTML.Events as HE
 import Halogen.HTML.Properties as HP
-import Model (Task)
```
AndrewBrownK: I used this todo example as my gateway to learning Halogen v5. The one thing I'm feeling uncertain on is how to deal with the situation where you want a query AND action for the same thing. In this case, that is "SetCompleted". I just duplicated the command and appended a ' in this case. This may or may not be idiomatic, I have no idea. It's not perfectly DRY, but neither is it terrible repetition either. Granted, it might not be terrible only because it's a tiny example.

examples/todo/src/Component/Task.purs
```diff
@@ -0,0 +1,82 @@
...
+-- | The messages the task component might send to a parent component.
+data TaskMessage
+  = NotifyRemove
+  | Toggled Boolean
+
+-- | The task component actions (arising from the rendered HTML).
+data TaskAction
+  = UpdateDescription String
+  | SetCompleted' Boolean
+  | Remove
```
AndrewBrownK: I used this todo example as my gateway to learning Halogen v5. The one thing I'm feeling uncertain on is how to deal with the situation where you want a query AND action for the same thing. In this case, that is "SetCompleted". I just duplicated the command and appended a ' in this case. This may or may not be idiomatic, I have no idea. It's not perfectly DRY, but neither is it terrible repetition either. Granted, it might not be terrible only because it's a tiny example.

MonoidMusician: What I usually do is add a constructor to the query that can execute any action: `QueryAction TaskAction a`. Then you can call `handleAction` from `handleQuery` and run any action from the outside. But it looks like you do slightly different things in each case (raise a notification or not), so maybe you still want to keep them separate.

examples/todo/src/Component/List.purs
```diff
@@ -1,79 +1,84 @@
...
-import Model (List, TaskId, initialList, initialTask)
-import Component.Task (TaskQuery(..), TaskMessage(..), task)
 
--- | The list component query algebra.
-data ListQuery a
-  = NewTask a
-  | AllDone a
-  | HandleTaskMessage TaskId TaskMessage a
+-- | The list component query algebra (queried from the outside).
+type ListQuery a = Unit
```
MonoidMusician: Two things: if you want to use this as a query type, it has to actually be a data type or newtype (like `type ListQuery = Const Void` or `Const Unit`), since type synonyms can't be partially applied. But since you aren't using it anywhere, maybe it is best to leave it off entirely.

AndrewBrownK: Great catch. Thank you. Will fix in next commit

examples/todo/src/Component/Task.purs
AndrewBrownK: This makes great sense. Will add to next commit.

thomashoneyman:
@AndrewBrownK Sorry I'm just getting to this and thank you for the update!

The todo list example was removed in Halogen 5 because it doesn't represent idiomatic use of the library (an ordinary Halogen application wouldn't make so many tiny components). It was intended as a more involved example for demonstration purposes, but once [Real World Halogen](https://github.com/thomashoneyman/purescript-halogen-realworld) was released that was no longer necessary here in the repository.

For that reason I don't think we want to re-add the example to the repository. That said, I appreciate the time you invested to update the example for Halogen 5 -- thank you!

PR #653, opened 2020-05-08
Migrate project to Spago

thomashoneyman:
This PR makes the following updates:

1. Switches the Halogen repository to use Spago as the package manager, including switching all examples to be Spago-based. With this update it also removes all references to Bower.
2. Uses a new Bower file generated by Spago. This is the same as the old one, as far as Pursuit / publishing is concerned, but with one change: it now includes `effect` explicitly as a dependency. It was already included transitively via several other dependencies, so including it isn't really a change.
3. Updates the Travis file to always point at the latest version of the PureScript compiler as is done in the `purescript-contrib` libraries.
4. Deletes the GUIDE.md link, as the /docs folder has been around for years now and there are no links to this file within the repository.

There are no changes to the underlying library or other code outside of simple tweaks needed to make sure everything's up to date. Just a simple transition from Bower -> Spago.

I verified that the new Bower file includes dependencies at the same major versions as the current Bower file.


package.json
```diff
@@ -1,28 +1,29 @@
...
+    "example-effects-aff-ajax": "spago --config examples/effects-aff-ajax/spago.dhall bundle-app --main Example.Effects.Aff.Ajax.Main --to examples/effects-aff-ajax/dist/example.js",
+    "example-effects-effect-random": "spago --config examples/effects-effect-random/spago.dhall bundle-app --main Example.Effects.Effect.Random.Main --to examples/effects-effect-random/dist/example.js",
+    "example-higher-order-components": "spago --config examples/higher-order-components/spago.dhall bundle-app --main Example.HOC.Main --to examples/higher-order-components/dist/example.js",
+    "example-interpret": "spago --config examples/interpret/spago.dhall bundle-app --main Example.Interpret.Main --to examples/interpret/dist/example.js",
+    "example-keyboard-input": "spago --config examples/keyboard-input/spago.dhall bundle-app --main Example.KeyboardInput.Main --to examples/keyboard-input/dist/example.js",
+    "example-lifecycle": "spago --config examples/lifecycle/spago.dhall bundle-app --main Example.Lifecycle.Main --to examples/lifecycle/dist/example.js"
   },
   "devDependencies": {
-    "pulp": "^13.0.0",
-    "purescript": "^0.13.0",
```
garyb, line 24: I think I'd rather keep PS as an explicit dependency rather than grabbing the latest. 

Although we've just switched over all the core/contrib repos to grab the latest compiler, that was kinda leftover from the days where we weren't free to make an npm release whenever we wanted. I think depending on a specific compiler version is preferable for libraries, rather than always grabbing the latest, since it might well be that there's a window where we're not compatible with the latest compiler. It can especially happen in projects with lots of dependencies, since everything upstream needs to be ready also.

thomashoneyman: Works for me! I've added PureScript back as a dev dependency and reverted that part of the Travis script.

PR #655, opened 2020-05-21
Remove use of Void from PlainHTML

thomashoneyman:
The `PlainHTML` type can be used for HTML that doesn't have any widgets or actions:

```purs
module Halogen.HTML where

type PlainHTML = HTML Void Void
```

It can be used, for example, like this:

```purs
import Halogen.HTML as HH

element :: HH.PlainHTML
element = HH.text "Hello, world"
```

Unfortunately, if you try to use this with other HTML types, like in a component that does have actions, then you get an error:

```purs
-- ERROR: Cannot match type `Void` with type `Action`
render :: forall m. State -> H.ComponentHTML Action () m
render _ = HH.button [ HE.onClick \_ -> Just Toggle ] [ element ]
```

To solve this, there's a helper function called `fromPlainHTML`:

```purs
module Halogen.HTML where

fromPlainHTML :: forall w i. PlainHTML -> HTML w i
```

which can be used to fix our problem:

```purs
render _ = HH.button [ HE.onClick \_ -> Just Toggle ] [ HH.fromPlainHTML element ]
```

But this introduces extra ceremony around a useful type, `PlainHTML`, which could be avoided if we used an inner `forall` instead of `Void`:

```purs
type PlainHTML = forall w i. HTML w i
```

With this our plain HTML element looks identical, but we get to skip calls to fromPlainHTML later:

```purs
module Halogen.HTML where

element :: HH.PlainHTML
element = HH.text "Hello, world"

-- No longer an error: `element` is fine.
render :: forall m. State -> H.ComponentHTML Action () m
render _ = HH.button [ HE.onClick \_ -> Just Toggle ] [ element ]
```

This PR introduces this change, which includes removing the `fromPlainHTML` function. But if we'd prefer not to have a breaking change there we could also keep `fromPlainHTML` in place, though it would no longer be necessary.

### DIFF
--- a/src/Halogen/HTML.purs
+++ b/src/Halogen/HTML.purs
@@ -3,7 +3,6 @@
 module Halogen.HTML
   ( ComponentHTML
   , PlainHTML
-  , fromPlainHTML
   , slot
   , memoized
   , lazy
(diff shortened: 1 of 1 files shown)


natefaubion:
The problem you might run into is that if you accept `PlainHTML` as an argument, you will incur a rank-2 type, which can lead to confusing errors.

natefaubion:
For example:
```
renderThis :: PlainHTML -> HTML ...
-- This won't compile due to rank-2 and $
renderThis $ HH.div ...
```

thomashoneyman:
Yes, that's a bit confusing.

```purs
-- ERROR: The type variable `w` has escaped its scope...
element :: (PlainHTML -> ComponentHTML Action () Aff) -> ComponentHTML Action () Aff
element render = render $ HH.text "hello"

-- Compiles
element :: (PlainHTML -> ComponentHTML Action () Aff) -> ComponentHTML Action () Aff
element render = render (HH.text "hello")
```

That's not pleasant.

garyb:
Yeah, that's why I went with `Void`. 🤔

thomashoneyman:
Closing this with that in mind, though it's a bit unfortunate.

garyb:
fake edit: You closed it while I was adding more thoughts, but it might be useful anyway:

I think I'd probably prefer to stick with `Void` and the coercion function. 

The use case I had in mind for `PlainHTML` is where you want to accept HTML as value, which is the exact situation that this error gets triggered in. I would usually type my plain html values as `HTML w i`, since they can be freely used anywhere then, but I would use `PlainHTML` when it's a component configuration option or something like that.

It's perhaps worth adding more commentary explaining this at least 🙂

Issue #668, opened 2020-06-09
not very pretty solution: appendChild is called thought there is no need to call it except for root component


srghma:
I have noticed that this line is called constantly thought there is no need to call it except for the root component (because yes, it should be appended to the <body> for example)

https://github.com/purescript-halogen/purescript-halogen/blob/807ba2b49641bbe0fff4ca66e2068969be709d7d/src/Halogen/VDom/Driver.purs#L165

First, to understand, we should know how `.appendChild` and `.insertBefore` works - if node being appended IS ALREADY appended elsewhere - it is first removed from current parent AND appended to the new parent. Actually even if it is already appended to the same parent - it removed and reappended (and rerender is triggered I think)

Now, example:

1. let's imagine we have the following component structure

```
component A
component B inside of component A
```

2. on first render it is rendered as following

```
Halogen.Aff.Driver.runComponent (from runUI) is called for A
Halogen.VDom.Driver.render (from renderSpec) is called for A
new VdomSpec is created for component A using mkSpec (this is too not very pretty, that there is a separate spec for each component, but it's a different story, maybe it's fine)
buildVDom is called for A

inside of buildVDom:
  Halogen.Aff.Driver.runComponent (from runUI) is called for B
  Halogen.VDom.Driver.render (from renderSpec) is called for B
  new VdomSpec is created for component B using mkSpec
  buildVDom is called for B
  new node is created, non-appended everywhere
  `DOM.appendChild node (HTMLElement.toNode container)` appends `B node` to `<body>` (!!!!)
  Halogen.VDom.Driver.render returns

Halogen.Aff.Driver.renderChild function returns
the halogen-vdom buildElem function inserts child nodes to parent node using Util.insertChildIx. 
  What exactly it does: removes `B node` from `<body>` and appends to the `Component A` (!!!)

component A node is created now with children inside
`DOM.appendChild node (HTMLElement.toNode container)` appends `A node` to `<body>`, as should
```

![2020-06-09-20:45:29-screenshot](https://user-images.githubusercontent.com/7573215/84182245-c416d800-aa92-11ea-9f47-cafdbd9482ff.png)
![2020-06-09-21:06:24-screenshot](https://user-images.githubusercontent.com/7573215/84183786-1b1dac80-aa95-11ea-8dc1-d31d71cc1cb6.png)
[...]

PR #670, opened 2020-06-10
fix(#668): appendChild is called thought there is no need to call it except for root component

srghma:
2020-06-10-13:53:04-screenshot
2020-06-10-13:53:09-screenshot


thomashoneyman:
@natefaubion @garyb I'm not familiar with this part of the code; do either of you have time for feedback wrt this change? If not, I can also take a little time to familiarize myself and then provide a review.

srghma:
@thomashoneyman tnx for pushing it

NB: as an alternative approach we can split the `render` function from spec on `renderChildComponent` and `renderRootComponent` (this would allow to add a check in tests on `renderRootComponent` that it should be called only once per test for example)

PR #671, opened 2020-06-10
Hydration

srghma:
It works, but the refactoring is needed to remove duplicated code and fix some todos

I didnt test on all examples, but I have tested on `components-inputs` example, and it's already something

I'm using this repo for rendering the page

https://github.com/srghma/purescript-halogen-nextjs/blob/d9c22d07c35c1cf1d821cf0ecfdfed44e9dd91e9/app/Server.purs#L84-L120
https://github.com/srghma/purescript-halogen-nextjs/blob/d9c22d07c35c1cf1d821cf0ecfdfed44e9dd91e9/app/Client.purs#L19-L24

Also, I'm using this function to link local repos

```
link-purescript-project () {
  from="$1"
  to="$2"
  version="${3:-master}"
  echo "linking $from to $to (version = $version)"
  rmf "$from" && mkdir -p "$from" && ln -s "$to" "$from/$version"
}

link-purescript-project $HOME/projects/purescript-halogen/.spago/halogen-vdom $HOME/projects/purescript-halogen-vdom "v6.1.3"
link-purescript-project $HOME/projects/purescript-halogen-nextjs/.spago/halogen-vdom $HOME/projects/purescript-halogen-vdom
link-purescript-project $HOME/projects/purescript-halogen-nextjs/.spago/halogen $HOME/projects/purescript-halogen
link-purescript-project $HOME/projects/purescript-halogen-realworld/.spago/halogen $HOME/projects/purescript-halogen "v5.0.0-rc.9"
```

2020-06-10-20:21:02-screenshot
2020-06-10-20:21:05-screenshot
2020-06-10-20:25:20-screenshot


srghma:
@garyb hi, this week I think I'm gonna finish the hydration, do you think it's better to rebase work on halogen-6 branch?

garyb:
I don't think there'll be much conflict between h5 and h6 for this, since most h6 changes so far are in the component API / eval for event sources.

Also, just as a warning in advance, I'm probably not going to have time to review or test this for a while. Sorry 😕.

srghma:
Will rebase.

That's hard to hear, maybe we can find some solution: I can show in the call that it works on my project or write a video,

examples/ace/spago.dhall
```diff
@@ -1,6 +1,6 @@
 let config = ../../spago.dhall
 
 in config // {
-  sources = config.sources # [ "examples/ace/**/*.purs" ],
+  sources = config.sources # [ "examples/ace/src/**/*.purs" ],
```
thomashoneyman, line 4: Why is this necessary? They appeared to be working properly before this change -- did you notice something amiss?

packages.dhall
```diff
@@ -1,7 +1,10 @@
 let upstream =
       https://github.com/purescript/package-sets/releases/download/psc-0.13.6-20200507/packages.dhall sha256:9c1e8951e721b79de1de551f31ecb5a339e82bbd43300eb5ccfb1bf8cf7bbd62
 
-let overrides = {=}
+let overrides =
+    { web-dom =
+        upstream.web-dom // { repo = "https://github.com/srghma/purescript-web-dom.git", version = "patch-1" }
```
thomashoneyman, line 6: This is OK while this is a draft PR, but I think Halogen versions should not have any overrides or additions except for libraries that are used in test code.

Let's make sure that `purescript-web-dom` gets updated before this PR merges.

spago.dhall
```diff
@@ -24,6 +24,7 @@
     , "unsafe-reference"
     , "web-file"
     , "web-uievents"
+    , "debug"
```
thomashoneyman, line 27: This should not be added to the `spago.dhall` file.

It's OK if you're using `debug` in development -- you might consider adding this to the test dependencies instead and running your build commands using that config (`spago -x test/test.dhall build`, for example). That way you can freely use `debug` but we don't inadvertently add it as a library dependency.

src/Halogen/Aff/Driver.purs
```diff
@@ -1,109 +1,47 @@
 module Halogen.Aff.Driver
-  ( RenderSpec
-  , runUI
+  ( runUI
+  , hydrateUI
+  , module Halogen.Aff.Driver.RenderImplementation
```
thomashoneyman, line 4: Can we keep the `runUI` and `hydrateUI` implementations directly in this module rather than indirectly via an "implementation" module?

In general, if code needs to be moved to another module (to avoid import cycles, or because it's internal and users should not import it directly) I would prefer that it is moved to an `Internal` namespace like `Halogen.Aff.Driver.Internal`.

In this case it doesn't seem clear to me that the code needs to be moved to another module. I'd rather just keep both implementations in this module unless you feel strongly that that isn't a good idea.

That has the added advantage of keeping the diffs a little more readable for reviewers -- if this module is expanded then we won't have to read through `runUI` code that wasn't changed, just moved, which will help us get your work reviewed sooner.

src/Halogen/Aff/Driver/HydrationImplementation.purs
```diff
@@ -0,0 +1,131 @@
...
+  -> Ref LifecycleHandlers
+  -> (act -> Aff Unit)
+  -> Ref (Slot.SlotStorage ps (DriverStateRef r))
+  -> Ref (Slot.SlotStorage ps (DriverStateRef r))
+  -> ComponentSlotBox ps Aff act
+  -> DOM.Node
+  -> Effect (RenderStateX r)
+renderChildHydrate renderSpec lchs handler childrenInRef childrenOutRef componentSlotBox currentNode =
+  unComponentSlot (\slot -> do
+    traceM { message: "Halogen.Aff.Driver.HydrationImplementation.runUI renderChild 1", slot, childrenInRef }
```
thomashoneyman, line 116: Are you still using these debug messages?

src/Halogen/VDom/Driver.purs
```diff
@@ -153,26 +241,56 @@ renderSpec document container =
...
     -> Effect (RenderState state action slots output)
-  render handler child (HTML vdom) =
+  render handler renderChild (HTML vdom) isRoot =
     case _ of
       Nothing -> do
-        renderChildRef <- Ref.new child
-        let spec = mkSpec handler renderChildRef document
+        traceM { message: "Halogen.VDom.Driver.renderSpec render", handler, renderChild, vdom, container }
+        renderChildRef <- Ref.new renderChild
+        let renderChildHydrate = \_componentSlotBox _element -> throw "should not be called" -- TODO!!!!!
```
thomashoneyman, line 252: Do you mind explaining this TODO?

thomashoneyman:
I'm not well-versed in server-side rendering, so I'd like to just take a moment and describe my understanding of hydration. If I'm making mistakes here, please feel free to correct my understanding so that I can better review what you're doing here.

It looks like you're introducing a new `hydrateUI` function in addition to the existing `runUI` function. This function is meant to cover the case where some content has been pre-rendered on the server and shipped to the client. When using server-side rendering you want to just attach event listeners and so on to the existing HTML in the DOM, whereas when rendering everything client-side you want to produce the markup and attach event listeners. "Hydration" then refers to attaching event listeners etc. to the existing HTML -- essentially, the first render has already occurred.

I'm using this understanding to review your `hydrateUI` function and the related changes you've made to the virtual DOM driver module, but let me know if I'm missing something.

examples/ace/spago.dhall
srghma: Yes, i have mentioned in commit message

https://github.com/purescript-halogen/purescript-halogen/commit/aa6d9118592de300f05dd40282f20321b5755a2a

src/Halogen/Aff/Driver.purs
srghma: I'll check, the code between renderImpl and hydrateImpl will be shared

src/Halogen/VDom/Driver.purs
srghma: To reduce number of possible errors / increase totality I was thinking about splitting vdomSpec (currently has rendering and hydration functionlity) to vdomSpec (only render) and hydrateVdomSpec (render, because it's used after hydration, and hydrate functionality)

srghma:
> When using server-side rendering you want to just attach event listeners and so on to the existing HTML in the DOM

right, that's one of 2 changes to the DOM that occur during hydration phase 

1. the event handlers are added here https://github.com/srghma/purescript-halogen-vdom/blob/fd94726ea7314f6f4c13bd77b8233dfe2fc3c67f/src/Halogen/VDom/DOM/Prop/Implementation.purs#L90-L95

2.

the second change is that the Text nodes are being split ([zips children (e.i. DOM and VDOM) of Elem and Keyed with side effect of splitting Text nodes](https://github.com/srghma/purescript-halogen-vdom/blob/fd94726ea7314f6f4c13bd77b8233dfe2fc3c67f/src/Halogen/VDom/DOM/Util.purs#L41-L64) , [2](https://github.com/srghma/purescript-halogen-vdom/blob/fd94726ea7314f6f4c13bd77b8233dfe2fc3c67f/src/Halogen/VDom/DOM/Elem.purs#L62), [3](https://github.com/srghma/purescript-halogen-vdom/blob/fd94726ea7314f6f4c13bd77b8233dfe2fc3c67f/src/Halogen/VDom/DOM/Keyed.purs#L56))

srghma:
I'm currently working on splitting js code per page in the `purescript-halogen-nextjs` project and preloading page js files when link to that page appears on viewport

but I would rename it to `purescript-halogen-isomorphic` and , if anyone interested, I could handle it to the `purescript-halogen` community

src/Halogen/Aff/Driver.purs
srghma: I have moved and split code to the Implementation directory

srghma:
@garyb @natefaubion if I understand correctly you are working on a new purescript release, but could you give at least a tentative timeframe when you are going to look at this pr?

.gitignore
```diff
@@ -11,3 +11,4 @@ node_modules
 output
 
 example/example.js
+github_deploy_key
```
thomashoneyman, line 14: Would you mind rebasing this PR so that changes unrelated to your work are not included?

src/Halogen/Aff/Driver.purs
```diff
@@ -1,296 +1,98 @@
...
 import Halogen (HalogenIO)
 import Halogen.Aff.Driver.Eval as Eval
-import Halogen.Aff.Driver.State (DriverState(..), DriverStateRef(..), DriverStateX, LifecycleHandlers, RenderStateX, initDriverState, mapDriverState, renderStateX, renderStateX_, unDriverStateX)
-import Halogen.Component (Component, ComponentSlot, ComponentSlotBox, unComponent, unComponentSlot)
-import Halogen.Data.Slot as Slot
-import Halogen.HTML.Core as HC
-import Halogen.Query.HalogenQ as HQ
-import Halogen.Query.Input (Input)
-import Halogen.Query.Input as Input
+import Halogen.Aff.Driver.Implementation.Hydrate as Hydrate
```
thomashoneyman, line 20: The `Implementation` modules may be a good idea in the future -- some refactoring certainly makes sense with this PR -- but in this pull request they make it difficult to see precisely what is changing. The best way to make sure the PR is reviewed is to make changes that produce the minimum diff necessary. That makes it easy on @garyb and @natefaubion to take a look and see exactly what's changed.

I understand why you'd want to do this refactoring as well, and once the main PR has been merged I think we'd all be fine with applying it afterwards.

.gitignore
srghma: this came from current master https://github.com/srghma/purescript-halogen/commit/d2f59b990a51152ca6cd53dee4dcd295f85d5a2d

you see it in this pr becuase I'm trying to merge to next-6 branch, that is not rebased on current master

I have merged halogen/master to my branch

this should disappear once halogen/next-6 branch is rebased on halogen/master

src/Halogen/Aff/Driver.purs
srghma: :( that would be hard for me to do

srghma:
@thomashoneyman could you please explain why closed?

thomashoneyman:
I'm sorry, @srghma, I didn't intend to close this pull request. The `next-6` branch has now merged to `master` and has been removed; as with the open portals PR I expected to see something like this:

Expected

Please feel free to reopen this pull request against the `master` branch of this library and it should be fine.

PR #683, opened 2020-07-08
docs: spell out type signatures + fixup

toastal:
While rediscovering Halogen after being in Elm land for the last two-ish year, I was happy to see that in the official Halogen docs on Pursuit that that the types are spelled out so it's easy for a newcomer to follow. The fact that these docs are out of sync with that philosophy seems like a error.

Also, while editing these type signatures, I found some typos as well.

I'd also reckon a new issue should be raised that `*.md` files should be accounted for in the `.editorconfig`.


thomashoneyman:
Oh — also, I’m happy to add .md files to the editorconfig.

toastal:
Yeah, I think as you get familiar with the conventions of what the single characters mean, it's extra typing, but for a beginner I think it helps ground them and matches the documentation.

PR #684, opened 2020-07-10
Instructions for adding missing properties / attributes

milesfrain:
These instructions shouldn't be required, but they're good to have until we can ensure all properties are accounted for.

### DIFF
--- a/docs/guide/01-Rendering-Halogen-HTML.md
+++ b/docs/guide/01-Rendering-Halogen-HTML.md
@@ -197,3 +197,29 @@ html = HH.div [ HP.placeholder "blah" ] [ ]
 ```
 
 This error tells you that you've tried to use a property with an element that doesn't support it. It first lists the property you tried to use, and then it lists the properties that the element _does_ support. Another example of Halogen's type safety in action!
+
+### Adding missing properties
+
+Not all properties are currently available in Halogen.
+
+For example, you may try to write:
+```purs
+html = HH.iframe [ HP.sandbox "allow-scripts" ]
+```
+Only to receive this error:
+```
+Unknown value HP.sandbox
+```
+Even though it seems like this property should be [supported](https://pursuit.purescript.org/packages/purescript-dom-indexed/docs/DOM.HTML.Indexed#t:HTMLiframe):
+```purs
+type HTMLiframe = Noninteractive (height :: CSSPixel, name :: String, onLoad :: Event, sandbox :: String, src :: String, srcDoc :: String, width :: CSSPixel)
+```
+The solution is to write your own implementation of this missing property:
+```purs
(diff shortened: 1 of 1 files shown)


milesfrain:
Moved discussion from original PR comment to https://discourse.purescript.org/t/autogenerate-libraries-from-html-spec/1553

docs/guide/01-Rendering-Halogen-HTML.md
```diff
@@ -197,3 +197,29 @@ html = HH.div [ HP.placeholder "blah" ] [ ]
 ```
 
 This error tells you that you've tried to use a property with an element that doesn't support it. It first lists the property you tried to use, and then it lists the properties that the element _does_ support. Another example of Halogen's type safety in action!
+
+### Adding missing properties
+
+Not all properties are currently available in Halogen.
```
thomashoneyman, line 203: I think we should explain that HTML is a [living standard](https://html.spec.whatwg.org/multipage), which means that the standard continues to change over time and Halogen has to keep up with those changes. That means that Halogen and Web.HTML can end up a bit behind the standard sometimes and need to be updated to match it. That's the main reason why not all properties might be in the library.

Then we can clarify what you wrote:

docs/guide/01-Rendering-Halogen-HTML.md
```diff
@@ -197,3 +197,29 @@ html = HH.div [ HP.placeholder "blah" ] [ ]
...
+```
+The solution is to write your own implementation of this missing property:
+```purs
+sandbox :: forall r i. String -> HH.IProp ( sandbox :: String | r ) i
+sandbox = HH.prop (HH.PropName "sandbox")
+```
+Then you can use it in your HTML element:
+```purs
+html = HH.iframe [ sandbox "allow-scripts" ]
+```
```
thomashoneyman, line 225: I'd also suggest to readers that they ought to open an issue or PR so we can add the missing property! It's an easy way they can contribute to Halogen.

docs/guide/01-Rendering-Halogen-HTML.md
```diff
@@ -197,3 +197,29 @@ html = HH.div [ HP.placeholder "blah" ] [ ]
 ```
 
 This error tells you that you've tried to use a property with an element that doesn't support it. It first lists the property you tried to use, and then it lists the properties that the element _does_ support. Another example of Halogen's type safety in action!
+
+### Adding missing properties
+
+Not all properties are currently available in Halogen.
+
+For example, you may try to write:
+```purs
```
thomashoneyman, line 206: Small nit: do you mind adding newlines around code blocks? ie.

```
For example, you may try to write:

<code block>

Only to receive this error:

<code block>

Even though...
```

milesfrain:
Made suggested changes. Also asking readers for help with automation. Maybe someone will have a good idea.

Issue #689, opened 2020-07-31
textarea should be a Node, but is a Leaf


srghma:
https://github.com/purescript-halogen/purescript-halogen/blob/3ababd93bff0fe592f3d6ab29b699188dad26d86/src/Halogen/HTML/Elements.purs#L740-L741

should be changed to Node

and 

https://github.com/purescript-halogen/purescript-dom-indexed/blob/c449ae115bffe1fce89e145da62b862ad16ed935/src/DOM/HTML/Indexed.purs#L591

value should be removed

____

because it says here https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea

```
Default content entered between the opening and closing tags. <textarea> does not support the value attribute.
```

srghma:
even better the content should be a `Text`, not array of Nodes

as per https://www.w3.org/TR/html52/sec-forms.html#the-textarea-element

```
Content model:
Text.

```

garyb:
That's the `value` property, not the value attribute: https://html.spec.whatwg.org/multipage/form-elements.html#the-textarea-element

srghma:
ah, then, probably, nothing should be changed

garyb:
I think you can write the content as the body through the vdom too, so it perhaps should be a node too - I'm not really sure.

This is one of those annoying situations where I guess the markup model doesn't quite suit the JS model backing it (or vice versa). The fact the prop/attribute distinction exists at all still bothers me.

garyb:
I guess this is something that needs special handling for hydration / vdom-text-rendering.

srghma:
> The fact the prop/attribute distinction exists at all still bothers me

agree

maybe dom-indexed should split attributes and props

currenly

```
type HTMLtextarea = Interactive
  ( autofocus :: Boolean
  , cols :: Int
  , disabled :: Boolean
  , form :: String
  , maxLength :: Int
  , name :: String
  , onChange :: Event
  , onInput :: Event
  , onScroll :: Event
  , onSelect :: Event
  , placeholder :: String
  , readOnly :: Boolean
  , required :: Boolean
  , rows :: Int
  , value :: String
  , wrap :: WrapValue
  )
```

to

```
type HTML_Props_textarea = Interactive
  ( autofocus :: Boolean
  , cols :: Int
  , disabled :: Boolean
  , form :: String
  , maxLength :: Int
  , name :: String
  , onChange :: Event
  , onInput :: Event
  , onScroll :: Event
[...]

PR #690, opened 2020-07-31
fix(#689): textarea should be a Node, but is a Leaf

srghma:


srghma:
yes, content as vdom child is supported by html spec (not only property), but it's better if we use only one way of setting text body (i.e. prop)

Issue #682, opened 2020-07-06
Halogen.HTML.Properties missing style attribute


shmish111:
I have a case where I need to add height to an element so I needed to add the property `style: "height: 55px"` however I had to manually add `attr (AttrName "style") "height: 55"` as it seems there is no `style` constructor even though there is a style property.

garyb:
This was intentional: there's a typed `style` property provided by [purescript-halogen-css](https://github.com/purescript-halogen/). But yeah, that's maybe not a good enough reason to omit the `String` one.

shmish111:
Oh, I'd never seen that library. It's a bit heavy handed for my use-case though.

thomashoneyman:
Sorry, forgot to close this issue -- this was closed in #692 and #698.

PR #692, opened 2020-07-31
fix(#682): Properties -> add `style` attr function

srghma:


thomashoneyman:
@srghma I'm on board with this change. Historically Halogen has omitted the property so as to encourage use of typed css via purescript-halogen-css, but I don't think it's necessary to enforce that via omission.

Still, it is good for folks to know they _can_ use typed CSS if they would like to. Can you add a documentation comment like the one below to this property?

```
-- | Sets the `style` attribute to the specified string.
-- |
-- | ```purs
-- | ... [ style "height: 50px;" ]
-- | ```
-- |
-- | If you prefer to use typed CSS for this attribute, you can use the purescript-halogen-css library:
-- | https://github.com/purescript-halogen/purescript-halogen-css
```

@garyb let me know if you'd prefer not to mention halogen-css here; otherwise I'm happy to merge with this comment.

PR #694, opened 2020-08-06
Split queryF from query

sigilion:
In [sigilion/purescript-run-halogen](https://github.com/sigilion/purescript-run-halogen) I find it useful to have a version of query that is not lifted into free, so that I can put the resulting HalogenF under VariantF in Run. I would prefer to have this split in the halogen source, rather than maintain my own version of query.

### DIFF
--- a/src/Halogen/Query/HalogenM.purs
+++ b/src/Halogen/Query/HalogenM.purs
@@ -137,7 +137,19 @@ query
   -> slot
   -> query a
   -> HalogenM state action slots output m (Maybe a)
-query label p q = HalogenM $ liftF $ ChildQuery $ CQ.mkChildQueryBox $
+query label p q = HalogenM $ liftF $ queryF label p q
+
+-- | An alternative to `query` which does not wrap the HalogenF in free
+queryF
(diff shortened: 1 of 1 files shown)


thomashoneyman:
I’m not necessarily against splitting out queryF (and the other constructors), but I also don’t think it’s a problem to maintain this separately downstream — I think it’s pretty rare to do that and this code doesn’t really change. I might like to wait for @garyb to have a chance to weigh in.

As a side note, I’m curious — what are you using run-Halogen for? Looks interesting!

sigilion:
run-halogen was split out of a side project I wrote to learn purescript. I don't think I can open source it; it was implementing the character generation process of the old pencil and paper RPG Traveler, and I think the process is copyrighted. The basic idea was to write the "business logic" using a bunch of domain specific algebraic effects (roll a die, choose from these options, enter a string etc.) plus State. I wrote a CLI version where these interactions are built around readline and the default state implementation, and a halogen version where the interactions are each delegated to a specific widget component and the state is the halogen state, so the page renders the state changes.

PR #695, opened 2020-08-12
Switch to GitHub Actions

thomashoneyman:
This PR switches Halogen over to use GitHub Actions for CI and for the publishing workflow. This makes certain things easier to do (for example: set up a PureScript toolchain, publish the mdbook, and so on) and is integrated into GitHub's fast build machines. Accordingly, I've removed the `.travis.yml` file.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,32 @@
...
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-node@v1
+
+      - name: Install purescript-psa
+        run: npm install -g purescript-psa
+
+      - uses: thomashoneyman/setup-purescript@v1
+        with:
+          purescript: "0.13.8"
```
thomashoneyman, line 18: With this, we could technically remove the `purescript` and `spago` dependencies from `package.json` as they aren't needed in CI. But I kept them there for now for local development in case folks want them.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,32 @@
...
+
+      - name: Cache PureScript dependencies
+        uses: actions/cache@v2
+        with:
+          key: ${{ runner.os }}-spago-${{ hashFiles('**/*.dhall') }}
+          path: |
+            .spago
+            output
```
thomashoneyman, line 26: Caching the `output` directory means that the compiler will only rebuild files that have changed since the last successful cache. I've tested this in some other projects (halogen-hooks, etc.) and it seems pretty reliable -- files are rebuilt when they've changed since the last run in CI. However, we could also just cache the `.spago` directory and always compile from scratch.

thomashoneyman:
Confirmed in commit 1df3b73 that publishing to GitHub Pages is working correctly.

garyb:
Thanks!

PR #704, opened 2020-09-27
Fix guide replacing onSubmit event by button onClick

### LINKED ISSUE #703: Example effects-aff-ajax button does not work
When clicking on the button, nothing seems to happen. It looks like the onSubmit form event is not triggered, at least with firefox and chrome.

TristanCacqueray:
Fixes #703

### DIFF
--- a/docs/guide/03-Performing-Effects.md
+++ b/docs/guide/03-Performing-Effects.md
@@ -168,6 +168,7 @@ import Halogen.HTML.Properties as HP
 import Halogen.VDom.Driver (runUI)
 import Web.Event.Event (Event)
 import Web.Event.Event as Event
+import Web.UIEvent.MouseEvent (toEvent)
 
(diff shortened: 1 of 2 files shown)


garyb:
I'd like to discuss more in #703 before merging this - avoiding `onSubmit` in favour of `onClick` is worse from a UX point of view (I know it's only an example, but we should at least encourage doing things the best way), since you can't type a name in the field and hit enter without `onSubmit`.

TristanCacqueray:
Thank you for the explanation on UX implication. Switching to `draft` for now.

thomashoneyman:
Form events work now that https://github.com/purescript/trypurescript/pull/203 merged; otherwise, I agree with @garyb that `onSubmit` is the preferred method and so I'm going to close this. Thank you for providing a fix, @TristanCacqueray!

PR #705, opened 2020-10-05
Add support for portals

thomashoneyman:
[Portals](https://reactjs.org/docs/portals.html) let you render children into a DOM node anywhere in the document, not just within the tree of the current component in the hierarchy. This is useful for elements like tooltips, notifications, and modals which need to visually "break out" of their container, especially when a parent has a `z-index` style or `overflow: hidden` and you don't have an easy CSS alternative.

I originally implemented portals outside of Halogen in the [`halogen-portal`](https://github.com/thomashoneyman/purescript-halogen-portal) library. Unfortunately, that has some problems:

1. It requires the use of `runUI` and is essentially running a new Halogen application (I'm not sure if this is a _problem_, per se, but it feels a bit wonky).
2. You could originally only use components in `Aff` (due to using `runUI`), but @MonoidMusician came up with a way to allow components which can easily be interpreted to `Aff` work as well. He may be better able to speak about other limitations this imposes, especially wrt. effects from the child component bubbling back up to the parent.
3. You have to use a component (you can't just render any arbitrary children, like a thunk or some regular HTML).

This PR adds full support for portals to Halogen. It can be demoed using the new portals example code. However, it essentially just boils down to:

```purs
-- | Renders `ComponentHTML` to an arbitrary element in the DOM instead of its
-- | actual location within a component's render tree.
portal
  :: forall action slots m
   . HTMLElement
  -> ComponentHTML action slots m
  -> ComponentHTML action slots m
portal elem html = Core.widget (PortalSlot elem html)
```

which can be used like this to render normal HTML:

```purs
HH.portal targetElement do
  HH.div ...
```

or like this to render a component:

```purs
HH.portal targetElement do
  HH.slot ...
```

The code appears to be working; using the portals example, you can see that the two portals within the `Button` example are rendering outside of their parent component element:

Screen Shot 2020-10-04 at 7 04 16 PM


spago.dhall
```diff
@@ -2,29 +2,30 @@
...
+  , "nullable"
+  , "ordered-collections"
+  , "parallel"
+  , "profunctor"
+  , "transformers"
+  , "unsafe-coerce"
+  , "unsafe-reference"
+  , "web-file"
+  , "web-uievents"
+  ]
```
thomashoneyman, line 28: This just adds the `debug` package (this doesn't change what users downstream receive when installing Halogen, as that's controlled by the package sets, so this is purely an internal change).

src/Halogen/HTML.purs
```diff
@@ -96,6 +98,15 @@ slot_
...
   Core.widget (ComponentSlot (componentSlot label p component input (const Nothing)))
 
+-- | Renders `ComponentHTML` to an arbitrary element in the DOM instead of its
+-- | actual location within a component's render tree.
+portal
+  :: forall action slots m
+   . HTMLElement
+  -> ComponentHTML action slots m
+  -> ComponentHTML action slots m
+portal elem html = Core.widget (PortalSlot elem html)
```
thomashoneyman, line 108: Another thing I considered was a type signature like this:

```purs
portal
  :: forall action slots m
   . Maybe HTMLElement
  -> ComponentHTML action slots m
  -> ComponentHTML action slots m
```

where a `Nothing` would mean to use the `<body>` tag, which from my experience is by far the most common case, and a `Just htmlElement` means to use the provided element. Curious what y'all think.

thomashoneyman:
The easiest way to test this code is to check out the branch, run `npm run example-portal`, and then open the corresponding `index.html` file. I'm curious if y'all have ideas of things I should test to ensure this feature is working correctly.

src/Halogen/HTML.purs
thomashoneyman: Or, alternately, another function:

```
portal_
  :: forall action slots m
   . ComponentHTML action slots m
  -> ComponentHTML action slots m
portal_ = portalImpl Nothing

-- Allows you to create a portal to a target element
portal
  :: forall action slots m
   . HTMLElement
  -> ComponentHTML action slots m
  -> ComponentHTML action slots m
portal elem = portalImpl (Just elem) 

-- Purely internal, shared by portal_ and portal but not exported
portalImpl
  :: forall action slots m
   . Maybe HTMLElement
  -> ComponentHTML action slots m
  -> ComponentHTML action slots m
```

src/Halogen/VDom/Driver.purs
```diff
@@ -102,7 +100,11 @@ mkSpec handler renderChildRef document =
           ThunkSlot t -> do
             step' <- EFn.runEffectFn2 V.step step t
             pure $ V.mkStep $ V.Step (V.extract step') (Just step') patch done
-        _ -> EFn.runEffectFn1 render slot
+          PortalSlot elem inner -> do
+            EFn.runEffectFn1 V.halt step
+            EFn.runEffectFn2 renderPortalSlot elem inner
```
thomashoneyman, line 105: I'm somewhat guessing about what to do here; this appears to work, but I don't know if I should instead be writing something closer to the `ThunkSlot` implementation.

thomashoneyman:
@natefaubion @garyb I'd like to clean this up by converting the example to demonstrate a modal (a motivating example for this feature in the first place) and then this could use a review before we move to release Halogen 6 with PureScript 0.14.

thomashoneyman:
I've updated the example to showcase a modal, which is a typical use case for this feature. I plan to merge this into the Halogen 6 branch tonight, unless there are any objections or additional review @natefaubion @garyb.

src/Halogen/VDom/Driver.purs
natefaubion: This is going to destroy the old DOM and build it from scratch on every render.

src/Halogen/VDom/Driver.purs
```diff
@@ -117,11 +119,21 @@ mkSpec handler renderChildRef document =
       let node = getNode rsx
       pure $ V.mkStep $ V.Step node Nothing patch done
 
+    renderPortalSlot
+      :: EFn.EffectFn2
+            DOM.HTMLElement
+            (HTML (ComponentSlot slots Aff action) action)
+            (V.Step (ComponentSlot slots Aff action) DOM.Node)
+    renderPortalSlot = EFn.mkEffectFn2 \elem inner -> do
+      step <- EFn.runEffectFn1 buildThunk (Thunk.thunked (\_ _ -> false) (const inner) unit)
```
natefaubion, line 128: What is the purpose of this?

natefaubion:
I don't think we should merge this yet. There are several confusing things, and I'm not totally convinced it's actually behaving as you are expecting. I think a big issue you are having is with `WidgetState`. This is confusing because state for component slots is _not_ in the `Step` state. This is stored in a Map within a Ref as part of the internal component's state. This is so that we can relocate a Slot elsewhere in the DOM without destroying it. Absent that, it's only potentially used to track Thunk state. Portals will need to have their own state as well, however. For example, if you swap out the Node you want to mount into, you will need to cleanup the old node and remove it before rendering it into the new one. Currently this is papered over by always destroying the old node and rendering it again. So, I think to merge this we should:

* Change `WidgetState` so that it's a sum type. Something like `data WidgetState = ThunkState ... | PortalState ... | NoState`
* The PortalState will have the configured parent Node, and the inner Step for the VDom (like what Thunk does).
* In patch, we should sync up the WidgetState with the slot type, and do a proper patch (like the branch for Thunk does), otherwise it should halt the old machine and run a new one.
* The Portal machine can store `PortalState` instead of trying to conjure up a Just thunk thing.

thomashoneyman:
Thanks for the feedback, @natefaubion. I'll need to circle back to this and address your questions and suggestions.

Swordlash:
Hi, is there any progress on this PR?

thomashoneyman:
I haven't made any recent progress 😅 but I'm still interested in finishing it and I think the feature makes a good addition to Halogen.

thomashoneyman:
Well, I've had no time for this, though I hope to come back to it sometime. Closing for now!

PR #709, opened 2020-10-21
Update 02-Introducing-Components.md

alpanagi:
Unused type variable in `forall`


thomashoneyman:
It looks like this section of the documentation is using a common style where a component with no slots uses a `()` explicitly instead of a `forall slots. ...` (as is done in the render function directly below this snippet). Would you mind returning to the use of `()` and delete the `slots` variable instead? Thanks!

PR #718, opened 2020-12-17
Add scope property

pmiddend:
The [halogen guide](https://github.com/purescript-halogen/purescript-halogen/blob/master/docs/guide/01-Rendering-Halogen-HTML.md) told me to add any property that's allowed but not in halogen yet.

So here I am, with nextto no knowledge of Purescript and Halogen creating this PR. Be gentle. :)

### DIFF
--- a/src/Halogen/HTML/Properties.purs
+++ b/src/Halogen/HTML/Properties.purs
@@ -25,6 +25,7 @@ module Halogen.HTML.Properties
   , rel
   , src
   , style
+  , scope
   , target
   , title
   , download
@@ -80,6 +81,7 @@ import DOM.HTML.Indexed.MenuitemType (MenuitemType(..)) as I
 import DOM.HTML.Indexed.OnOff (OnOff(..)) as I
 import DOM.HTML.Indexed.OrderedListType (OrderedListType(..)) as I
 import DOM.HTML.Indexed.PreloadValue (PreloadValue(..)) as I
+import DOM.HTML.Indexed.ScopeValue(ScopeValue(..)) as I
 import DOM.HTML.Indexed.StepValue (StepValue(..)) as I
 import Data.Maybe (Maybe(..))
 import Data.MediaType (MediaType)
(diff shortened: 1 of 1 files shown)


src/Halogen/HTML/Properties.purs
```diff
@@ -80,6 +81,7 @@ import DOM.HTML.Indexed.MenuitemType (MenuitemType(..)) as I
 import DOM.HTML.Indexed.OnOff (OnOff(..)) as I
 import DOM.HTML.Indexed.OrderedListType (OrderedListType(..)) as I
 import DOM.HTML.Indexed.PreloadValue (PreloadValue(..)) as I
+import DOM.HTML.Indexed.ScopeValue(ScopeValue)
```
garyb, line 84:
```suggestion
import DOM.HTML.Indexed.ScopeValue (ScopeValue(..)) as I
```

I'd suggest echoing the other property-value-imports (like `PreloadValue`, `OrderedListType`, etc. above it). 

By doing this it re-exports the `ScopeValue` constructors (since `module I` is exported), which makes things a bit easier when using the HTML DSL, since you don't have to go hunting for the module to import to provide the value to `HP.scope` then - it'll be available as `HP.ScopeRow` (or whichever) too.

src/Halogen/HTML/Properties.purs
```diff
@@ -201,6 +203,9 @@ src = prop (PropName "src")
 style :: forall r i. String -> IProp (style :: String | r) i
 style = attr (AttrName "style")
 
+scope :: forall r i. ScopeValue -> IProp (scope :: ScopeValue | r) i
```
garyb, line 206:
```suggestion
scope :: forall r i. I.ScopeValue -> IProp (scope :: I.ScopeValue | r) i
```

If you accept the other suggestion, this will need doing too 🙂

garyb (changes requested):
Thanks for the contribution! I have a suggestion though:

pmiddend:
@garyb Thanks for the suggestion and explanation. I changed it accordingly.

garyb (approved):
Thanks!

PR #720, opened 2020-12-18
Add migration guide for Halogen 5 -> Halogen 6

thomashoneyman:
This PR adds a first draft of the migration guide from Halogen 5 to Halogen 6. All changes here are from merged pull requests and commits into the `next-6` branch, with the exception of the section on portals (the pull request for that hasn't been merged yet).

I'm getting this migration started now because we're getting so close to releasing PureScript 0.14, and we've planned the Halogen 6 release to occur at the same time. Given that the features and changes for Halogen 6 are just about all merged and ready to go it feels like a good time to get a draft of the migration guide in as well.

This is missing linked headers to each subsection, but otherwise is in shape for a preliminary review of the structure and contents.


garyb:
Great idea. Probably should have started a document like this when we started making changes, and kept it in sync!

thomashoneyman:
I’d like to do that for the future (same as we’re now doing in the contrib libraries) and can get that started once we ship!

PR #721, opened 2020-12-27
Support Purescript 0.14

thomashoneyman:
This PR updates the Halogen 6 branch for compatibility with PureScript 0.14. This can be merged into the `halogen-6` branch, but before we can actually release the library we'll need to wait on these dependencies:

- [ ] https://github.com/LiamGoodacre/purescript-filterable/pull/20
- [ ] https://github.com/paf31/purescript-event/pull/7
- [ ] https://github.com/garyb/purescript-quickcheck-laws/pull/53
- [ ] https://github.com/ethul/purescript-freeap/pull/19

This could use a quick review to ensure everything checks out, but the changes are minimal.


srghma:
https://github.com/paf31/purescript-event/pull/7 is archived

garyb:
Hmm, could perhaps just inline the event type into this library. Or maybe I'll fork event into contrib.

JordanMartinez:
Forking it to `contrib` seems like a better solution in my opinion. There are other things that could depend on and build on top of it then.

PR #723, opened 2020-12-27
Halogen 6

thomashoneyman:
This PR consolidates all the work that has gone into Halogen 6 this year so that we can merge to the default branch and begin the work for a proper release to go along with PureScript 0.14. Notable changes included in this pull request include:

- An updated changelog and guide to explain the changes in Halogen 6
- Simplifications to event handlers and event sources
- Internal updates for PureScript 0.14
- Plenty of small tweaks and polishing to make Halogen a pleasure to use day-to-day

This pull request deserves a glance over by core maintainers, but it's not necessary to fully review; it brings together many merged pull requests which were individually reviewed before being merged into the `next-6` branch in the first place. I have made a number of documentation updates over the past few days that do warrant an extra look, as do my changes to accommodate PureScript 0.14, but these are all conservative and I don't anticipate trouble.

From this point onward, we're waiting: 

- for the PureScript 0.14 release
- for our dependencies (notably `event` and `freeap`) to update for PureScript 0.14
- for #705 to land, which adds the final feature for this release (portals)


.github/workflows/ci.yml
```diff
@@ -4,7 +4,6 @@ on:
   push:
     branches: master
   pull_request:
-    branches: master
```
JordanMartinez, line 7: On most other repos, we have this included:
```yml
on:
  push:
    branches: [master]
  pull_request:
    branches: [master]
```

.github/workflows/ci.yml
```diff
@@ -16,7 +15,7 @@ jobs:
 
       - uses: purescript-contrib/setup-purescript@main
         with:
-          purescript: "0.13.8"
+          purescript: "0.14.0-rc4"
```
JordanMartinez, line 18: Needs to be `v0.14.0-rc5` now.

docs/changelog/v6.md
```diff
@@ -0,0 +1,214 @@
+# Changes in v6
+
+This is a crash-course on the changes from Halogen 5 to Halogen 6. Please ope nan issue or PR if you notice missing information or ways this guide could be improved!
```
JordanMartinez, line 3: Typo:
```diff
- Please ope nan issue
+ Please open an issue
```

docs/changelog/v6.md
```diff
@@ -0,0 +1,214 @@
...
+
+state <- io.query $ H.mkRequest SomeRequestQuery
+_ <- io.query $ H.mkTell SomeTellQuery
+```
+
+Added in [#621](https://github.com/purescript-halogen/purescript-halogen/pull/621).
+
+## Event Sources
+
+Event sources have been implemented with `Event` from the `event` library instead of the Halogen event sources implementation on top of `coroutines`. This simplifies the library internals and connects Halogen with a more widely-used FRP library.
```
JordanMartinez, line 154: We might want to add a link to that repo's issue tracker, so that people can become more aware of the current issues/bugs in this new dependency. It might save them some time.

docs/changelog/v6.md
```diff
@@ -0,0 +1,214 @@
...
+
+```purs
+import FRP.Event as Event
+import Halogen as H
+
+do
+  { event, push } <- H.liftEffect Event.create
+  void $ H.subscribe event
+  H.liftEffect $ push MyAction
+```
```
JordanMartinez, line 186: Should there be two examples here? One that never unsubscribes and a second that does unsubscribe at some point?

Also, should this be rendered as a `diff` to better highlight what did change between the two examples?

docs/guide/02-Introducing-Components.md
```diff
@@ -39,9 +38,9 @@ component =
 
   render state =
     HH.div_
-      [ HH.button [ HE.onClick \_ -> Just Decrement ] [ HH.text "-" ]
+      [ HH.button [ HE.onClick \_ -> Decrement ] [ HH.text "-" ]
       , HH.text (show state)
-      , HH.button [ HE.onClick \_ -> Just Increment ] [ HH.text "+" ]
+      , HH.button [ HE.onClick \_ -> Increment ] [ HH.text "+" ]
```
JordanMartinez, line 43: This means we can have `HE.onClick' Increment` where `onClick' a = onClick \_ -> a`. Very nice!

packages.dhall
```diff
@@ -1,24 +1,15 @@
 let upstream =
-      https://github.com/purescript/package-sets/releases/download/psc-0.13.6-20200507/packages.dhall sha256:9c1e8951e721b79de1de551f31ecb5a339e82bbd43300eb5ccfb1bf8cf7bbd62
+      https://raw.githubusercontent.com/purescript/package-sets/prepare-0.14/src/packages.dhall sha256:2c0a5af7ed5158218e0068f2328101fd9f0461e17ea37298e5af6875a96f34ac
```
JordanMartinez, line 2: Shouldn't this SHA be removed for the time being?

src/Halogen/HTML/Properties.purs
```diff
@@ -180,8 +182,11 @@ width = prop (PropName "width")
 href :: forall r i. String -> IProp (href :: String | r) i
 href = prop (PropName "href")
 
-id_ :: forall r i. String -> IProp (id :: String | r) i
-id_ = prop (PropName "id")
+id :: forall r i. String -> IProp (id :: String | r) i
+id = prop (PropName "id")
+
+id_ :: forall r i. TypeError.Warn (TypeError.Text "`id_` is deprecated. Prefer `id` instead.") => String -> IProp (id :: String | r) i
```
JordanMartinez, line 188:
```diff
- deprecated. Prefer `id` instead.
+ deprecated. Use `id` instead.
```

docs/guide/02-Introducing-Components.md
garyb: I was actually thinking, if we were to have an `HE.onClick'`, it would be to reintroduce the version that includes the `Maybe`, so perhaps not :wink:.

We could maybe have `HE.onClick_` for what you're talking about, but having `'` and `_` of each function might not be the friendliest API to discover. (I did it in affjax at one point, before I knew better, but a lot of those variants got removed eventually due to other changes anyway).

JordanMartinez: Ah... Well, maybe as a project-specific Prelude then?

garyb: I was thinking of another option when writing that comment earlier too, perhaps the event definitions that can "fail" or be "filtered" (those with the `Maybe`) could live in a separate module - they're required less often, so it would be no great hardship to have a second import for them.

docs/changelog/v6.md
```diff
@@ -0,0 +1,214 @@
...
+
+This is a breaking change, but one which is easily fixed: remove this parameter from your components and any related functions and types.
+
+Added in [#616](https://github.com/purescript-halogen/purescript-halogen/pull/616).
+
+## Event Handler Types
+
+We've also made event handlers a little nicer to work with. The `Maybe action` return value has been removed from event handlers, which now return `action` directly.
+
+In Halogen 5, when you wanted to respond to a click event, or a message from a child component, the output was of type `Maybe action`. This allowed you to selectively emit outputs (you could emit `Nothing` for some events). In practice, though, almost no one ever does this.
```
MonoidMusician, line 72: Would you like to add a note that users can add a `DoNothing` constructor to their action type or replace the action type with `Maybe action` instead, if they need to keep the old behavior?

.github/workflows/ci.yml
thomashoneyman: `[master]` and `master` are equivalent -- though now that I think of it, we really ought to support CI on pull requests against any branch, not just the master. branch.

docs/changelog/v6.md
thomashoneyman: The entire body would be highlighted in the diff, and one of the imports, leaving only the `Halogen` import and the `do` lines untouched. For that reason I think it's better to have separate examples.

docs/changelog/v6.md
thomashoneyman: Added!

thomashoneyman:
I plan to merge this PR tonight unless anyone has an objection @natefaubion @garyb @MonoidMusician @JordanMartinez. From then on the default Halogen branch will be for Halogen 6, which we'll need to continue iterating on until it's ready for release (largely, updating dependencies and the target package set).

PR #729, opened 2021-02-03
update to v0.14

srghma:
error after https://github.com/purescript/purescript-ordered-collections/pull/38


srghma:
all errors are fixed for me @thomashoneyman

src/Halogen/Data/Slot.purs
```diff
@@ -112,13 +117,14 @@ slots
   => Proxy sym
   -> SlotStorage slots slot
   -> Map s (slot query output)
-slots sym (SlotStorage m) = Map.foldSubmap Nothing Nothing go m
```
thomashoneyman, line 115: I don't think this needs to be removed -- `foldSubmap` still exists in ordered-collections

src/Halogen/Data/Slot.purs
```diff
@@ -112,13 +117,14 @@ slots
...
+slots sym (SlotStorage m) = mapMaybeWithKeyToNewKey go m
   where
   key = reflectSymbol sym
 
+  go :: Tuple String (OrdBox Any) -> Any -> Maybe (Tuple s (slot query output))
   go (Tuple key' ob) val
-    | key == key' = Map.singleton (unOrdBox (coerceBox ob)) (coerceVal val)
-    | otherwise = mempty
```
thomashoneyman, line 121: If this is what's giving you the error, you can just do `Map.empty` here instead of `mempty` but otherwise leave things as the were.

src/Halogen/Data/Slot.purs
srghma: yes, I have tried to impl it with SemigroupMap but it didnt work

but it required Semigroup on v

```
instance semigroupSemigroupMap :: (Ord k, Semigroup v) => Semigroup (SemigroupMap k v) where
  append (SemigroupMap l) (SemigroupMap r) = SemigroupMap (unionWith append l r)
```

MonoidMusician: you can use the Alternate newtype to get back the old behavior: https://pursuit.purescript.org/packages/purescript-control/4.2.0/docs/Data.Monoid.Alternate#t:Alternate

PR #730, opened 2021-02-13
Update `onClick` to use the new type signature

decorator-factory:
Event handlers accept a function that returns a `Maybe`, according to the reference: https://pursuit.purescript.org/packages/purescript-halogen/5.0.0/docs/Halogen.HTML.Events#v:onClick

However, in the Halogen guide, the event handlers don't return a `Maybe`, so the example don't even compile. 
Maybe I'm missing something?

### DIFF
--- a/docs/guide/02-Introducing-Components.md
+++ b/docs/guide/02-Introducing-Components.md
@@ -21,6 +21,7 @@ module Main where
 
 import Prelude
 
+import Data.Maybe (Maybe(..))
 import Halogen as H
 import Halogen.HTML as HH
 import Halogen.HTML.Events as HE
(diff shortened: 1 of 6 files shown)


garyb:
You're looking at `master` which is ahead of the published version on Pursuit 🙂.

Take a look at the guide for the tagged release instead: https://github.com/purescript-halogen/purescript-halogen/tree/v5.1.1/docs/guide

decorator-factory:
Ah, sorry then. I was reading the github.io version https://purescript-halogen.github.io/, which looks prettier than just markdown. Is it possible to make the github.io page refer to the stable version?

PR #731, opened 2021-02-17
Docs: wrap onClick actions in a Maybe

mark-gerarts:
The [first example in the guide](https://purescript-halogen.github.io/purescript-halogen/guide/index.html) does not compile because the `onClick` actions have to be wrapped in a maybe.  It is quite possible that this is not the correct way to fix it, since I'm only just starting out with Halogen (hence why I encountered it :slightly_smiling_face:).

### DIFF
--- a/docs/guide/README.md
+++ b/docs/guide/README.md
@@ -9,6 +9,7 @@ module Main where
 
 import Prelude
 
+import Data.Maybe (Maybe(..))
 import Effect (Effect)
 import Halogen as H
 import Halogen.Aff as HA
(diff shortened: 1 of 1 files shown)


decorator-factory:
@mark-gerarts I encountered it as well! https://github.com/purescript-halogen/purescript-halogen/pull/730

I think it's very confusing that https://purescript-halogen.github.io/ refers to the latest version and not the stable version, which makes it practically unusable for learning.

garyb:
Yeah, I need to figure out how to make the action only publish that when releases are made, but I've not had the time to look into it yet. Or ideally, even have it publish multiple versions like readthedocs does. I didn't set this up so I'll need to do some reading.

decorator-factory:
I don't understand why GitHub doesn't allow you to select a folder other than `/` or `/docs`, or to choose a tag :thinking:

mark-gerarts:
I hadn't seen your PR (I searched through open issues). If you're aware of the issue, all is well!

thomashoneyman:
Halogen 6 has now been released, so everything's up to date. I jumped the gun a little early on updating the master branch to Halogen 6; thought it would have been a little faster to ship. Thanks for the PR!

PR #734, opened 2021-03-07
Migrate events to halogen-subscriptions and complete 0.14 update

thomashoneyman:
This PR takes two actions to unblock the Halogen 6 release:

1. It completes the migration to PureScript 0.14
2. It moves Halogen from the `event` library (which is archived) to the new `halogen-subscriptions` library, which includes updating the documentation in this repository.

The new `halogen-subscriptions` library is inspired by `event` but has been moderately changed to work in the context of Halogen applications. The library is not tied to Halogen specifically and can be used in other contexts as well. This PR has been marked a draft until we're ready to publish the v1 release of that library:

https://github.com/purescript-halogen/purescript-halogen-subscriptions

Once `halogen-subscriptions` has been released we can update our `bower.json` and `packages.dhall` files and publish this library as well.


src/Halogen/Query/Event.purs
```diff
@@ -4,22 +4,22 @@ import Prelude
...
+import Halogen.Subscription as HS
+import Web.Event.Event as Event
+import Web.Event.EventTarget as EventTarget
 
--- | Constructs an `Event` for a DOM event. Accepts a function that maps event
+-- | Constructs an `Emitter` for a DOM event. Accepts a function that maps event
 -- | values to a `Maybe`-wrapped action, allowing it to filter events if
 -- | necessary.
-eventListenerEventSource
+eventListener
```
thomashoneyman, line 14: I renamed this function because `eventSource` isn't a concept any longer. Another alternative would be `eventListenerEmitter`, but I found that a bit weird to read. 

Ultimately Halogen already shadows functions from other libraries (`Halogen.subscribe` for `Halogen.Subscription.subscribe`, etc.), so this seemed like the best bet.

garyb (approved):

PR #736, opened 2021-03-11
02-Introducing-Components H.Component type params

shaunplee:
As of Halogen 6, `H.Component` takes four type parameters (`query input output m`), but the text says it takes "five".


garyb:
Thanks!

PR #739, opened 2021-03-15
Add tellAll and requestAll functions

### LINKED ISSUE #738: requestAll and tellAll functions
Section 5 of the Halogen Guide references functions `requestAll` and `tellAll`, but these don't seem to exist. In the example code `H.requestAll` is used, with H the qualified Halogen module import, but this function is not located in that module (or any other module I've looked in).

thomashoneyman:
Fixes #738 by adding `requestAll` and `tellAll` functions -- these were omitted as an oversight in 6.0.0.

### DIFF
--- a/src/Halogen/Query.purs
+++ b/src/Halogen/Query.purs
@@ -4,9 +4,11 @@ module Halogen.Query
   ( Tell
   , mkTell
   , tell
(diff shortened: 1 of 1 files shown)


garyb (approved):
Ooops

PR #743, opened 2021-04-07
Fix memoized

### LINKED ISSUE #742: `memoized` doesn't seem to work
`lazy`* is fine, but neither me or Nate could get `memoized` to work right.

thomashoneyman:
Fixes #742 by using the same non-eta-expanded implementation as Spork. I tested this locally by updating the `components-inputs` example to use the previous implementation and the Spork implementation and could verify that the `memoized` function only works when done the Spork way.


thomashoneyman:
If desired, I could also add a new example that demonstrates the `lazy` and `memoized` functions in action. It would most likely require unsafe logging in the render function to prove that the render fired or didn't fire.

thomashoneyman:
I adjusted the existing components-inputs example to make sure this works :) I’d like to add a new example, though, just so we can more easily test in the future.

thomashoneyman:
#747 demonstrates that this update does indeed fix the `memoized` function.

garyb (approved):

PR #750, opened 2021-04-19
Fix warnings found by v0.14.1 PS release

JordanMartinez:


garyb (approved):
I am astounded that this is all there is 😆

PR #760, opened 2021-07-29
Fixing error in the guide: requestAll function is in Halogen.Query.

KaneRoot:
As explained in [Issue 759](https://github.com/purescript-halogen/purescript-halogen/issues/759), the example in the guide was wrong: `requestAll` was failing.

I found out that `requestAll` actually is a function in `Halogen.Query`and not directly in `Halogen`.

I tried this version, this now works in `Try Purescript`.

### DIFF
--- a/docs/guide/05-Parent-Child-Components.md
+++ b/docs/guide/05-Parent-Child-Components.md
@@ -667,6 +667,7 @@ import Effect (Effect)
 import Effect.Class (class MonadEffect)
 import Effect.Class.Console (logShow)
 import Halogen as H
+import Halogen.Query as HQ
 import Halogen.Aff as HA
 import Halogen.HTML as HH
 import Halogen.HTML.Events as HE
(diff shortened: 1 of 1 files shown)


kritzcreek:
I think `requestAll` should probably be exported from `Halogen` instead: https://github.com/purescript-halogen/purescript-halogen/blob/40cd8440470a3137e394c85f5b8ff6ec3bc7a9ca/src/Halogen.purs#L23

Maybe it was lost in some restructuring?

thomashoneyman:
Good find, @KaneRoot! I think @kritzcreek is right, and the error is that `requestAll` isn't being re-exported from the Halogen module, and not that the guide is incorrect. In the Halogen 5 -> Halogen 6 transition this function did get lost.

thomashoneyman:
If you make that change I'm happy to merge this!

KaneRoot:
I guess I'll make another PR.

PR #765, opened 2021-08-09
Introduce purs-tidy formatter

thomashoneyman:
This PR introduces [`purs-tidy`](https://github.com/natefaubion/purescript-tidy) to Halogen so we can easily keep consistent formatting as new pull requests come in. It also adds a check in CI for this.

In the `.tidyrc.json` file I've set the unicode preference to "source" to avoid making changes. My personal preference is not to use unicode in multi-developer / community projects, but I'm aware that unicode has historically been used in Halogen. @garyb, do you have thoughts on this?


src/Halogen/Aff/Driver/Eval.purs
```diff
@@ -72,14 +69,14 @@ evalM
    . Renderer r
   -> Ref (DriverState r s f act ps i o)
   -> HalogenM s act ps o Aff
-  ~> Aff
+       ~> Aff
```
thomashoneyman, line 72: These aren't ideal; we could rewrite them without the natural transformation operator, or live with the indentation. Thoughts @natefaubion @garyb?

src/Halogen/HTML/Events.purs
```diff
@@ -121,11 +125,14 @@ onFileUpload
...
 onFileUpload f = handler ET.change $
-  ( Event.target >=>
-    HTMLInputElement.fromEventTarget >=>
-    HTMLInputElement.files >>> unsafePerformEffect ) >>>
-  maybe none items >>>
-  f
+  ( Event.target
+      >=> HTMLInputElement.fromEventTarget
+      >=>
+        HTMLInputElement.files >>> unsafePerformEffect
```
thomashoneyman, line 131: We can pull this up a line using parentheses, ie. `(HTMLInputElement.files >>> unsafePerformEffect)`

src/Halogen/HTML/Events.purs
```diff
@@ -121,11 +125,14 @@ onFileUpload
...
+  )
+    >>> maybe none items
+    >>>
+      f
```
thomashoneyman, line 135: Similarly, we should probably pull this up a line.

examples/lifecycle/src/Child.purs
```diff
@@ -76,11 +78,13 @@ cell initialState =
...
-        , initialize = Just Initialize
-        , finalize = Just Finalize
-        })
+    , eval: H.mkEval
+        ( H.defaultEval
+            { handleAction = handleAction
+            , initialize = Just Initialize
+            , finalize = Just Finalize
+            }
+        )
```
thomashoneyman, line 87: If we don't like this, we could rewrite to `H.mkEval $ H.defaultEval`  instead.

src/Halogen/Aff/Driver.purs
```diff
@@ -93,8 +93,8 @@ import Halogen.Subscription as HS
 -- | The `dispose` function is called when the top level component is disposed of
 -- | via `HalogenIO`.
 type RenderSpec r =
-  { render
-      :: forall s act ps o
+  { render ::
+      forall s act ps o
```
thomashoneyman, line 97: This is a bit odd (the `::` being pulled up a line) -- thoughts on this, @natefaubion?

src/Halogen/Aff/Driver/Eval.purs
```diff
@@ -38,10 +38,7 @@ import Halogen.Subscription as HS
 import Unsafe.Reference (unsafeRefEq)
 
 type Renderer r
-  = forall s f act ps i o
-   . Ref LifecycleHandlers
-  -> Ref (DriverState r s f act ps i o)
-  -> Effect Unit
+  = forall s f act ps i o. Ref LifecycleHandlers -> Ref (DriverState r s f act ps i o) -> Effect Unit
```
natefaubion, line 41: This looks like a bug?

src/Halogen/Aff/Driver.purs
```diff
@@ -117,25 +117,28 @@ runUI renderSpec component i = do
...
 
   where
 
   evalDriver
     :: forall s f' act ps i' o'
      . Ref Boolean
     -> Ref (DriverState r s f' act ps i' o')
-    -> forall a. (f' a -> Aff (Maybe a))
+    -> forall a
+     . (f' a -> Aff (Maybe a))
```
thomashoneyman, line 137: This is also a bit strange. Should we parenthesize this whole line?

```purs
-> (forall a. f' a -> Aff (Maybe a))
```

src/Halogen/Aff/Driver/Eval.purs
thomashoneyman, line 41: It would be nice to still have this break over lines, but it's not awful to be all on one line either.

thomashoneyman: I was about to comment on that. This doesn't look like it's the correct behavior.

thomashoneyman:
Hm -- `purs-tidy check` is failing in CI with:

```
Error: Cannot find module 'worker_threads'
```

as seen in this workflow run:
https://github.com/purescript-halogen/purescript-halogen/pull/765/checks?check_run_id=3283492228

natefaubion:
If it can't find worker_threads, the node version is too old.

src/Halogen/Aff/Driver.purs
natefaubion: It's debatable, but it's currently consistent with the value level syntax for fields.

src/Halogen/Aff/Driver.purs
natefaubion: If you want it to appear on the same line, you would need to parenthesize it. That's because `forall` is a quantifier like `->`. For example `forall a. forall b. forall c ...` should format as
```
foo
  :: forall a
   . forall b
   . forall c
   . ...
```

thomashoneyman:
In d4fabcd I switched to `unicode: never` because there are actually only a handful of unicode symbols in the entire codebase.

src/Halogen/Aff/Driver/Eval.purs
garyb: I would think this is perhaps not entirely intentional? I suppose outside of arrow-like operators it is odd to split them across lines though. Any thoughts Nate?

I would definitely rewrite rather than use indentation here if we have no other choice.

natefaubion: This is expected. `~>` is an operator and doesn't have the same precedence as quantifiers. This goes double if you ever try to use `~>` in any position besides the final return value.

src/Halogen/HTML/Events.purs
```diff
@@ -120,12 +124,14 @@ onFileUpload
...
-  maybe none items >>>
-  f
+onFileUpload f = do
+  let
+    filesFromTarget =
+      Event.target
+        >=> HTMLInputElement.fromEventTarget
+        >=> (HTMLInputElement.files >>> unsafePerformEffect)
+
+  handler ET.change $ filesFromTarget >>> maybe none items >>> f
```
garyb, line 134: Nitpick: I'd do this with a `where` instead.

garyb:
> 
> 
> In [d4fabcd](https://github.com/purescript-halogen/purescript-halogen/commit/d4fabcdd763d76ad9f795373e8598e8aa64cf046) I switched to `unicode: never` because there are actually only a handful of unicode symbols in the entire codebase.

Yeah, they only snuck in because I have muscle memory for inserting the symbols! Totally in favour of tidying them out. :+1:

src/Halogen/HTML/Events.purs
```diff
@@ -93,12 +93,16 @@ import Effect.Unsafe (unsafePerformEffect)
...
+    et
+    \ev -> Just (Action (f ev))
 
 handler' :: forall r i. EventType -> (Event -> Maybe i) -> IProp r i
 handler' et f =
   (unsafeCoerce :: (EventType -> (Event -> Maybe i) -> Prop i) -> EventType -> (Event -> Maybe (Input i)) -> IProp r i)
-    Core.handler et \ev -> Action <$> f ev
+    Core.handler
+    et
+    \ev -> Action <$> f ev
```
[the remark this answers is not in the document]
natefaubion: You would need a different type for the unsafeCoerce. My suggestions would be to do

```
  ...
  where
  handler' :: ...
  handler' = unsafeCoerce Core.handler
```

src/Halogen/Aff/Driver/Eval.purs
garyb: It might be "expected" but it's nasty! 😄 By "not entirely intentional" here I meant more along the lines of, it might be following the rules, but the consequence is something unwanted.

natefaubion: If you can think of a rule that doesn't hardcode an unqualified `~>`, then I'm all ears ;P

garyb: I guess a question I have is why does it indent based on precedence? Or is the precedence we're talking about here syntax vs identifiers?

thomashoneyman:
By the way -- @natefaubion, @garyb feel free to push any changes you want to this branch if you know you'd like something formatted in a specific way!

src/Halogen/HTML/Events.purs
garyb: 👍

src/Halogen/Aff/Driver/Eval.purs
natefaubion: In this case, yes, it is syntax vs identifiers. Type expressions will be indented past the arrow. We do indent based on operator precedence though elsewhere!

thomashoneyman:
I've run this again with purs-tidy 0.4.6.

src/Halogen/VDom/Driver.purs
```diff
@@ -91,8 +91,8 @@ mkSpec handler renderChildRef document =
 
     patch
       :: EFn.EffectFn2 (WidgetState slots action)
-            (ComponentSlot slots Aff action)
-            (V.Step (ComponentSlot slots Aff action) DOM.Node)
+        (ComponentSlot slots Aff action)
+        (V.Step (ComponentSlot slots Aff action) DOM.Node)
```
thomashoneyman, line 95: I suppose this is because of the lack of a `forall` or `->`.

natefaubion: This is fixed in the 0.5 branch of purs-tidy.

thomashoneyman: I'm still seeing it happen with 0.5.1, but maybe I'm accidentally hooking in to an older version.

thomashoneyman: Working better now that this is updated to 0.5.1, thanks!

src/Halogen/HTML/Events.purs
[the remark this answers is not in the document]
thomashoneyman: This no longer happens with purs-tidy 0.5.1

thomashoneyman:
I'm pretty happy with the formatting now, and I plan to merge sometime next week unless y'all have objections.

PR #771, opened 2021-08-18
Make example work again in Try Purescript

nilp0inter:


thomashoneyman:
This was actually fixed in Halogen v6.1.3; we accidentally removed the re-export in an earlier version but the function is still supposed to be available from the main Halogen module.

The real fix here is to update Try PureScript to use the latest version of Halogen, which I can take care of tomorrow!

nilp0inter:
Great, thank you!

thomashoneyman:
Try loading this on Try PureScript now!
https://github.com/purescript-halogen/purescript-halogen/blob/5723e5ab654def0fe1f754ebdef61384e8038c8a/docs/guide/05-Parent-Child-Components.md#full-example

nilp0inter:
It works! Thx!

PR #772, opened 2021-08-30
Add fp made easier book

janat08:
By pages like 1/4 of the book works with halogen. While it's paid buying it isn't a poor proposition.


thomashoneyman:
This feels like it may be better-suited to a "learn PureScript" collection of resources, which I'm not sure Halogen should have listed out in its main README. Does the book specifically focus on Halogen?

janat08:
No it focuses backend and the language itself too.

PR #777, opened 2021-11-03
Fix outdated example in Guide section 6

newlandsvalley:
Use Halogen.Subscription instead of FRP.Event.  Also, remove reference
to the component as B which is actually incorporated into Main.

### DIFF
--- a/docs/guide/06-Running-Application.md
+++ b/docs/guide/06-Running-Application.md
@@ -79,29 +79,28 @@ import Prelude
 import Data.Maybe (Maybe(..))
 import Effect (Effect)
 import Effect.Console (log)
-import Example.Driver.IO.Button as B
-import FRP.Event as Event
 import Halogen (liftEffect)
 import Halogen as H
 import Halogen.HTML as HH
 import Halogen.Aff as HA
 import Halogen.HTML.Events as HE
 import Halogen.HTML.Properties as HP
+import Halogen.Subscription as SUB
 import Halogen.VDom.Driver (runUI)
 
 main :: Effect Unit
 main = HA.runHalogenAff do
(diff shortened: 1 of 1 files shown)


docs/guide/06-Running-Application.md
```diff
@@ -79,29 +79,28 @@ import Prelude
...
 import Effect.Console (log)
-import Example.Driver.IO.Button as B
-import FRP.Event as Event
 import Halogen (liftEffect)
 import Halogen as H
 import Halogen.HTML as HH
 import Halogen.Aff as HA
 import Halogen.HTML.Events as HE
 import Halogen.HTML.Properties as HP
+import Halogen.Subscription as SUB
```
thomashoneyman, line 88: Do you mind switching this to:

```suggestion
import Halogen.Subscription as HS
```

That keeps it the same as everywhere else we import this. Thanks!

docs/guide/06-Running-Application.md
```diff
@@ -79,29 +79,28 @@ import Prelude
...
 import Halogen.VDom.Driver (runUI)
 
 main :: Effect Unit
 main = HA.runHalogenAff do
   body <- HA.awaitBody
-  io <- runUI B.component unit body
+  io <- runUI component unit body
 
-  _ <- liftEffect $ Event.subscribe io.messages \(Toggled newState) -> do
+  _ <- liftEffect $ SUB.subscribe io.messages \(Toggled newState) -> do
```
thomashoneyman, line 96:
```suggestion
  _ <- liftEffect $ HS.subscribe io.messages \(Toggled newState) -> do
```

thomashoneyman (approved):

PR #787, opened 2022-02-08
Add support for raw HTML

flip111:
tests pass
most examples are working. The ones that don't i don't think it's because of this PR but didn't look into it
comes with example of usage

Addresses https://github.com/purescript-halogen/purescript-halogen/issues/324

* Not sure for the best way to implement patch
* [ComponentSlot](https://github.com/flip111/purescript-halogen/blob/raw-html-widget/src/Halogen/Component.purs#L204) should be renamed to `HalogenWidget` imo because the name doesn't cover the extra constructor. This would be a breaking change, while currently what is added i don't think it breaks anything so it could be updated in minor version of halogen library.
* add documentation, only first element in raw html string will be used
* rename function `rawHTML` just to `html` or `raw`?

Review appreciated.


src/Halogen/VDom/Driver.js
```diff
@@ -0,0 +1,12 @@
+"use strict";
+
+exports.createElementFromHTML = function(html) {
+  // used: https://stackoverflow.com/a/35385518/1833322
+  // alternatively: https://stackoverflow.com/a/494348/1833322
+  var template = document.createElement('template');
+  html = html.trim(); // Never return a text node of whitespace as the result
+  template.innerHTML = html;
+  var newElement = template.content.firstChild;
```
MonoidMusician, line 9: Does this mean it only supports a single HTML element? That should at least be documented somewhere. Maybe with support for fragments it could be improved to capture all children.

flip111: I hadn't considered that .. but good point. I guess it does mean that.

It doesn't necessarily need to be like this but inserting multiple elements would need to happen at the parent of the current element. Which i think would make the code unnecessary more complicated and might not even be possible (what to do if the raw-html was the top halogen element).

Also i think it's unwanted to enter multiple nodes in one location. Because you can just put a node around it. i.e. `<b>node1</b><b>node2</b>` into `<div><b>node1</b><b>node2</b></div>`.

What could be done is count the amount of children and if it's more than 1 to show a [console warning](https://developer.mozilla.org/en-US/docs/Web/API/Console/warn) on runtime.

Yes and documentation.

flip111:
[How can i compare the html string](https://github.com/flip111/purescript-halogen/blob/raw-html-widget/src/Halogen/VDom/Driver.purs#L109)

How can i make this more efficient ? Like comparing the new html string with the old one and not render in case they are the same ?

flip111:
Hey, I've come back to this after a long time. The branch is now merged up with upstream master and the implementation is reworked.

The original version had a bug where old DOM nodes were never removed during re-renders. Every time the component updated, a new element was inserted but the previous one stayed in the DOM. This caused duplicate nodes to pile up and all kinds of weird browser behavior.

The root cause was using `Nothing` as the widget state for RawHTML. The patch and halt functions had no reference to the old node so they couldn't clean it up. The fix replaces the `Maybe`-based `WidgetState` with a proper sum type that tracks the DOM node for RawHTML widgets. Now `patch` can do an in-place `replaceWith()` and `done` can remove the node on cleanup.

There were also some JS bugs. `outerHtml` should have been `outerHTML`. String concat used `.` instead of `+`. The `outerHTML` setter approach itself was flawed because it detaches the original node, leaving the VDom with a stale reference. All replaced with a simple `replaceWith()` call.

The example is expanded to cover dynamic updates, conditional rendering, mixed content, and growing lists. I verified in the browser that all scenarios work correctly including rapid toggling and rapid state updates.

Regarding the open questions from before: I'm keeping `ComponentSlot` as-is (renaming is a breaking change for little gain), keeping `rawHTML` as the function name, and keeping the single-element restriction (wrapping in a div is easy enough).

Happy to address any review feedback.

PR #790, opened 2022-04-28
Update for PureScript 0.15

### LINKED ISSUE #788: Halogen Ace example throws a ReferenceError
I downloaded prescript-halogen-6.1.3 and I tried to run the Ace example. When I open the index.html file I get a ReferenceError in the browser:
```ReferenceError: Can't find variable: require```

Comment by @natefaubion : It looks like the ace FFI previously relied on globals only, but was changed to use require, and I don’t think the halogen examples handle npm dependency bundling.

I got it working by using rollup for the npm dependency bundling. I had to manually install ace-builds.

### PR DESCRIPTION
**Pull request by thomashoneyman**
Updates Halogen for PureScript 0.15. No API changes, only dependency upgrades. Also fixes #788.

### CODE REVIEW

**Review by garyb — APPROVED** [review]

**Comment by thomashoneyman** [comment]
@garyb is there anything else you think we should quickly get in before doing a major release?

**Comment by garyb** [comment]
@thomashoneyman #785 would be worth doing! I'll open a PR later on for it, will take a look through the properties and see if there are any others that are worth changing similarly too.

**Comment by garyb** [comment]
I might have one more addition to go in too, I'm trying it out at the moment, it might not work/be safe enough to go in :wink: (`join` operation for forked `HalogenM`s) but I'll have an idea later today.

File: bower.json [hunk 1]
```diff
@@ -1,58 +1,51 @@
...
+    "type": "git",
+    "url": "https://github.com/purescript-halogen/purescript-halogen"
+  },
+  "ignore": ["**/.*", "node_modules", "bower_components", "output"],
+  "dependencies": {
+    "purescript-aff": "^v7.0.0",
+    "purescript-bifunctors": "^v6.0.0",
+    "purescript-console": "^v6.0.0",
+    "purescript-control": "^v6.0.0",
+    "purescript-dom-indexed": "^9.0.0",
```
thomashoneyman, line 14:
```suggestion
    "purescript-dom-indexed": "^10.0.0",
```

Issue #749, opened 2021-04-12
Attribute "autocomplete" is a String -- not a Boolean 


toastal:
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete

https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fe-autocomplete

It can take a large number of values like `"given-name`", et. al.

Related: https://github.com/purescript-web/purescript-web-html/issues/56

PR #791, opened 2022-04-29
Extras for the PS 0.15 release

### LINKED ISSUE #785: Attribute "value" for tag "meter" is broken
https://github.com/purescript-halogen/purescript-halogen/blob/v6.1.3/src/Halogen/HTML/Properties.purs#L238

https://github.com/purescript-halogen/purescript-dom-indexed/blob/v8.0.0/src/DOM/HTML/Indexed.purs#L446

garyb:
Resolves #785

I think #749 is worth looking at too perhaps (mostly needs changing in dom-indexed). As much as I dislike the idea of stringly typing, the context-based behaviour of the values that can be provided here suggests that maybe a more precise modelling isn't much use anyway. I don't know... it might still be nice to enumerate the many various options in a type anyway, since at least then it prevents typos / allows discovery by editor autocomplete, etc. What do you think @thomashoneyman?

### DIFF
--- a/src/Halogen/HTML/Properties.purs
+++ b/src/Halogen/HTML/Properties.purs
@@ -242,7 +242,7 @@ noValidate = prop (PropName "noValidate")
 type_ :: forall r i value. IsProp value => value -> IProp (type :: value | r) i
 type_ = prop (PropName "type")
 
-value :: forall r i. String -> IProp (value :: String | r) i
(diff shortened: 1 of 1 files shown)


thomashoneyman:
I think it's OK to move it to be a `String`, personally.

thomashoneyman (approved):

thomashoneyman:
I've added this as a sum type: https://github.com/purescript-halogen/purescript-dom-indexed/pull/29

PR #794, opened 2022-05-07
add support for pointer events

gbagan:
Add support for pointer events
https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events

A PR to dom-indexed will follow.

### DIFF
--- a/packages.dhall
+++ b/packages.dhall
@@ -5,6 +5,6 @@ let upstream =
 in  upstream
   with dom-indexed =
     { repo = "https://github.com/purescript-halogen/purescript-dom-indexed.git"
-    , version = "v10.0.0"
+    , version = "v10.1.0"
     , dependencies = upstream.dom-indexed.dependencies
     }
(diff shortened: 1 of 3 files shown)


thomashoneyman:
Thanks, @gbagan. Ping me when the dom-indexed PR is ready!

thomashoneyman:
How would you like to handle this change, @garyb, given that this is a major version upgrade to `dom-indexed`?

garyb:
A major bump to a dependency is a major bump to a dependant... so major bump I guess! :smile:

I don't have anything else breaking just now, I don't think it's worth holding this up until we do.

thomashoneyman:
Yea I meant in the way of handling Halogen 8, as we just released Halogen 7 last week 😅

garyb:
I think we just make a release without any fanfare, I'm not sure if people are paying that close attention to what the big number is... I wasn't sure what number we were up to now at least :smile:

Issue #770, opened 2021-08-15
New example for websocket


flip111:
People say with the changes in halogen 6 with Halogen.Query.Event.eventListener an Emitter for the websocket can be created directly and there no longer a need for the use of coroutines in the websocket example https://github.com/purescript-halogen/purescript-halogen/blob/master/examples/driver-websockets/src/Main.purs#L69 would be create to have a new version of this example that doesnt use coroutine

rskew:
Hi @flip111 , I've had a go at creating an example that uses `Halogen.Query.Event.eventListener` to subscribe to a websocket without requiring coroutines. I find it much simpler, hopefully you do to.

flip111:
That's great, thanks a lot @rskew  :smile:

PR #804, opened 2022-08-14
New example for websockets (#770)

rskew:
As raised in issue https://github.com/purescript-halogen/purescript-halogen/issues/770, new Halogen event machinery can allow use of websockets without the need for coroutines.

This updated example has the component managing its own websocket connection, which allows it to take alternative action if the connection state is not "open".

I'd be happy to action feedback to make this example more useful, or fit in better with the repo.


garyb:
The focus of the original example wasn't really the websocket, it was illustrating two way communication with the top level component through `DriverIO` - I think it's maybe a problem with the way the examples are "marketed" right now, but they're all supposed to usages of Halogen features, not how it interacts with other libraries. The routing example kinda has the same problem, people only see "routing" in the name and not "driver" part. :smile:

There's definitely value in having examples of things that would commonly be used with Halogen, but perhaps we should set up a kind of "cookbook" repo or something like that instead. Thoughts @natefaubion / @thomashoneyman?

natefaubion:
There is https://github.com/JordanMartinez/purescript-cookbook which already has a bunch of halogen recipes.

garyb:
Well, that's embarrassing, sorry @JordanMartinez!

Should probably add that to the readme. And yeah, I guess it is a bit hook-heavy, but I figure non-hook contributions would probably be welcome too.

JordanMartinez:
Feel free to contribute to the cookbook repo.
The issue I experienced in the cookbook repo is that showing both classic and hooks style examples tends to clutter the repo.

rskew:
Thanks all,

@garyb I've reinstated the use of `DriverIO`, it's ended up looking pretty similar to the old version.
I find it more intuitive to work with event listeners rather than coroutines, but I don't mind if the old version is preferred.

And thanks @JordanMartinez , I wasn't even aware of Halogen's hooks functionality! I'll check out the examples in purescript-cookbook and maybe contribute their next time.

garyb:
> I find it more intuitive to work with event listeners rather than coroutines

Completely agree! Coroutines have their place, and used to be required for dealing with messages coming out of `DriverIO`, but since we reworked subscriptions in Halogen they shouldn't be necessary in most situations.

I'll take a closer look at this later this evening, but at an initial glance it looks good to me. Thanks!

PR #817, opened 2023-08-10
Support general mapping of non-isomorphic state, action, output and monad.

bakhtiyarneyman:
Not tested for performance regressions.


garyb:
Thanks!

PR #819, opened 2023-09-07
Add onBeforeInput property

MonaMayrhofer:
This adds a property "onBeforeInput" to react to [beforeinput](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/beforeinput_event) events.

~~Currently blocked by https://github.com/purescript-halogen/purescript-dom-indexed/pull/32~~

### DIFF
--- a/src/Halogen/HTML/Events.purs
+++ b/src/Halogen/HTML/Events.purs
@@ -98,6 +98,7 @@ import Web.UIEvent.KeyboardEvent (KeyboardEvent)
 import Web.UIEvent.KeyboardEvent.EventTypes as KET
 import Web.UIEvent.MouseEvent (MouseEvent)
 import Web.UIEvent.MouseEvent.EventTypes as MET
+import Web.UIEvent.InputEvent.EventTypes as IET
 import Web.PointerEvent (PointerEvent)
 import Web.PointerEvent.EventTypes as PET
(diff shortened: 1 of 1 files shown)


MonaMayrhofer:
Whoops, It seems like i got a bit mixed up, while trying to upstream my changes, first i forgot to change `ET.input` to `ET.beforeinput` and second i didn't realize we are using on `Web.HTML.Event.EventTypes` instead of `Web.UIEvent.InputEvent.EventTypes`, which doesn't yet have beforeinput

MonaMayrhofer:
I am not sure whether this is the correct way to do this, and I am especially confused why the `input` event is imported from `Web.HTML.Event.EventTypes`? Is there a reason why most events are duplicated in `web-html` and `web-uievents` and some are not?

Thanks a lot, and sorry for the noise around such a small change. :sweat_smile:

garyb:
I don't think it matter too much where the `EventType` comes from since it's just a wrapper over a string. It probably should only be defined in `Web.UIEvent.InputEvent.EventTypes` and `Web.HTML.Event.EventTypes` might not need to exist at all.

MonaMayrhofer:
Ahh, okay!

Well i guess in that case this should be ready :P

garyb:
Thanks!

Issue #766, opened 2021-08-14
Component rendered improperly when raising messages during initialization


simonolander:
# SSCCE

https://try.purescript.org/?gist=0b8ea8a9d25aa093b1876278d45e09f0

## Expected

Child component should render "Initialized" after initialization.

## Actual

Child component renders both "Loading..." and "Initialized" after initialization.

### Leads

The following changes, individually, makes the problem disappear:
* Changing `text "Loading...` to `p_ [ text "Loading..." ]` in the child component
* Removing the `connect ...` in the child component
* Removing the `raise ...` in the child component
* Changing `div_ [ slot ... ]` to `slot ...` in the parent component
* Removing the handling of the message in the parent component

garyb:
I think this is a duplicate of #657. The main trigger here is the `purescript-halogen-store` machinery, as from looking at the code of that it looks like it will render a "naked" slot as the HTML for the component returned by `connect`.

garyb:
There's something tricky going on here. I've not managed to replicate this without using `halogen-store` as in this case, or `halogen-hooks` as in the linked repo in the other issue. I've tried a few things, but mustn't have quite replicated the exact states that those libraries pass through yet. 

At least there's one piece of information I have now, that this isn't caused simply by using a slot as the render product for a component, nor a nested arrangement of such.

garyb:
I know what's happening now, but not exactly how the state arises. It's really odd that it requires at least a three-deep component hierarchy to trigger it, it seems like if it happens with three it should be reproducible with two, but I've not managed to do that yet.

There are a number of conditions required for this bug to occur, but one of them is that the innermost component that is rendering the content must change its root DOM node. So from `HH.text ""` to a `HH.div`, or a `HH.hr` to a `HH.p`, or whatever. When it renders a consistent top level node, I've not seen any of these kinds of problems occur (that's not a fix or workaround, just a comment on the conditions that are buggy).

From liberally logging the processes going on, something happens such that we process a `Step` for a component where it has the initial value for the root element of the component rather than the current. So if it was first rendering `HH.hr` and then some `HH.div ...`, it does what it should in one place to render the new stuff, but then it encounters the `Step` for the component again afterwards, which has the `HH.hr` still, so it patches it back in. I've observed it `halt`ing the element that is removed, producing the new `Step` with the correct node, and then it somehow the earlier version of it appears again later on in a `patchElem`.

[...]

Issue #657, opened 2020-05-24
Unexpected behavior with child components with a single slot


marcusbuffett:
Recently encountered an issue with child components that just slot in other components. Here's what the component tree looks like:

```
- Router component
  - Slot "home"
    - Home component
      - Slot "innerHomeComponent"
        - Inner home component
  - Slot "about"
    - About component
      - Slot "innerAboutComponent"
        - Inner about component
```

In the router, the "home" and "about" slots should never both be rendered, like
this:

```
  render { route } = do
    HH.div [ class_ (ClassName "should-only-have-one-child") ]
      [ case route of
          Just r -> case r of
            Home -> HH.slot (SProxy :: _ "home") 0 Home.component unit absurd
            About -> HH.slot (SProxy :: _ "about") 1 About.component unit absurd
          Nothing -> HH.div_ [ HH.text "Oh no! That page wasn't found." ]
      ]
```

But after routing from home to about, I get both home and about rendered in the
`.should-only-have-one-child` div.

What's triggering the issue is that the Home/About components that the router is
rendering are just slots for their "inner" components. In the example this
doesn't make much sense but in my real code I have that outer component
providing some input to the child component (the parent component is sorta like
the redux connect HOC in react).

Wrapping the inner component's slot in a div fixes the issue:

```diff
-    Hooks.pure $ HH.slot _innerComponent unit innerComponent unit absurd
+    Hooks.pure $ HH.div [] [ HH.slot _innerComponent unit innerComponent unit absurd ]
```

Anyway, I'm sure this hasn't done a great job of explaining the issue, I made a
repo with a reproduction of the issue: https://github.com/marcusbuffett/halogen-slots-issue

And here's a PR with the div fix mentioned above: https://github.com/marcusbuffett/halogen-slots-issue/pull/2/files

The example code does use hooks, but it's not a hooks issue, converting the parent components to "plain" components causes the same issue.

BebeSparkelSparkel:
Slots have always behaved a bit strange for me and this brings to light a lot of issues I have had with them. I thought it was me.

PR #820, opened 2023-10-10
Store parent node in render #802

### LINKED ISSUE #802: Child component gets rendered outside of parent, being moved to bottom of HTML body
I've been experiencing some bizarre non-deterministic behaviour in an app I'm writing, which I think might be a Halogen issue. Basically I've got a child component which occasionally somehow gets moved to the very bottom of the HTML body, outside of its parent. Unless my mental model of how components are supposed to work is completely wrong I can't see how this could ever happen in Halogen or in the code I've written, but trying to isolate the issue I think I've narrowed it down to a higher-order component I wrote like:

```
premap
    :: forall a b o m query
     . Monad m
    => (forall i r. a -> HTML i r)
    -> (a -> m b)
    -> Component query b o m
    -> Component query a o m
premap html f h = mkComponent
    { initialState: Left
    , render: case _ of
        Left a -> html a
        Right a -> slot _nil unit h a Raise
    , eval: mkEval defaultEval
        { handleAction = handleAction
        , handleQuery = handleQuery
        , receive = Just <<< Receive
        , initialize = Just Initialize
        }
    }
  where
    handleAction (Raise o) = raise o
    handleAction (Receive i) = do
        b <- lift (f i)
        put (Right b)
    handleAction Initialize = do
        s <- get
        case s of
            Left a -> do
                b <- lift (f a)
                put (Right b)
            _ -> pure unit
    handleQuery :: forall a. query a -> HalogenM _ _ _ _ m (Maybe a)
    handleQuery f = request _nil unit (const f)
```
I'm wondering if there are any issues using higher-order components like this? I've used this combinator in a few places with no issues up until now, but it only started occurring once I introduced a new component that happens to be built using this; removing the call to `premap` seems to remove the bug. The call to `premap` is nested inside another component which is also built using it, and introducing the child component detaches things starting from the first parent which also uses `premap`, so I start up with something like this:

```
<grandparent>
    <parent>
         <child></child>
    </parent>
</grandparent>
```
And end up with
```
<grandparent></grandparent>
<parent>
    <child></child>
</parent>
```
Because of the non-determinism it's hard to be sure that I've absolutely found the cause of the issue, but I can usually manually trigger the issue in less than a minute when the new component is wrapped in the call to `premap`, even if the call is just `premap (\_ -> HH.text "") pure`. I haven't managed to trigger it without it.

I've even managed to detach the component, click around the app so that a _new_ component gets rendered correctly inside the grandparent, then detach that one as well, ending up with multiple components being rendered outside of the grandparent. This is without using integers or anything for the keys to slots, just proxies, which I assumed was necessary for Halogen to manage state internally, so I should think they would all need to share state, but once they get detached they don't seem to.

I'm really perplexed by the behaviour in any case. Inspecting the html and looking at the render functions together makes me think what is occurring shouldn't be possible.

roryc89:
I believe this may fix #802, #766 and #657. I have tested this on a similar problem I had with higher order components that had render branches that only rendered a single child component with no wrapping elements. 

It seemed to have buggy behaviour when `substInParent` was passed `Nothing` for the parent node. To fix this I added the last found parent node to the `RenderState`


garyb:
I am very interested in this! I don't have time to check it out right now, but will as soon as I can. I've periodically worked on this without much luck, but a lot of that time was spent trying to isolate exactly what was happening since I had no concrete hypothesis for what was causing the issue.

garyb:
Ah thanks, that wasn't anything to do with you, it was unhappily formatted in `main`.

roryc89:
I see that `removeChild` just removes the node from the DOM. I don't know if my changes will break this as `RenderState`s with a node no longer attached to the DOM may still have the new node inserted from `substInParent`.

PR #824, opened 2024-02-20
Add roles

### LINKED ISSUE #823: The Component type should have role annotations
Currently the Component type does not have any role annotations
https://github.com/purescript-halogen/purescript-halogen/blob/4b2144adad9b3809c45d79349ce6b5e97585b11e/src/Halogen/Component.purs#L49-L54

This results in weird, and definitely unsound, things like these to pass the typechecker (note that these are `coerce` and not `unsafeCoerce`):
```purescript
doBadThings :: forall q i o m. Monad m => H.Component q i o Maybe -> H.Component q i o Aff
doBadThings component = coerce component

doEvilThings :: forall q i o m. Monad m => H.Component q i o Effect -> H.Component q i o Aff
doEvilThings component = coerce component

doUglyThings :: forall q i o m. Monad m => H.Component q i o (ReaderT Int m) -> H.Component q i o (ReaderT String m)
doUglyThings component = coerce component
```

From what I understood, I think that that all of the parameters should have a `representational` role, however I am not 100% sure and someone who has a bit more experience and insight into Halogen should give that a second thought.
```purescript
type role Component representational representational representational representational
```

[Some Relevant examples on Try Purescript](https://try.purescript.org/?code=LYewJgrgNgpgBAWQIYEsB2cDuALGAnGAKEJWAAcQ8AXOABQKgjCJPMpoBEkqkA6AMRBQwSAEaw4ACgBmQsAEpWFanACi06TADGNSes07FpZTQAqeAJ60IBAMpa8KMruwBGADRxsAJk9lPVDAAHlSeUCgAzqFw6Mxo0eFoANaeBGjMeJ5a4DBGbCr62lS8AMJQSBERpSBoEULwklAgAOZ5JmoaRbwASjDScBVwvdJK7HC2SJrV%2BFpExmMAEkhNzTAYgwujKlw8vMgWog37h21jhTq8AIIaUtfSpyolNVR4Qns1SGA9MJ-4Ur2-PCmMIoaRURSEMAgABCn1M2HQzQicAAXCi4LI8MsoHAAI4xOAgODAXiID5gYlwAC8AD44Atqmw0GsaPiUITEEgDvAALR0hlPJksvEEol3SEwuEItBIuDZIXxalykAzeDyijM%2BLEKGqABuKCg8MRyLRGMo2JF7KJJLJaE%2BlNp9MZGuFbI55xofKdgpdirdYo0Er1BqNMuR6pqwqpytVyoVVG1IAAqs0oBZQ7LTZiLf7iaSEOSHfznZG-aL-j8MqY4ABJRXAeRwL0CkDxy0cyQAqvjF6I4mKKEptMZ8Ot300aPZWMRzUJwjAVAYU0euBJtAoBML9BKqGEOBwMg2eAQddUIA)

IMO, This is especially important as coercing components is advertised in the Halogen Real World example:

https://github.com/thomashoneyman/purescript-halogen-realworld/blob/ab13d62f95f4eb3ca589c0f043fe415af8408ff9/src/AppM.purs#L86-L87

garyb:
Resolves #823

The other way to resolve this would be to use `foreign import data` rather than nullary `data`, but the disadvantage of that is there's no way to name the type parameters for documentation purposes.

### DIFF
--- a/src/Halogen/Aff/Driver/State.purs
+++ b/src/Halogen/Aff/Driver/State.purs
@@ -74,6 +74,8 @@ data DriverStateX
   (f :: Type -> Type)
   (o :: Type)
 
+type role DriverStateX representational representational representational
+
 mkDriverStateXRef
   :: forall r s f act ps i o
    . Ref (DriverState r s f act ps i o)
--- a/src/Halogen/Component.purs
+++ b/src/Halogen/Component.purs
@@ -52,6 +52,8 @@ data Component
   (output :: Type)
   (m :: Type -> Type)
 
+type role Component representational representational representational representational
+
 -- | The spec for a component.
 -- |
 -- | The type variables involved:
@@ -196,6 +198,8 @@ data ComponentSlotBox
   (slots :: Row Type)
   (m :: Type -> Type)
   (action :: Type)
+
+type role ComponentSlotBox representational representational representational
 
 instance functorComponentSlotBox :: Functor (ComponentSlotBox slots m) where
   map f = unComponentSlot \slot ->
--- a/src/Halogen/Data/Slot.purs
+++ b/src/Halogen/Data/Slot.purs
@@ -45,6 +45,8 @@ foreign import data Any :: Type
 data Slot :: (Type -> Type) -> Type -> Type -> Type
 data Slot (query :: Type -> Type) output slot
 
+type role Slot representational representational representational
+
 newtype SlotStorage (slots :: Row Type) (slot :: (Type -> Type) -> Type -> Type) =
   SlotStorage (Map (Tuple String (OrdBox Any)) Any)
 


acple:
IIRC, type role annotations are definitely needed even if using `foreign import data` .
Type parameters of nullary `data` would be `phantom` but of `foreign import data` seems to be `nominal`.
It makes those data types too much restrictive I think.

acple:
How about other nullary `data` in halogen: `Halogen.Aff.Driver.State.RenderStateX`, `Halogen.Query.ChildQuery.ChildQueryBox`
These are very internal but the below code is still valid:

```purescript
test :: forall a b. RenderStateX a -> RenderStateX b
test = coerce
```

garyb:
Ah thanks, I had a feeling there was at least one other type I was missing, but my search wasn't bringing it up.

I also need to figure out how to fix this build 😐

acple:
I had hit similar issues in some private project about an year ago.
I think this is spago's internal bug regarding to the git integration or something like that.
I personally investigated it but there was no solution, so I switched to spago@next 😞

PR #825, opened 2024-09-13
docs: beginner's guide: mention not to run tutorial code directly

Hi-Angel:
When just starting out, I bet one of the most frequent mistakes is to copy-paste the "Hello World" code into the editor and execute a `spago run` or the like. This will result in the exception though because the tutorial code doesn't create a window, and requires more steps for that to work.

So let's mention that explicitly.

Fixes: https://github.com/purescript-halogen/purescript-halogen/issues/758


Hi-Angel:
…that said, I'm wondering if there's a documentation section that describes how to create such HTML file, etc, that I could link from the new paragraph to.

PR #827, opened 2024-09-17
Read selfRef in finalize

purefunctor:
This changes the implementation of `finalize` in `runUI` to read the `selfRef` similar to what's being done in `dispose` for the `rendering` field.

This fixes a bug wherein finalizers from newly-rendered slots aren't run when `dispose` is called. For example, if you have a router root component, the current `dispose` would only run finalizers for the initial page, and not respond to changes in the router component.

This is needed because the following line in `render` effectively makes it so that the `DriverState` bound in the `dispose` for `HalogenIO` is older than the one that's currently being used in the driver.

```hs
    flip Ref.modify_ ds.selfRef $ mapDriverState \ds' ->
      ds' { rendering = Just rendering, children = children }
```

### DIFF
--- a/src/Halogen/Aff/Driver.purs
+++ b/src/Halogen/Aff/Driver.purs
@@ -255,7 +255,8 @@ runUI renderSpec component i = do
     -> DriverStateX r f' o'
     -> Effect Unit
   finalize lchs = do
-    unDriverStateX \st -> do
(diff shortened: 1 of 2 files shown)


garyb (approved):
Nice find, thanks!

PR #835, opened 2025-06-08
adding onTimeUpdate event handler

m-rinaldi:
This PR adds support for the `timeupdate` event via the `onTimeUpdate` handler, enabling monitoring of playback time.

For more info, have a look at: https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/timeupdate_event


m-rinaldi:
Follow-up for the previous PR in [`dom-indexed`](https://github.com/purescript-halogen/purescript-dom-indexed/pull/36)
@garyb it would be great if you could take a look here too 🙏

Thank you 😊

